Issue #103, opened 2013-01-13
Create `age` validator (Deprecates `minimumAge`)


Petah:
This resumes the discussion happened below until @caferrari [comment][1]:

1. Create an `age` validation rule: `v::age($min, $max=null)`.
2. Deprecate (remove) `mininumAge` validation rule in favor of the `age` rule.

Original message (for historical purposes):
> There is a minimumAge but no maximumAge.
> Any reason, or just missed? 

[1]: https://github.com/Respect/Validation/issues/103#issuecomment-12889825

henriquemoody:
This rule is going to be available in `9.0` version.

PR #150, opened 2014-02-18
Add rule "Age" to validation.

malukenho:
Add "Age" rule in favor of "MinimumAge".

Age rule works pretty much the same way “minimumAge”
worked but it optionally accepts a maximum age as second
argument.

See #103.

### Blockers for merge

- [ ] Add `age` rule on README.
- [ ] Add `age` rule on Validator docblocs.
- [ ] Remove `minimumAge` rule from Validator docblocs.
- [x] Modify README to reflect removal of `minimumAge` (point deprecation in favor of `age`).
- [x] Clean up unrelated commits and merges.
- [x] Make `age` compatible with `minimumAge`.


henriquemoody:
I'm sorry by our late.

@malukenho, can you please rebase with our current master an update documentation (README.md and Validator docblocks)?

library/Rules/Age.php
```diff
@@ -0,0 +1,45 @@
...
+            if ($this->maxAge) {
+                $limit = new DateTime('now + '.$this->maxAge. ' year');
+                return $birthday > $input and $limit < $input;
+            }
+
+            return $birthday > $input;
+
+        } elseif (!is_string($input) || (is_null($this->format) && false === strtotime($input))) {
+            return false;
+        } else {
```
henriquemoody: Avoid to use `else`.

library/Rules/Age.php
```diff
@@ -0,0 +1,45 @@
...
+            $age = ((date('Ymd') - date('Ymd', strtotime($input))) / 10000);
+
+            if ($this->maxAge) {
+                return $age >= $this->minAge and $age <= $this->maxAge;
+
+            }
+            return $age >= $this->minAge;
+        }
+}
```
henriquemoody: Missin `    ` here.

library/Rules/MinimumAge.php
```diff
@@ -1,36 +0,0 @@
-<?php
-namespace Respect\Validation\Rules;
-
-use DateTime;
-
-class MinimumAge extends AbstractRule
```
henriquemoody: Do the same was done in [Vowels](https://github.com/Respect/Validation/blob/master/library/Rules/Vowels.php#L9-L10) rule, for example, them we can remove it in the next major release.

tests/library/Respect/Validation/Rules/Age.php
```diff
@@ -0,0 +1,69 @@
+<?php
+namespace Respect\Validation\Rules;
+
+use DateTime;
+
+class AgeTest extends \PHPUnit_Framework_TestCase
```
henriquemoody: Wrong file name, also wrong path, it must be `tests/Rules/AgeTest.php`

henriquemoody:
@malukenho, I'm sorry to make you change it again, but we decided to keep BC between [minor versions](http://semver.org/), so just let `maximumAge()` as it is unless `age()` is totally compatible with `maximumAge()` rule, if that's the case you can just use inheritance on `MaximumAge`.

tests/library/Respect/Validation/Rules/Age.php
malukenho: My fail, that need to be removed

henriquemoody:
@malukenho, is there any progress on that issue?

AndyWendt:
I could look into this if you like and get it fixed up.  Unless there is something else you would rather have me work on.

henriquemoody:
If @malukenho wants to pass this to another one I'm ok with that, let's just wait for his answer.

henriquemoody:
@AndyWendt, since @malukenho didn't answer our question and there are 21 days since the last interaction, I'd be glad if you finish it.

PR #158, opened 2014-03-21
Allow support for IDN with subdomains

vrnagy:
I just moved the IDN rulo from checks array to otherParts validation, because the old approach did not supported an IDN with subdomain


henriquemoody:
@vrnagy, I'm sorry by our late. Can you rebase it with our current `master` branch, and, if possible squash your commits?

PR #172, opened 2014-05-04
Make between inclusive by default

flip111:
I think more developers expect between to be inclusive. Like [MySQL](https://dev.mysql.com/doc/refman/5.7/en/comparison-operators.html#operator_between).


henriquemoody:
@flip111, can you send this pull request against `1.0` branch (after some rebase) but also update the documentation?

Plus, would be awesome if you make the same change on `Min` and `Max` rules.

Issue #154, opened 2014-03-05
How to validate notEmpty bit value 0/1


NeedToMeet:
I noticed that the NotEmpty rule uses PHP's empty(), which always evaluates the value of numeric 0 as false. Values of non-nullable 0 or 1 are common in bit-type fields where an on/off value is set. Any suggestions on how to validate this as non-empty?

dereuromark:
You could have a method boolNotEmpty() which would allow 0 and 1 to pass as not empty, whereas "null" or empty string would be empty then. Just an idea.

henriquemoody:
Will be fixed on version 1.0, see [changelog](https://github.com/Respect/Validation/blob/311891b/CHANGELOG.md#100---2015-10-24) for more info about the version.

To fix this issue now all rules are mandatory, but you can use [optional()](https://github.com/Respect/Validation/blob/311891b/docs/Optional.md) rule to handle optional values, which now is `null` and an empty string by default.

PR #173, opened 2014-05-04
notEmpty fix

flip111:
For strings only: notEmpty now means "there was content supplied", rather than "the content was not empty according to the php empty function".

Effect: notEmpty no longer validates to false when string "0" is supplied.

This make more sense when validating HTML forms.

Related issue: https://github.com/Respect/Validation/issues/154

## Backward Compatibility Break

Currently the `notEmpty` rule works like:

```php
v::int()->notEmpty()->validate(0); //false
v::int()->notEmpty()->validate("0"); //false
```

If this PR is accepted:

```php
v::int()->notEmpty()->validate(0); //false
v::int()->notEmpty()->validate("0"); //true <-- BC Break
```

## Missing on this PR

If this PR gets accepted by other members, it still misses some stuff which I will point out just for the sake of not trusting my memory:

- [ ] Test to ensure current behavior.
- [ ] Test checking behavior using different types (string, int, float, boolean) to check for empty values.
- [ ] Change on README.
- [ ] Better short commit message (Ex: "NotEmpty returns true to 0 (zero) for strings.").

### DIFF
--- a/library/Respect/Validation/Rules/NotEmpty.php
+++ b/library/Respect/Validation/Rules/NotEmpty.php
@@ -6,7 +6,7 @@ class NotEmpty extends AbstractRule
     public function validate($input)
     {
         if (is_string($input)) {
-            $input = trim($input);
+            return trim($input) !== '';
         }
 
         return !empty($input);


henriquemoody:
Actually I agree with this change, but to keep the BC it must be optional.

@flip111, can you put an optional argument on `NotEmpty`'s constructor?

IMO, the two lines bellow must return `true` with this new behavior
```
v::int()->notEmpty(true)->validate('0');
v::int()->notEmpty(true)->validate(0);
```

bmorg:
> For what its worth... in meantime:
> 
> ```
> v::oneOf(
>     v::numeric(), 
>     v::notEmpty()
> )->validate("0"); 
> //true
> ```
> ...verbose.

Unfortunately this does not work when using `assert()` or `check()` (see #235).

dereuromark:
You could call it stringNotEmpty() to keep the old BC method in place as it was.

qrazi:
We encountered the same issue and ended up implementing a custom rule which we called NotOptional. The reasoning behind this stems from the documentation, [specifically optional input](https://github.com/Respect/Validation#input-optional). The documentation describes the behaviour as values being optional by default, and an empty string having the meaning of a not-supplied-value. Since the default behavior is for everything to be optional, reversing that rule would logically be called NotOptional. The documentation and the code specifies an empty string as representing a not-supplied value, so the NotOptional rule should check just against that.

Would there be interest in a pull request for our implementation?

bmorg:
> Would there be interest in a pull request for our implementation?

That would be great!

Even if the pull request is rejected by the core team, it would serve as a reference for people who need a working fix right now and are stuck with implementing workarounds.

henriquemoody:
@qrazi, can you send the pull request agains 1.0 branch?

qrazi:
@henriquemoody Yes, I will. I have to do some cleanup of the code etc, but will try to finish up the pr this weekend.

qrazi:
@henriquemoody I made a commit (several actually, still learning git... ) that adds the NotOptional rule without breaking the current behaviour of NotEmpty and NotNull

FoxxMD:
:+1: Really need this functionality

qrazi:
@henriquemoody I'm not sure if I did this properly with my commit. Should I have opened a new PR instead of referencing this PR?

henriquemoody:
@qrazi, yes, you need to create a pull request. Be sure you sent it agains 1.0 branch.

bmorg:
Has anyone come up with a workaround yet using only Request/Validation?

I have tried the following validator:

```php
v::oneOf(
    array(
        v::int()->equals(0),
        v::int()->notEmpty()
    )
);
```

But it doesn't work since it accepts the empty string in the first part.

henriquemoody:
Hey, guys, it's been a long time since my last comment on this issue but it was never forgotten.

I'm working on [1.0](https://github.com/Respect/Validation/tree/new_api) version and this is one of the problems we have on our current code base but we cannot change it without breaking BC, so on this new version we are going to fix this and other issues we have.

After this discussion looks like we have 3 validators with are similar but not the same:
- **NotBlank**: blank values
- **NotEmpty**: the opposite behavior of PHP's empty()
- **NotOptional**: the optional values for input, which I consider as an empty string (`''`) and null (`null`) only.

I would like to head about you guys what you think about it:

Value       | Is it optional? | Is it empty? | Is it blank?
------------|-----------------|--------------|----------
 `null`     | Yes             | Yes          | Yes
 `""`       | Yes             | Yes          | Yes
 `" "`      | No              | No           | Yes
 `0`        | No              | Yes          | Yes
 `"0"`      | No              | Yes          | Yes
 `0.0`      | No              | Yes          | Yes
 `"0.0"`    | No              | Yes          | Yes
 `false`    | No              | Yes          | Yes
 `[""]`     | No              | No           | Yes
 `[" "]`    | No              | No           | Yes
 `[0]`      | No              | No           | Yes
 `["0"]`    | No              | No           | Yes
 `[false]`  | No              | No           | Yes

### TL-DR

Just tell us if you agree with the table above.

henriquemoody:
Also, I would like to know what you, @Respect/troubleshooters, think about my last comment.

FoxxMD:
@henriquemoody Thanks for taking the time to work through this issue.

Am I correct in interpreting the table to mean that providing `0` to a `notOptional` will evaluate to true?

henriquemoody:
Exactly, @FoxxMD.

FoxxMD:
Awesome :+1:

qrazi:
@henriquemoody Table looks good to me :+1:

bmorg:
@henriquemoody What would the three validators return for an integer value of -1?

henriquemoody:
I've just updated the comment, but all rules must return `true` for `-1`, since it's not _blank_, _empty_ or _optional_.

bmorg:
> I've just updated the comment, but all rules must return true for -1, since it's not blank, not empty and not optional.

Thanks for clearing that up. I was confused because the comment above the table says "number lower than 0".

The `notOptional` validator looks good to me. Definitely what I initially expected `notEmpty` to do.

I for one would be even more radical and rename the old ```notEmpty``` to `notPhpEmpty`.

Also to me it was not immediately evident what `notBlank` would be doing.

These last two remarks might only be me though.

krazyjakee:
@henriquemoody Table looks good dude

henriquemoody:
Will be fixed on version 1.0, see [changelog](https://github.com/Respect/Validation/blob/311891b/CHANGELOG.md#100---2015-10-24) for more info about the version.

To fix this issue now all rules are mandatory, but you can use [optional()](https://github.com/Respect/Validation/blob/311891b/docs/Optional.md) rule to handle optional values, which now is `null` and an empty string by default.

I tried to keep all input as optional on version 1.0, but without changing the concrete API I have to do a very poor workaround, and after many issues created because of that damn feature I decided that making all rules mandatory would be for the best.

On version 2.0, that changes the concrete API, we will be able to do that, but I think it's best to keep it as is, because with the `v::optional()` the created chain is more obvious:

```php
v::int()->validate('');
```
`''` is not a integer, so that should really pass?

```php
v::optional(v::int())->validate(''); 
```

It will pass not because it's a integer, but because it's optional and that's clear in the chain.

Will be nice to have some feedback on that, I will can revert the changes if it's the case.

PR #176, opened 2014-05-07
Allow to define custom rule's namespaces/prefixes.

henriquemoody:
```php
v::with('My\\Validation\\Rules\\');
v::myRule(); // Try to load "My\Validation\Rules\MyRule" if any
```

By default `with()` appends the given prefix, but you can change this behavior
in order to overwrite default rules:

```php
v::with('My\\Validation\\Rules\\', true);
v::alnum(); // Try to use "My\Validation\Rules\Alnum" if any
```

Closes #206.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,23 @@ $errors = $exception->findMessages(array(
 
 For all messages, the `{{name}}` and `{{input}}` variable is available for templates.
 
+### Custom Rules
+
+You also can use your own rules:
+
+```php
+v::with('My\\Validation\\Rules\\');
+v::myRule(); // Try to load "My\Validation\Rules\MyRule" if any
+```
+
+By default `with()` appends the given prefix, but you can change this behavior
+in order to overwrite default rules:
+
+```php
+v::with('My\\Validation\\Rules\\', true);
+v::alnum(); // Try to use "My\Validation\Rules\Alnum" if any
+```
+
 ### Validator Name
 
 On `v::attribute()` and `v::key()`, `{{name}}` is the attribute/key name. For others,
--- a/library/Factory.php
+++ b/library/Factory.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Respect\Validation;
+
+use ReflectionClass;
+use Respect\Validation\Exceptions\ComponentException;
+
+class Factory
+{
+    protected $rulePrefixes = array('Respect\\Validation\\Rules\\');
+
+    public function getRulePrefixes()
+    {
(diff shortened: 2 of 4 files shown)


henriquemoody:
I think it's better we don't merge this pull request for all the reasons @augustohp already said.

alganet:
I'm still up for implementing something around `class_alias`. It patches the Zend naming hash table directly, so manual array walks aren't needed each time a new rule is found. Something in the lines of:

```php
class Validator 
{
    // ....
    public function appendRulePrefix($prefix)
    {
        $prefixLoader = function ($ruleSuffix) use ($prefix) {
            class_alias('Respect/Validation/Rules/' . $ruleSuffix, $prefix . $ruleSuffix );
        }
        $this->prefixLoaders[$prefixLoader] = $prefix;
        spl_autoload_register($prefixLoader);
    }
    // ....
}
```

Prepending would require a magic trick listing autoloaders, reordering them and so on. It's complex, but it is one time only. Just adding/removing prefixes sounds more sane. Removal can be done with spl_autoload_unregister, that's why the `prefixLoaders` collection (suggested as an SplObjectStorage to keep the closures and prefixes accessible). I haven't tested this in recent PHP/HHVM versions though, so I might be completely wrong about how this works.

henriquemoody:
Actually prepend would not be a problem, since `spl_autoload_register()` accepts a boolean value as third argument which tells the autoloader if the given autoloader callback is prepended or not.

alganet:
I didn't knew about this, awesome!

henriquemoody:
But I'm not sure if we should use this instead of #206, actually I don't like both APIs, so if anyone has a better idea it would be awesome.

alganet:
The only class that uses this namespace thing is Validator. We can even create a new one. The cleanest thing I could think about is:

a) Make `Respect\Validation\Validator::buildRule` use a property instead of a variable for `$validatorFqn`. This will allow it to be extended properly to different namespaces. 
b) Create a method for switching namespaces.

API would look like:

```php

v::with('My\\Application\\ValidationRules\\')
 ->foreignUser($fingerpring)
 ->with('SpanishWorldAssociation\\RegrasDeValidacion\\')
 ->spanishUser($apellido)
 ->with(v::NS)
 ->length(1,300);
```

henriquemoody:
I like the idea, the only problem I see is that you have to use `with()` every time you want a custom rule, any way, it's much better than what I proposed.

henriquemoody:
Pull request has been updated. According to my tests there is no performance problems.

PR #184, opened 2014-06-04
Change docblocks to reflect magic static methods

rizqidjamaluddin:
Some IDEs use a class' `@method` docblock annotations to autocomplete, suggest and typehint magic methods for users. The `validator.php` class has these, but doesn't mark them as static, so the suggested `V::method()` syntax continuously throws errors:

phpstorm

IDE I'm using is PHPStorm 7.1.3 (latest as of writing).

It appears that adding `static` annotations doesn't affect normal (dynamic) calls because PHP is okay with calling statics normally. Clearly the other way around doesn't seem to work!

(I noticed a similar request from long ago - I apologize if there's already a rationale for not doing this.)


henriquemoody:
I'm sorry by our late. Seems like there is a lot of people complaining about that, we just discuss it on #202 2 hours ago. In fact we cannot use only `static` declarations since you can use it in chains, but the best think to do, IMO is create 2 declarations for every rule, what you guys think about it?

henriquemoody:
Actually, it a method is static it can me used as non-static, so the `static` must work for both.

Issue #86, opened 2012-10-10
Why are validation shown against value rather than field name?


silentworks:
Currently If I try to validate $_POST['username'] I get the value in the validation message rather than the name username?

```php
try {
    v::alnum()
        ->noWhitespace()
        ->length(4,22)
        ->check($_POST['username']);
} catch (\InvalidArgumentException $e) {
    var_dump($e->getMainMessage());
}
```

result
```
"" must not contain whitespace
```

what I was expecting
```
"username" must not contain whitespace
```

FoxxMD:
:+1:

asperon:
Anybody looking into this?

henriquemoody:
Looks like we still haven't changed it yet, even with the [Key](http://respect.li/Validation/docs/key.html) rule.

This code:
```php
try {
    v::key('username', v::alnum()->noWhitespace()->length(4,22))->check($_POST);
} catch (ValidationExceptionInterface $exception) {
    echo $exception->getMainMessage().PHP_EOL;
}
```

Displays this message:
```
"a b" must not contain whitespace
```

But what you was expecting was:
```
"username" must not contain whitespace
```

I'm not sure if we can even fix this on the current version because even if it's not the expected behaviour by many of you I don't see this as a Bug, it's just a different behaviour and there are test counting with this behaviour.

However, this is going to change on our first major version, as you can see on this test: [simple-full-message-with-key-4](https://github.com/Respect/Validation/blob/new_api_2/tests/functional/assert/simple-full-message-with-key-4.phpt). So, you may consider this as covered.

henriquemoody:
Any way, this looks easy to change: https://github.com/henriquemoody/Validation/commit/9700821

I'm just not sure if we should release this on 0.10 or just on 1.0. Any thoughts?

augustohp:
> I'm just not sure if we should release this on 0.9<, 0.10 or just on 1.0. Any thoughts?

0.10, I would not change this on `0.9` which is the current version, right?

henriquemoody:
Yeah, I just kept it on master only.

chozilla:
Hello, as fare as I can tell, the described problem is not fixed.
The **->setName()** function still does not behave as expected or described in the use cases in this discussion.

```php
// example
$testValue = array();
$testValue['nameFirst'] = 'Jane DOOO0E FILLTEXT! MUCH LONG';
$testValue['nameLast'] = 'Doe';

[...]

PR #185, opened 2014-06-04
Exception messages using field name instead of value

SBhojani:
Fixing the issue discussed in https://github.com/Respect/Validation/issues/86 so atleast the following code gives the field names in the error messages:

    try{
        v::string()->setName('First Name')->notEmpty()->noWhitespace()->alpha()->assert($_POST['txt_first_name']);
    }catch (\Respect\Validation\Exceptions\ValidationException $e){
        echo $e->getFullMessage();
    }

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,4 +1,4 @@
 {
-    "name": "s-bhojani/Validation",
+    "name": "s-bhojani/validation",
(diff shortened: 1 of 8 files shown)


henriquemoody:
Looks like this PR went to a different path and has an own packagist package.

PR #188, opened 2014-06-25
Add (German) bank account validation with malkusch/bav

malkusch:
Hi
I maintain the library malkusch/bav for validation of German bank accounts. This pull requests adds following rules to include bav into your library:

* ~~`v::germanBank()`~~
* ~~`v::germanBankAccount()`~~
* ~~`v::germanBIC()`~~
* `v::bank(string $countryCode)`
* `v::bankAccount(string $countryCode, string $bank)`
* `v::bic(string $countryCode)`

This PR supports only the country code "de".

Enjoy.


henriquemoody:
@malkusch, first of all, I'm sorry by our late.

This pull request must be rebased with the current master branch, also, I don't think a good idea to have a rule for an specific country, can you do something more generic like what was done in #226?

malkusch:
German bank validation has more than 100 validation rules. I'm too lazy to extend the validation for other countries. I will rebase if there is demand for a german bank validation.

henriquemoody:
I didn't say you have to extend the validation for other countries. What I'm trying to avoid is to have `germanBank()`, `braziliamBank()`, `americamBank()` and such... Define a country code in the constructor of the validator sounds too much better for me.
For a while, doesn't matter if you have `german*` methods in the validator, but since you keep the API clean we can refactor this when other validators come.

malkusch:
Ok, Sorry I got you wrong. This sounds of course reasonable. I'll rebase and refactor the validator to a universal API with a country code dependency.

henriquemoody:
Thanks :smile:

malkusch:
Rebased to master and refactored to `Bank($countryCode)`, `BankAccount($countryCode)` and `BIC($countryCode)`.

henriquemoody:
Thanks, @malkusch, now I can review it!

Fist of all, you should fix the detected [issues](https://scrutinizer-ci.com/g/Respect/Validation/inspections/59fae009-7aec-4845-ace3-a3a030641cd0/issues/) on the [Scrutinizer's build](https://scrutinizer-ci.com/g/Respect/Validation/inspections/59fae009-7aec-4845-ace3-a3a030641cd0).

I will start to review it now.

README.md
```diff
@@ -493,6 +499,46 @@ See also:
...
+```php
+v::bank("de")->validate("70169464"); //true
+v::bank("de")->validate("12345"); //false
+```
+
+These country codes are supported:
+
+ * "de" (Germany)
+
+This validator needs `malkusch/bav` as dependency.
```
henriquemoody: This is a dependency of `de` only, so I think you should put it on the same line of `de`.

README.md
```diff
@@ -493,6 +499,46 @@ See also:
...
+```php
+v::bankAccount("de", "70169464")->validate("1112"); //true
+v::bankAccount("de", "70169464")->validate("1234"); //false
+```
+
+These country codes are supported:
+
+ * "de" (Germany)
+
+This validator needs `malkusch/bav` as dependency.
```
henriquemoody: Same here.

README.md
```diff
@@ -535,6 +581,43 @@ See also:
...
+
+$configuration = new DefaultConfiguration();
+
+$pdo = new \PDO("mysql:host=localhost;dbname=test");
+$configuration->setDataBackendContainer(new PDODataBackendContainer($pdo));
+
+return $configuration;
+```
+
+The German validator needs `malkusch/bav` as dependency.
```
henriquemoody: Same here.

README.md
```diff
@@ -535,6 +581,43 @@ See also:
...
+*bav/configuration.php*:
+```php
+namespace malkusch\bav;
+
+$configuration = new DefaultConfiguration();
+
+$pdo = new \PDO("mysql:host=localhost;dbname=test");
+$configuration->setDataBackendContainer(new PDODataBackendContainer($pdo));
+
+return $configuration;
```
henriquemoody: I'm not sure we should keep it into our documentation, it looks more a recommendation for the library than of the rule in itself.

Any thoughts?

library/Rules/BIC.php
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Sets the country code.
+     * 
+     * The country code is not case sensitive.
+     * 
+     * @param string $countryCode The ISO 639-1 country code.
+     */
+    public function __construct($countryCode)
+    {
+        parent::__construct();
```
henriquemoody: This call is not necessary.

library/Rules/BIC.php
```diff
@@ -0,0 +1,57 @@
...
+        $this->countryCode = $countryCode;
+    }
+    
+    public function validate($input)
+    {
+        switch (strtolower($this->countryCode)) {
```
henriquemoody: A country code must be always in uppercase, otherwise it's wrong.

library/Rules/BIC.php
```diff
@@ -0,0 +1,57 @@
...
+            case "de":
+                $bav = new BAV();
+                return $bav->isValidBIC($input);
+                
+            default:
+                $message = sprintf(
+                        "Cannot validate BIC for country '%s'.",
+                        $this->countryCode);
+                throw new ComponentException($message);
```
henriquemoody: This is better on constructor.

library/Rules/Bank.php
```diff
@@ -0,0 +1,52 @@
...
+                $bav = new BAV();
+                return $bav->isValidBank($input);
+                
+            default:
+                $message = sprintf(
+                        "Cannot validate bank for country '%s'.",
+                        $this->countryCode);
+                throw new ComponentException($message);
+        }
+    }
```
henriquemoody: The same changes I recommended to [BIC](https://github.com/Respect/Validation/pull/188/files#diff-5) rule.

library/Rules/BankAccount.php
```diff
@@ -0,0 +1,59 @@
...
+            case "de":
+                $bav = new BAV();
+                return $bav->isValidBankAccount($this->bank, $input);
+                
+            default:
+                $message = sprintf(
+                        "Cannot validate bank account for country '%s'.",
+                        $this->countryCode);
+                throw new ComponentException($message);
+        }
```
henriquemoody: The same changes I recommended to [BIC](https://github.com/Respect/Validation/pull/188/files#diff-5) rule.

README.md
```diff
@@ -329,6 +329,12 @@ Reference
   * [v::uploaded()](#vuploaded)
   * [v::writable()](#vwritable)
 
+### Banking
+
+ * [v::bank()](#vbankstring-countrycode)
+ * [v::bankAccount()](#vbankaccountstring-countrycode-string-bank)
+ * [v::bic()](#vbicstring-countrycode)
```
henriquemoody: You should put these rules into Validator docblock too

henriquemoody:
If you use `Callback` instead of `AbstractRule` as you base class I think it's going to be better, but even using `Callback` you have to create instances inside the method and it is bad in many aspects because you cannot use dependency injection on it, that's why you need to configure and call `malkusch/bav`' objects instead of mock them.

I have to think about something better but now my headache doesn't let me do it. :tired_face:

library/Rules/BIC.php
malkusch: My IDE complains that the object might be wrong initialized. And I tempt to agree as the body of the existing parent constructor is an implementation detail which I don't know.

library/Rules/BIC.php
malkusch: C'mon ThIS Is pHP. NoBoDy caRes aBout CaSe. I refered in the docblock to ISO 639-1 which I found e.g. here in lower case: http://www.loc.gov/standards/iso639-2/php/English_list.php The standard itself is behind a paywall.

The code above is BTW case insensitive. Is it a respect convention that a country code has to be in upper case? I'm just curious. I'll change that code to fit your conventions.

library/Rules/BIC.php
henriquemoody: You may keep it, actually it's a good practice what you did, but in our case it's unnecessary ([see the constructor yourself](https://github.com/Respect/Validation/blob/master/library/Rules/AbstractRule.php#L14-L17)), but keep it would not be a problem at all.

library/Rules/BIC.php
henriquemoody: My bad, I just took a look on [countryCode()](https://github.com/Respect/Validation/blob/master/library/Rules/CountryCode.php#L29).

You must keep it.

malkusch:
> but even using Callback you have to create instances inside the method and it is bad in many aspects because you cannot use dependency injection on it

Yes, that comes with the universal validator which gets only the country code injected. You probably want to keep the API clean and simple i.e. prefer `__construct($countryCode)` over `__construct(BankLocale $implementation)`.

henriquemoody:
I have to think more about it, my first thought was  `__construct($countryCode, CallableRepository $repository = null)` but it doesn't looks like other Respect stuff. Any thoughts, @alganet?

henriquemoody:
Any way, it's not a blocker, we can refactor it, since we have a clean API.

@malkusch, just move the validation to the rules' constructors and, if you possible, squash these commits (there are a lot of them for just 3 rules) and I guess we can merge it as it is.

malkusch:
I'm on it.

malkusch:
Refactored and squashed.

library/Rules/BIC.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $callback = null;
+
+        switch (strtolower($countryCode)) {
+            case "de":
+                $bav = new BAV();
+                $callback = function ($bic) use ($bav) {
+                    return $bav->isValidBIC($bic);
+                };
+                break;
```
henriquemoody: Why not?

```php
$callback = array(new BAV(), 'isValidBIC');
```

library/Rules/Bank.php
```diff
@@ -0,0 +1,48 @@
...
+    public function __construct($countryCode)
+    {
+        $callback = null;
+
+        switch (strtolower($countryCode)) {
+            case "de":
+                $bav = new BAV();
+                $callback = function ($bank) use ($bav) {
+                    return $bav->isValidBank($bank);
+                };
```
henriquemoody: Same here.

library/Rules/BankAccount.php
```diff
@@ -0,0 +1,49 @@
...
+    public function __construct($countryCode, $bank)
+    {
+        $callback = null;
+
+        switch (strtolower($countryCode)) {
+            case "de":
+                $bav = new BAV();
+                $callback = function ($account) use ($bank, $bav) {
+                    return $bav->isValidBankAccount($bank, $account);
+                };
```
henriquemoody: Here too.

malkusch: Because this is the place where the array callback doesn't work anymore. The parameter order wouldn't match for `isValidBankAccount($bank, $account)`. `Callback::validate()` would call `isValidBankAccount($account, $bank)`. Even if the `array_unshift()` would be a `array_push()` it's very bad practice to rely on that implementation detail.

So the best solution here was the anonymous function. As you Respect is PHP>=5.3 there is no problem using them. I used them in the other places because of consistency.

henriquemoody: :dancer:

henriquemoody:
Thanks for all your work!!

PR #195, opened 2014-07-14
Delete package.xml

GrahamCampbell:
Do we still need this?


henriquemoody:
@GrahamCampbell, I'm sorry by out late, actually I just get back to respect few minutes ago.

I created Respect/project-info#17 in order to discuss that, but it's going to be solved very soon.

PR #198, opened 2014-07-16
Add RGBColor validator

DavidePastore:


henriquemoody:
First of all, I'm sorry by out late.

Seems like this is for RGB Hex Triplet Color, only do I don't think RGB color is a good name, I have these ideas:

- hexColor()
- hexRgbColor()

Also, I don't think we should keep the `#` in the beginning, at least it should me optional.

maboiteaspam:
seems great, let us know if you want some support for testing or whatever !

> Also, I don't think we should keep the # in the beginning, at least it should me optional.
Possibly yes. Maintainer will let us know.

DavidePastore:
Ok, other changes are here. Check them :cake:

library/Rules/HexRgbColor.php
```diff
@@ -0,0 +1,22 @@
...
+        if(v::oneOf(v::object(), v::arr(), v::nullValue())->validate($input)){
+            return false;
+        }
+        
+        if(!v::startsWith('#')->validate($input)){
+            $input = '#' . $input;
+        }
+        
+        return v::xdigit()->validate(mb_substr($input, 1))
+                        && hexdec(mb_substr($input, 1)) < 16777216;
```
henriquemoody: Since you're validating only hexadecimal values you don't need to use `mb_*` functions

tests/library/Respect/Validation/Rules/HexRgbColorTest.php
```diff
@@ -0,0 +1,39 @@
...
+        $validator = new HexRgbColor();
+        $this->assertTrue($validator->assert('#0'));
+        $this->assertTrue($validator->assert('#123456'));
+        $this->assertTrue($validator->assert('#1234'));
+        $this->assertTrue($validator->assert('#1234'));
+        $this->assertTrue($validator->assert('#FFFFF'));
+        $this->assertTrue($validator->assert('123123'));
+        $this->assertTrue($validator->assert('FFFFF'));
+        $this->assertTrue($validator->assert(1));
+        $this->assertTrue($validator->assert(0));
```
henriquemoody: RGB triplet colors must be composed by 3 hexadecimal numbers or 3 pairs of them, so, there are only 2 allowed lengths: 3 or 6. That's because this schema is to represent the RGB channels (Red, Green, Blue), each one should not be greater them 255 (FF) and not lesser than 0 (00).

library/Rules/HexRgbColor.php
```diff
@@ -0,0 +1,22 @@
+<?php
+namespace Respect\Validation\Rules;
+
+use Respect\Validation\Validator as v;
+
+class HexRgbColor extends AbstractRule
+{
+    public function validate($input)
+    {
+        if(v::oneOf(v::object(), v::arr(), v::nullValue())->validate($input)){
```
henriquemoody: As [I already](https://github.com/Respect/Validation/pull/198/files#r22520098) said about this schema, it's not a number, they are 3 concatenated numbers. If you allow integers we can have problems, think about this blue: 000099; when converting to string it will be only `99` not and a string with `000099`.

tests/library/Respect/Validation/Rules/HexRgbColorTest.php
```diff
@@ -0,0 +1,39 @@
+<?php
+namespace Respect\Validation\Rules;
+
+class HexRgbColorTest extends \PHPUnit_Framework_TestCase
+{
+    public function testHexRgbColorValuesONLYShouldReturnTrue()
+    {
+        $validator = new HexRgbColor();
+        $this->assertTrue($validator->assert('#0'));
```
henriquemoody: It's a good practice to have 1 assert per test, can you please use data providers instead?

henriquemoody:
Thanks for the changes, @DavidePastore, I think this name is much better!
Unfortunately you need to rebase with our current master, I guess the only conflict you're going to have is on Validator docblocks. If you need any help just tell me. BTW, there is a new chat on Gitter, check it out: https://gitter.im/Respect/Validation

library/Rules/HexRgbColor.php
DavidePastore: Sorry but where is supposed to use `strlen()` function?

henriquemoody: Yeah, it's better. [mb_string is not enabled by default](http://php.net/manual/en/mbstring.installation.php)

henriquemoody: I'm sorry! I'm talking about `substr()` instead. :sweat_smile:

tests/library/Respect/Validation/Rules/HexRgbColorTest.php
henriquemoody: @DavidePastore, to count the number in the hex color, you should use `strlen()`.

library/Rules/HexRgbColor.php
DavidePastore: I used it because I saw that **StartsWith** also uses it. I'll change it :smile: .

henriquemoody: Yeah, there are other rules which uses mbstring to but they need to be refactored. These functions should only be used if mbstring is enabled.

DavidePastore:
@henriquemoody Thanks for your feedback. I've merged all from `master` branch. Please, check them :smiley_cat:

henriquemoody:
Thanks, it's already on master :D

henriquemoody:
Next time use rebase instead of merge: http://git-scm.com/book/en/v2/Git-Branching-Rebasing

DavidePastore:
Thanks for the tip @henriquemoody :cake:

henriquemoody:
You're welcome, thanks for contributing!

PR #202, opened 2014-07-31
fixed phpDocs for Validator.php

KonstantinKuklin:
php Docs didn't work correct for static calls like instructions shows v::int() marked in ide like problem.


henriquemoody:
Actually these methods are allowed in [non-static mode](https://github.com/Respect/Validation/blob/master/library/Validator.php#L127-L139).

The [__callStatic()](https://github.com/Respect/Validation/blob/master/library/Validator.php#L97-L106) method just creates a new instance of Validator.

I don't think this pull request should be merged, how about you @Respect/troubleshooters?

alganet:
IDEs are huge PITAs. Each one has its way to auto-complete magic methods, because they don't care that much. The hard job is to download a couple of IDEs and try out which syntax could work for all of them, but I found none yet.

henriquemoody:
The right thing to do here is to create a documentation for `__callStatic()` method which returns an instance of `Validator`.

alganet:
I Agree. Also, there is the `create` static method created alongside the doc blocks by its contributor: https://github.com/Respect/Validation/commit/5036617e3e7c50c5c6e8458c2a94bc336352c242

henriquemoody:
I will close this pull request and create another one about these discussion.

henriquemoody:
I created #225.

henriquemoody:
It was done in #229

PR #203, opened 2014-08-06
Efficient version of PrimeNumber.php

yamaszone:
Consider using this more efficient version of primality test with O(sqrt(n)) instead of existing O(n)


henriquemoody:
@yamaszone, can you tell us what this change covers that the current does not cover?

henriquemoody:
Closed in favor of #264.

PR #204, opened 2014-08-06
Corrected a data set for valid prime number

yamaszone:
Previous empty input ('') is not a valid data set.


henriquemoody:
I agree with you that `''` is not a valid data set, so, we have a bug right here, but removing the `''` from the data provider does not fix the problem.

PR #220, opened 2014-12-13
Updated Domain and Mac Address docs.

CommandZ-io:
Domain validator incorrectly stated that an ip would validate.  See `Respect\Validation\Rules\DomainTest`

Added second valid Mac Address format demonstration using `-`.  See `Respect\Validation\Rules\MacAddressTest`

Slight grammatical fix on Domain Doc


henriquemoody:
Thanks

Issue #208, opened 2014-09-11
AbstractRule getName() method bug


blumfontein:
Hello!

Thanx for great library! Recently I wanted to use Respect/Validation in my project and noticed a bug. The $rule->getName() returns nothing. Lets see in the code.

``` php
// Respect/Validation/Rules/AbstractRule.php
public function getName()
{
    if (empty($this->name))
        preg_replace('/.*\\\/', '', get_class($this));
    return $this->name;
}
```

Does line 2 of the method miss some returning or assignment statement? I guess it should be something like this.

```php
public function getName()
{
    if (empty($this->name))
        $this->name = lcfirst(preg_replace('/.*\\\/', '', get_class($this)));
    return $this->name;
}
```

Thanx

henriquemoody:
Seems you have found dead code! :tada:

Whatever, a rule without a defined name must return `NULL` any way, take a look on #213

Issue #213, opened 2014-10-02
setName() is working incorrectly


videde:
# Issue
When I tried test to the code 
```php
$usernameValidator = v::alnum()->noWhitespace()->length(1,15)->setName('username');
try {
    $usernameValidator->assert('really messed up screen#name');
} catch(\InvalidArgumentException $e) {
    var_dump($e->findMessages(array('alnum', 'length', 'noWhitespace')));
}
```
Result is
```
array(3) {
  ["alnum"]=>
  string(79) ""really messed up screen#name" must contain only letters (a-z) and digits (0-9)"
  ["length"]=>
  string(66) ""really messed up screen#name" must have a length between 1 and 15"
  ["noWhitespace"]=>
  string(58) ""really messed up screen#name" must not contain whitespace"
}
```
but I want expected result is when use ```setName('username')``` function
```
array(3) {
  ["alnum"]=>
  string(57) "username must contain only letters (a-z) and digits (0-9)"
  ["length"]=>
  string(44) "username must have a length between 1 and 15"
  ["noWhitespace"]=>
  string(36) "username must not contain whitespace"
}
```
# Reason
I find down why ```setName()``` function is working incorrectly. I use ```var_dump($usernameValidator)``` to reveal the variable ```$usernameValidator``` and there is result
```php
object(Respect\Validation\Validator)#2 (3) {
["rules":protected]=>
  array(3) {
    ["000000002989de9200000000401eb328"]=>
    object(Respect\Validation\Rules\Alnum)#4 (3) {
    ["additionalChars"]=>
      string(0) ""
    ["name":protected]=>
      NULL
      ["template":protected]=>
      NULL
    }
    ["000000002989de9300000000401eb328"]=>
    object(Respect\Validation\Rules\NoWhitespace)#5 (2) {
    ["name":protected]=>
      NULL
      ["template":protected]=>
      NULL
    }
    ["000000002989de9000000000401eb328"]=>
    object(Respect\Validation\Rules\Length)#6 (5) {
    ["minValue"]=>
      int(1)
      ["maxValue"]=>
      int(15)
      ["inclusive"]=>
      bool(true)
      ["name":protected]=>
      NULL
      ["template":protected]=>
      NULL
    }
  }
  ["name":protected]=>
  string(8) "username"
["template":protected]=>
  NULL
}
```
[...]

PR #239, opened 2015-01-10
A rule without name must resturn NULL in getName()

LanceFryar:
I fixed this little bug #208

### DIFF
--- a/library/Rules/AbstractRule.php
+++ b/library/Rules/AbstractRule.php
@@ -46,7 +46,7 @@ abstract class AbstractRule implements Validatable
     public function getName()
     {
         if (empty($this->name)) {
-            preg_replace('/.*\\\/', '', get_class($this));
+            return null;
         }
 
         return $this->name;


henriquemoody:
Thanks, but actually I'm not so sure about this.

If we take a look at the [documentation](https://github.com/Respect/Validation#getting-messages) it must define the name according to the class name (the rule name in itself) but this will conflict with what was made in #213.

Any thoughts, @Respect/troubleshooters?

library/Rules/AbstractRule.php
```diff
@@ -46,7 +46,7 @@ public function check($input)
     public function getName()
     {
         if (empty($this->name)) {
-            preg_replace('/.*\\\/', '', get_class($this));
+            return null;
         }
```
henriquemoody: @LanceFryar, just remove this entire block because if `$this->name` is empty return it is enough.

henriquemoody:
Closed by #261

Issue #235, opened 2015-01-07
Different results for validate(), assert() and check()


micrarior:
Am I doing something wrong? 
Why assert and check don't throw exception?
```php
use Respect\Validation\Validator as v;

$array = array('foo' => '');
var_dump(v::key('foo', v::positive())->validate($array)); // false
var_dump(v::key('foo', v::positive())->assert($array)); // true
var_dump(v::key('foo', v::positive())->check($array)); // true
```

henriquemoody:
Actually the but, as I see is because you should have the same result for `validate()`, `assert()` and `check()`, but all values are optional, take a look in the documentation: https://github.com/Respect/Validation#input-optional

micrarior:
> https://github.com/Respect/Validation#input-optional

Sorry, I din't notice this section.
But that means the bug is in the validate method

henriquemoody:
You're right, it is.

bmorg:
The same bug can be reproduced for other validators, e.g. the `numeric` validator:

```php
$validator = v::numeric();
$val = '';

var_dump($validator->validate($val)); // false
var_dump($validator->assert($val)); // true
var_dump($validator->check($val)); // true
```

I am not sure if this is tested for yet.

henriquemoody:
The fix for `key()`, `call()` and `attribute()` rules are available on `0.7.5` and `0.8.1` versions. We still need to fix the behavior on other rules.

henriquemoody:
Will be fixed on version 1.0, see [changelog](https://github.com/Respect/Validation/blob/311891b/CHANGELOG.md#100---2015-10-24) for more info about the version.

To fix this issue now all rules are mandatory, but you can use [optional()](https://github.com/Respect/Validation/blob/311891b/docs/Optional.md) rule to handle optional values, which now is `null` and an empty string by default.

Issue #196, opened 2014-07-15
notEmpty "0" is true


maboiteaspam:
hi,

php ver

```
PHP 5.5.9-1ubuntu4.3 (cli) (built: Jul  7 2014 16:34:16) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.3, Copyright (c) 1999-2014, by Zend Technologies
```

demo

```
        var_dump(Validator::string()->notEmpty()->validate('0') );
        var_dump(Validator::string()->notEmpty()->validate('food') );
```

returns

```
bool(false)
bool(true)
```

expected

```
bool(true)
bool(true)
```

fix

```
namespace Respect\Validation\Rules;

class NotEmpty extends AbstractRule
{
    public function validate($input)
    {
        if (is_string($input)) {
            $input = trim($input);
            return $input!=="";
        }

        return !empty($input);
    }
}
```

henriquemoody:
Will be fixed on version 1.0, see [changelog](https://github.com/Respect/Validation/blob/311891b/CHANGELOG.md#100---2015-10-24) for more info about the version.

To fix this issue now all rules are mandatory, but you can use [optional()](https://github.com/Respect/Validation/blob/311891b/docs/Optional.md) rule to handle optional values, which now is `null` and an empty string by default.

Issue #205, opened 2014-08-13
Allow null values to pass optional validation rules


zanderbaldwin:
Currently `null` values do not pass optional validation rules (by not adding `notEmpty()`), for example:

```php
<?php

    $rule = v::string()->length(28,28);

    // Passes validation:
    $rule->validate('');

    // Does not pass validation:
    $rule->validate(null);
```

I'm using this validation library for an API library, not form submission, so accepting `null` values is important. It would be nice if this could be worked into a future version.

henriquemoody:
Will be fixed on version 1.0, see [changelog](https://github.com/Respect/Validation/blob/311891b/CHANGELOG.md#100---2015-10-24) for more info about the version.

To fix this issue now all rules are mandatory, but you can use [optional()](https://github.com/Respect/Validation/blob/311891b/docs/Optional.md) rule to handle optional values, which now is `null` and an empty string by default.

PR #245, opened 2015-01-16
Fixes #244, #154, #196, #173, #235 and #239

### LINKED ISSUE #154: How to validate notEmpty bit value 0/1

### LINKED ISSUE #244: Int notempty should return true if 0
The validation of `v::int()->notEmpty(0)->validate()` returns false.

This is not intuitive.  0 is an integer, therefore it's not empty.  

Intuitively should work as follows:
```php
v::int()->notEmpty(0)->validate();    //true
v::int()->notEmpty('')->validate();    //false
v::int()->notEmpty('0')->validate(0);    //true
v::int()->notEmpty(null)->validate(0);    //false
```

alganet:
- Moves code from  __invoke on AbstractRule into NotEmpty.
  - Moves foreign type check to the chain.
  - Fixes tests using the chain instead of concrete instances.
  - Adds a parameter for strict type checking on NotEmpty. (no BC break)
  - Adds new tests for the BC behavior and new parameter.
  - Fixes AbstractRelated to allow empty referenced values.
  - Removes dead code found in #239 

Fixes #154, #196, #173 and #235.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1475,14 +1475,19 @@ See also:
 
   * [v::noneOf()](#vnoneofv1-v2-v3)
 
-#### v::notEmpty()
+#### v::notEmpty($jugglesTypes = true)
 
 Validates if the given input is not empty or in other words is input mandatory and
 required. This function also takes whitespace into account, use `noWhitespace()`
 if no spaces or linebreaks and other whitespace anywhere in the input is desired.
 
+
 ```php
 v::string()->notEmpty()->validate(''); //false
+v::string()->notEmpty()->validate('0'); //false
+
+// Disable type juggling
+v::string()->notEmpty(false)->validate('0'); //true
 ```
 
 Null values are empty:
--- a/README.md
+++ b/README.md
@@ -1475,7 +1475,8 @@ See also:
 
   * [v::noneOf()](#vnoneofv1-v2-v3)
 
-#### v::notEmpty($jugglesTypes = true)
+#### v::notEmpty()
+#### v::notEmpty($strict = false)
 
 Validates if the given input is not empty or in other words is input mandatory and
 required. This function also takes whitespace into account, use `noWhitespace()`
@@ -1486,8 +1487,8 @@ if no spaces or linebreaks and other whitespace anywhere in the input is desired
 v::string()->notEmpty()->validate(''); //false
 v::string()->notEmpty()->validate('0'); //false
 
-// Disable type juggling
-v::string()->notEmpty(false)->validate('0'); //true
+// Enable strict checking
+v::string()->notEmpty(true)->validate('0'); //true
 ```
 
 Null values are empty:
--- a/library/Rules/AbstractRelated.php
+++ b/library/Rules/AbstractRelated.php
@@ -66,10 +66,12 @@ abstract class AbstractRelated extends AbstractRule
     public function validate($input)
     {
         $hasReference = $this->hasReference($input);
+
         if ($this->mandatory && !$hasReference) {
             return false;
         }
 
-        return $this->decision('validate', $hasReference, $input);
+        return $this->decision('validate', $hasReference, $input) ||
+        $this->getReferenceValue($input) === '';
     }
 }
--- a/library/Rules/NotEmpty.php
+++ b/library/Rules/NotEmpty.php
@@ -3,12 +3,31 @@ namespace Respect\Validation\Rules;
 
 class NotEmpty extends AbstractRule
 {
+    /** Wheter to juggle PHP types **/
+    public $jugglesTypes = true;
+
+    public function __construct($jugglesTypes = true)
+    {
+        $this->jugglesTypes = $jugglesTypes;
+    }
+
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -1475,14 +1475,19 @@ See also:
 
   * [v::noneOf()](#vnoneofv1-v2-v3)
 
-#### v::notEmpty()
+#### v::notEmpty($jugglesTypes = true)
```
henriquemoody: You must update the links too [links](https://github.com/Respect/Validation#vnotempty)

henriquemoody:
@alganet, since you're doing this, cal you fix #235 too?

README.md
```diff
@@ -1475,14 +1475,19 @@ See also:
...
 required. This function also takes whitespace into account, use `noWhitespace()`
 if no spaces or linebreaks and other whitespace anywhere in the input is desired.
 
+
 ```php
 v::string()->notEmpty()->validate(''); //false
+v::string()->notEmpty()->validate('0'); //false
+
+// Disable type juggling
+v::string()->notEmpty(false)->validate('0'); //true
```
henriquemoody: I was expecting `v::int()->notEmpty(false)->validate(0)` to return `true` too, but it didn't happen

library/Rules/NotEmpty.php
```diff
@@ -3,12 +3,31 @@
...
+            return !empty($input);
         }
 
-        return !empty($input);
+        return '' !== $input;
+    }
+
+    public function __invoke($input)
+    {
+        return $this->validate($input);
```
henriquemoody: What you did here and on `AbstractRule` was great! Very clever :clap:

library/Rules/NotEmpty.php
```diff
@@ -3,12 +3,31 @@
...
+        $this->jugglesTypes = $jugglesTypes;
+    }
+
     public function validate($input)
     {
-        if (is_string($input)) {
-            $input = trim($input);
+        if (!is_string($input)) {
+            return !empty($input);
+        }
```
henriquemoody: Just an idea for the first major version: when the input is an array we can use `array_filter()` to remove empty values.

README.md
alganet: Do you think it should? That would make `notEmpty` a string-only rule.

alganet:
I've added the fix for #235. It was actually a bug on `AbstractRelated` (the base class for Key, Attribute, Each and indirect validators).

Guys, I'm not sure about the parameter on `NotEmpty`. I've changed it to a simpler name. My implementation only disallows quoted zeroes like `'  0'` and `"0"` (with or without whitespace), which seems to be the cause of the trouble.

henriquemoody:
Don't forget to update Validator docblock, too

README.md
```diff
@@ -1476,13 +1476,19 @@ See also:
   * [v::noneOf()](#vnoneofv1-v2-v3)
 
 #### v::notEmpty()
+#### v::notEmpty($strict = false)
```
henriquemoody: Not a blocker, but I guess `v::notEmpty(boolean $strict)` is better.

README.md
henriquemoody: My expectations in non-strict mode:

| Value             | NotEmpty |
| -------------     | ---------|
| `null`            | No       |
| `""`              | No       |
| `array()`         | Yes      |
| `"0"`             | Yes      |
| `0`               | Yes      |
| `"0.0"`           | Yes      |
| `0.0`             | Yes      |
| `true`            | Yes      |

Does it makes sense for you? Take a look on #205.

henriquemoody:
#235 is not fixed yet, I think to fix that we have to overwrite all possible `validate()` methods...

library/Rules/AbstractRelated.php
```diff
@@ -66,10 +66,12 @@ public function check($input)
...
     {
         $hasReference = $this->hasReference($input);
+
         if ($this->mandatory && !$hasReference) {
             return false;
         }
 
-        return $this->decision('validate', $hasReference, $input);
+        return $this->decision('validate', $hasReference, $input) ||
+        $this->getReferenceValue($input) === '';
```
malukenho: `empty()`

henriquemoody:
I will close this pull request because `NotEmpty` has a new feature and new features must be send against to `master` branch.
Also I toke some part of code on this pull request and I sent #298.

About the NotEmpty, I was wondering about this, I'm in favor of creating a new rule, called `NotBlank` which should have more or less the same behavior of our current `NotEmpty` and the `NotEmpty` rule should use `empty()` but since it's a major BC it's better we do this in the 1.0 version, IMO.

Any way, thanks you very much for this pull request, #298 is already available on `0.7.5` and `0.8.1` versions and on `master` branch as well.

PR #261, opened 2015-01-20
Remove dead code on AbstractRule

### LINKED ISSUE #208: AbstractRule getName() method bug

henriquemoody:
Fix #208

### DIFF
--- a/library/Rules/AbstractRule.php
+++ b/library/Rules/AbstractRule.php
@@ -45,10 +45,6 @@ abstract class AbstractRule implements Validatable
 
     public function getName()
     {
-        if (empty($this->name)) {
-            preg_replace('/.*\\\/', '', get_class($this));
-        }
-
         return $this->name;
     }
 


alganet:
:fire: :panda_face: :fire: you pandas are on fire!

PR #264, opened 2015-01-23
NumberPrimes.php optimization algorithm and performance.

kmilotxm:
An optimization algorithm and performance improvement was made. Below is a description of what was done.

Here, there is the input is numeric and less than or equal to 1, as the prime numbers begin starting the 2.
if(!is_numeric($input) || $input <= 1)
            return false;

After valid if it is even and different from 2. For, in addition to 2, there are no pairs prime numbers. In this case it is an even number, will not need to perform an iteration.
        if($input!= 2 && $input % 2 ==  0)
            return false;

Finally we conducted a loop, starting counting the two, because any number is divisible by 1 and excluding the number itself,
because every number is divisible by himself. Thus, if there is a single divider in this loop, the input is not prime. 
        for($i=2; $i < $input; $i++){
            if($input % $i == 0)
                return false;
        }


henriquemoody:
Thanks! Do you mind to squash these commits? 5 is a lot for this change.

henriquemoody:
Also, take a look on Scrutinizer [reports](https://scrutinizer-ci.com/g/Respect/Validation/inspections/61456c0e-8d14-4550-835b-f9ab9167c559/issues/files/library/Rules/PrimeNumber.php?status=new&orderField=path&order=asc)

kmilotxm:
No problem. How do you want me to squash the commits? in a single commit?

henriquemoody:
Yeah, a single commit is enough. :smile: 

PS.: Don't forget to fix the coding standards, too.

henriquemoody:
@kmilotxm, why did you close?

PR #273, opened 2015-01-30
composer.lock file should be committed to repo

AndyWendt:
See the [composer docs](https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file) and [this EngineYard.com blog post](https://blog.engineyard.com/2014/composer-its-all-about-the-lock-file) for the reasoning why the `.lock` file should be commited.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,1213 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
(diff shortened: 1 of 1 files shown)


henriquemoody:
I appreciate that, but this is not applied to libraries:

Some examples:

- https://github.com/symfony/symfony
- https://github.com/Seldaek/monolog
- https://github.com/zendframework/zf2

PR #274, opened 2015-01-30
Better place for documentations 

### LINKED ISSUE #259: Better place for documentations
I love our documentation but it's getting hard to keep it.
There are many instructions and documented rules and in one single file ([README.md](https://github.com/Respect/Validation/blob/master/README.md)).

I'm +1 for GitHub pages on this repository.

- [x] Slit "README.md" file by topics
- [x] Put docs into `docs` directory
- [x] Create `gh-pages` branch
- [x] Use [couscous](http://couscous.io/) to render documentation page
- [ ] Implements [lunrjs](http://lunrjs.com)

henriquemoody:
Fix #259, #296.

Closes #295.


williamespindola:
Cool. Certainly the documentation menu will list all the sessions.

henriquemoody:
Unfortunately looks like there is no way to make nested menus.

couscous.yml
```diff
@@ -0,0 +1,38 @@
...
+        main:
+            name: Table of contents
+            items:
+                home:
+                    text: Home
+                    relativeUrl: /
+                contributing:
+                    text: Contributing
+                    relativeUrl: CONTRIBUTING.html
+                documentation:
```
williamespindola: @henriquemoody maybe [something like this](https://github.com/myclabs/Work/blob/master/couscous.yml#L26)

henriquemoody: Yes, on `sections` you can have more than one menu.
I can do it if you like, the only problem is because the links will broken on GitHub. Do you think it worth?

henriquemoody: TBH, I don't like it, but I can live with.

henriquemoody: How about put all rules in the menu? It will create some duplication but would be cool.

williamespindola:
I think a great approach:

- Documentation
  - Feature Guide
  - Validators

Because has a lot of validators this on menu would be bad.

henriquemoody:
I will do it.

henriquemoody:
Cannot be done because couscous doesn't use ids on titles.

wesleyvicthor:
Am I the only one feeling a little bit uncomfortable on reading the docs with these colors palette?

henriquemoody:
I didn't like it either, but I saw just two templates for Couscous and a [basic template](https://github.com/CouscousPHP/Template-Basic). The Dark theme seems better than the Light, IMO.

Also, looks like Couscous needs a lot of improvements, I thought to contribute with the project but I cannot get involved very much for a while.

henriquemoody:
Any way, we can easily change the template, my main point here is about the content disposition.

Any thoughts?

PR #281, opened 2015-02-03
Replaced phpunit mocks of exceptions with concretions to fix hhvm issue

AndyWendt:
It looks like phpunit has an issue with mocking exceptions while running in HHVM: https://github.com/sebastianbergmann/phpunit-mock-objects/issues/207

Added a class that extends `AbstractNestedException` inside of `AbstractNestedExceptionTest` so that `AbstractNestedException` doesn't have to be mocked.


henriquemoody:
Build is ok now, looks like it was a temporary problem.

PR #286, opened 2015-02-05
Required improvements on "PostalCode" rule

### LINKED ISSUE #285: PostalCode/Country existence validation
Hello,

Is there anyway that we can check if a country has a PostalCode address?

I can see taht $postalCodes variable in PostalCode is not static, we need to create a PostalCode instance to do an 'in_array'.

Any idea about how to resolve this issue?

Regards,
Ghazi

henriquemoody:
- Turn rule case insensitive
- Use default pattern for countries who doesn't have postal code

Fix #285.


henriquemoody:
Wrong branch...

PR #288, opened 2015-02-06
Optimize PrimeNumber rule

l-x:
Decrease the number of max required iterations from `$input - 2` to `ceil(sqrt($input) / 2))`

### DIFF
--- a/library/Rules/PrimeNumber.php
+++ b/library/Rules/PrimeNumber.php
@@ -13,7 +13,7 @@ class PrimeNumber extends AbstractRule
             return false;
         }
 
-        for ($i = 2; $i < $input; $i++) {
+        for ($i = 3; $i <= ceil(sqrt($input)); $i+=2) {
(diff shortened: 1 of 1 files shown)


henriquemoody:
Have you idea of how much improvement we have with this change?

l-x:
Hi @henriquemoody,

the improvement heavily depends on the tested number. The highest impact will be on testing big prime numbers or big numbers with big prime factors.

If you test the number `982,451,653` (which actually is a prime) the old algorithm performs 982,451,651 iterations, the new algorithm only performs 15,672 iterations. You can calculate the required iterations (when the tested number is a prime number) yourself.

Old Algorithm: `$iterations = $number - 2`
New Algorithm: `$iterations =~ sqrt($number) / 2`

henriquemoody:
I see, it's a huge difference.

Thanks for contributing.

PR #289, opened 2015-02-06
Make Roman rule internally use Regex rule

l-x:
Just for consistency... :smile:

### DIFF
--- a/library/Rules/Roman.php
+++ b/library/Rules/Roman.php
@@ -1,10 +1,11 @@
 <?php
 namespace Respect\Validation\Rules;
 
-class Roman extends AbstractRule
+class Roman extends Regex
 {
-    public function validate($input)
+    public function __construct()
     {
-        return (boolean) preg_match('/^M{0,4}(CM|CD|D?C{0,3})(XC|XL|L?X{0,3})(IX|IV|V?I{0,3})$/', $input);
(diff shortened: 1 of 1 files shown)


henriquemoody:
Since you're doing that, can you extend Regex rule instead? (like [Yes](https://github.com/Respect/Validation/blob/master/library/Rules/Yes.php#L13) rule)

l-x:
I considered doing so. IMO the Roman Rule is not a special Regexp Rule, it just utilizes the Regexp Rule. Solving this by inheritance would break the OOP design principles.

henriquemoody:
The `Regex` rule has no specific method signature, only the same of `AbstractRule`; and a public variable `$regex` which, if needed, can be used (and useful) on `Roman` context as well, think about it:

```php
function validateAndExplain(Regex $regex, $input)
{
    if (!$regex->validate($input)) {
        echo $input.' is not valid for REGEXP '.$regex->regex.PHP_EOL;
    }
}

validateAndExplain($regex, $input);
validateAndExplain($roman, $input);
validateAndExplain($yes, $input);
validateAndExplain($not, $input);
```

I know, _Roman is not Regex_, but I think better we do this than create a hard coded instance of `Regex` inside `Roman::validate()` which would break OOD principles too and, IMHO, worse them use inheritance in this case.

Unfortunately we also have other design problems and we have plans to use a different design (but keeping the fluent API) in the first major release.

l-x:
Ok, I've updated the PR.

henriquemoody:
Thank for contributing! :smile:

Issue #283, opened 2015-02-04
[RFC] - Use of strcit EmailValidator


egulias:
Hi guys!
Awesome library!
I just checked the email validator you use (https://github.com/Respect/Validation/blob/master/library/Rules/Email.php#L8) and it happens to be php's `filter_var`.
[In this gist](https://gist.github.com/egulias/10989413) you can see where `filter_val` fails to validate some emails.
Also `filter_var` is incapable to validate UFT8 emails, which are in fact valid (RFCs [6530](http://tools.ietf.org/html/rfc6530#section-4.6), [6531](http://tools.ietf.org/html/rfc6531), [6532](http://tools.ietf.org/html/rfc6532) that allow for SMTPUTF8 international emails, also [Gmail started accepting them](http://gmailblog.blogspot.com.es/2014/08/a-first-step-toward-more-global-email.html))

The library [EmailValidator](https://github.com/egulias/EmailValidator) covers all that and includes warnings (this needs improvement, truth to be said) and dns check if wanted.

Would you consider integrating it? I can provide the PR if you are ok with it.

Cheers!

henriquemoody:
Thanks for open this discussion!

This sounds like something we really need to do, but I don't want to insert such dependency in validation like `v::email()` which is used very frequently.

Currently we have only 3 external dependencies ([Zend](https://github.com/zendframework/Component_ZendValidator), [Symfony](https://github.com/symfony/Validator) and [BAV](https://github.com/bav-php/bav)) but they are all optional and I don't want to make `v::email()` one of them, even I don't like the idea of repeat code.

Any way, this is just my opinion, how about you @Respect/troubleshooters?

PS.: Nice library @egulias!

alganet:
It is a nice library indeed, congratulations @egulias!

I agree with @henriquemoody. This is an important and popular validator, we should have it within our architecture. Our currently implementation sucks though, and we need to change it.

We already have a [Domain](https://github.com/Respect/Validation/blob/master/library/Rules/Domain.php) rule that could be improved to support the missing features (international domains, IPV6, etc). Our [Tld](https://github.com/Respect/Validation/blob/master/library/Rules/Tld.php) rule also needs improvements, it was designed before custom TLDs existed.

I'm also not sure how to express the several different options that could be configured (to allow or not IPs as domains, to allow or not local addresses, etc) in a clean way. This would be a problem for the library integration as well (how to consider its warnings).

egulias:
[...]

PR #299, opened 2015-02-15
#283 - Suggest egulias/emailvalidator and usage proposal

egulias:
I didn't found a way for specifying if you want to use either strict or not, so I left  it to whether is installed the lib or not.
Besides, instead of a simple regexp I left the original `filter_var`. I'm ok with either option, you decide.

### DIFF
--- a/library/Rules/Email.php
+++ b/library/Rules/Email.php
@@ -5,6 +5,14 @@ class Email extends AbstractRule
 {
     public function validate($input)
     {
-        return is_string($input) && filter_var($input, FILTER_VALIDATE_EMAIL);
+        if (class_exists('\Egulias\EmailValidator\EmailValidator')) {
+            $strictValidator = new \Egulias\EmailValidator\EmailValidator();
+
+            $validationResult = $strictValidator->isValid($input);
+        } else {
+            $validationResult = is_string($input) && filter_var($input, FILTER_VALIDATE_EMAIL);
+        }
+
+        return $validationResult;
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
         "ext-mbstring": "Multibyte String Functions",
         "malkusch/bav": "German bank account validation",
         "symfony/validator": "Use Symfony validator through Respect\\Validation",
-        "zendframework/zend-validator": "Use Zend Framework validator through Respect\\Validation"
+        "zendframework/zend-validator": "Use Zend Framework validator through Respect\\Validation",
+        "egulias/email-validator": "Strict (RFC compliant) email validation"
     },
     "autoload": {
         "psr-4": {


library/Rules/Email.php
```diff
@@ -5,6 +5,14 @@ class Email extends AbstractRule
...
     public function validate($input)
     {
-        return is_string($input) && filter_var($input, FILTER_VALIDATE_EMAIL);
+        if (class_exists('\Egulias\EmailValidator\EmailValidator')) {
+            $strictValidator = new \Egulias\EmailValidator\EmailValidator();
+
+            $validationResult = $strictValidator->isValid($input);
+        } else {
+            $validationResult = is_string($input) && filter_var($input, FILTER_VALIDATE_EMAIL);
+        }
```
henriquemoody: @egulias, can you avoid to use else, you can just return `$strictValidator->isValid($input);`

library/Rules/Email.php
```diff
@@ -5,6 +5,14 @@ class Email extends AbstractRule
 {
     public function validate($input)
     {
-        return is_string($input) && filter_var($input, FILTER_VALIDATE_EMAIL);
+        if (class_exists('\Egulias\EmailValidator\EmailValidator')) {
+            $strictValidator = new \Egulias\EmailValidator\EmailValidator();
```
henriquemoody: If you get this object by dependency injection you can test it, there is no test for this new behaviou on this pull request.

library/Rules/Email.php
egulias: The else part is for using `filter_var` as fall back. If I avoid the else part, the lib should be a hard dependency and not a soft one. I understood you wanted to be soft dep., right?

henriquemoody: You can do that:

```php
        return $strictValidator->isValid($input);
    }

    return is_string($input) && filter_var($input, FILTER_VALIDATE_EMAIL);
}

egulias: Yes, my bad. I wake up with that in mind.

henriquemoody:
Have you made any progress on this pull request, @egulias?

egulias:
Sorry for the delay @henriquemoody , it's been a busy week.

library/Rules/Email.php
egulias: Should I inject it here https://github.com/Respect/Validation/blob/0.8/library/Factory.php#L27 right?

henriquemoody: You can use a setter and a lazy load getter on the same class. This is necessary to test if Validation is using your library correctly.

henriquemoody:
No problem, @egulias, take your time. :)

egulias:
Thanks :). Hopefully this weekend I'll have time to work on it.

library/Rules/Email.php
henriquemoody: PS: When testing use a mock of `\Egulias\EmailValidator\EmailValidator` and pass it as dependency to the rule instead of use the original class, them you can assert if Validation is making the correct call to EmailValidator.

egulias: But you can't type hint, class may or may not exist.

henriquemoody: Actually you can.

The autoloader is just called when you create an instance of the object.

henriquemoody:
@egulias, I hope you don't mind. I've created #311 with your commit and some changes I made. I'd be glad if you take a look before I merge it.

egulias:
Thanks, I appreciate it!

2015-02-27 13:06 GMT+01:00 Henrique Moody <notifications@github.com>:

henriquemoody:
It's closed by #311.

PR #311, opened 2015-02-27
Use "egulias/emailvalidator" for email validation

henriquemoody:
Closes #299.

### DIFF
--- a/library/Rules/Email.php
+++ b/library/Rules/Email.php
@@ -1,10 +1,32 @@
 <?php
 namespace Respect\Validation\Rules;
 
+use Egulias\EmailValidator\EmailValidator;
+
 class Email extends AbstractRule
 {
+    public function __construct(EmailValidator $emailValidator = null)
+    {
+        $this->emailValidator = $emailValidator;
+    }
+
+    public function getEmailValidator()
+    {
+        if (!$this->emailValidator instanceof EmailValidator
+            && class_exists('Egulias\EmailValidator\EmailValidator')) {
+            $this->emailValidator = new EmailValidator();
+        }
+
+        return $this->emailValidator;
+    }
+
     public function validate($input)
     {
+        $emailValidator = $this->getEmailValidator();
+        if (null !== $emailValidator) {
+            return $emailValidator->isValid($input);
+        }
+
         return is_string($input) && filter_var($input, FILTER_VALIDATE_EMAIL);
     }
 }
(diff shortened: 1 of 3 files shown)


library/Rules/Email.php
```diff
@@ -1,10 +1,32 @@
 <?php
 namespace Respect\Validation\Rules;
 
+use Egulias\EmailValidator\EmailValidator;
```
egulias: What will happen if you install `Respect/Validation` with `--no-dev` ? `Egulias\EmailValidator\EmailValidator` will not exists and it will break.

library/Rules/Email.php
```diff
@@ -1,10 +1,32 @@
...
 class Email extends AbstractRule
 {
+    public function __construct(EmailValidator $emailValidator = null)
+    {
+        $this->emailValidator = $emailValidator;
+    }
+
+    public function getEmailValidator()
+    {
+        if (!$this->emailValidator instanceof EmailValidator
```
egulias: Here I think you can just ask for `null`. If not null, is already type hinted and the class exists.

tests/Rules/EmailTest.php
```diff
@@ -1,8 +1,85 @@
 <?php
 namespace Respect\Validation\Rules;
 
+function class_exists($className)
```
egulias: Nice trick :)

library/Rules/Email.php
henriquemoody: If you don't run `--no-dev` you cannot run PHPUnit either, I mean, you can run it, but there is no guarantee you're using the same PHPUnit we have.

There are some other dependencies yet, but, as I said before, doesn't matter the type hinting (or the use statement), the only problem you will get is if you create an instance of the object, until them PHP will not care about declarations.

You can have many `use` statement, PHP will only complain about them when you create an instance (when the autoloader is called).

library/Rules/Email.php
henriquemoody: Anyone can change any property of an object, even the private ones (using the reflection API), doing this we avoid other problems (it's kinda neurosis, I know).

tests/Rules/EmailTest.php
henriquemoody: :panda_face:

library/Rules/Email.php
egulias: Ok then, was just to lower the complexity.

egulias:
Looks good to me, thanks for taking the time to finish it  :)

henriquemoody:
Thanks for you feedback, its all on commit 13c3c92, now.

Issue #312, opened 2015-03-02
GB Postal code not working


ghazispeakserve:
Hello,

According the official UK documentation the regex for validating "GB" postal code have some issues. I have tried validating "

The link for the RegEx is here on page 6 https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/359448/4__Bulk_Data_Transfer_-_additional_validation_valid.pdf 

A correct UK Postal code is "PR1 9LY". It can be added to unit tests. The same postal code without the median (PR19LY) space is invalid with the government RegEx but valid with Respect RegEx. It should'nt.

henriquemoody:
This change is now available on versions `0.7.7` and `0.8.3` already available on Packagist.

henriquemoody:
Thanks for reporting it and also for the feedback on the pull request. :smile:

ghazispeakserve:
Thank you too. Great support :smiley:

PR #314, opened 2015-03-02
Update REGEX for postal code of "GB" country

henriquemoody:
Related to #312.


ghazispeakserve:
Did you notice that the latest trunk of the RegEx is inverted? In the document it is [A-Za-z]. It seems that you forgot the dash between the A and the Z.

library/Rules/PostalCode.php
```diff
@@ -53,7 +53,7 @@ class PostalCode extends Regex
         "FM" => "/^(\d{5})$/",
         "FO" => "/^(?:FO)*(\d{3})$/",
         "FR" => "/^(\d{5})$/",
-        "GB" => "/^(([A-Z]\d{2}[A-Z]{2})|([A-Z]\d{3}[A-Z]{2})|([A-Z]{2}\d{2}[A-Z]{2})|([A-Z]{2}\d{3}[A-Z]{2})|([A-Z]\d[A-Z]\d[A-Z]{2})|([A-Z]{2}\d[A-Z]\d[A-Z]{2})|(GIR0AA))$/",
+        "GB" => "/^([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([AZa-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9]?[A-Za-z])))) [0-9][AZa-z]{2})$/",
```
ghazispeakserve: Here it should be 

```php
([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([AZa-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9]?[A-Za-z])))) [0-9][A-Za-z]{2})
```

henriquemoody: I update, but there is another `AZa-z`, is it correct?

ghazispeakserve:
Thanks. Yes the correct "AZa-z" is in the middle of the RegEx.

library/Rules/PostalCode.php
```diff
@@ -53,7 +53,7 @@ class PostalCode extends Regex
         "FM" => "/^(\d{5})$/",
         "FO" => "/^(?:FO)*(\d{3})$/",
         "FR" => "/^(\d{5})$/",
-        "GB" => "/^(([A-Z]\d{2}[A-Z]{2})|([A-Z]\d{3}[A-Z]{2})|([A-Z]{2}\d{2}[A-Z]{2})|([A-Z]{2}\d{3}[A-Z]{2})|([A-Z]\d[A-Z]\d[A-Z]{2})|([A-Z]{2}\d[A-Z]\d[A-Z]{2})|(GIR0AA))$/",
+        "GB" => "/^([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([AZa-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9]?[A-Za-z])))) [0-9][A-Za-z]{2})$/",
```
henriquemoody: I know it is in the PDF you send, but `[AZa-z][0-9][A-Za-z]` doesn't make sense, `[A-Za-z][0-9][A-Za-z]`, are you sure I should keep this `[AZa-z][0-9][A-Za-z]`?

henriquemoody: @ghazispeakserve, I will wait for you feedback before I merge it.

ghazispeakserve: I agree that the ```[AZa-z]``` is non-sense. I have sent a request to the UK gov using contact form to ask about this. I am sure it is an issue. I don't have any concrete example to test with and tell if it is really an error.

By the way I used https://regex101.com/ to validate the RegEx. It may help.

henriquemoody: I'll update it, so.

Issue #119, opened 2013-02-25
How to localize validator main error message


liaolliso2012:
Now we know that we can localize all the validator messages through findMessages(), is there a way to localize the main error message from getMainMessage()?

yudina-m:
There is a 'translator' param in \Respect\Validation\Exceptions\ValidationException::getMainMessage method and this pull request https://github.com/Respect/Validation/pull/57

alganet:
In my experience, it is better to localize the messages outside the validator chain, when retrieving exceptions:

```php
try {
    v::string()->alnum('_')->length(1,15)->assert('alganet');
} catch (\Respect\Validation\Exceptions\NestedValidationExceptionInterface $e) { 
    $messages = $e->findMessages([
        'alnum' => myTranslator('Username must contain only letters, numbers and _.'),
        'length' => myTranslator('Username must have more than 1 and less than 15 chars.')
    ]);
}
```

You can also globally change exception message templates by rule:

```php
Respect\Validation\Exceptions\Alnum::$defaultTemplates = array(
        self::MODE_DEFAULT => array(
            self::STANDARD => '{{name}} deve conter apenas letras (a-z) e dígitos (0-9)',
            self::EXTRA => '{{name}} deve conter apenas letras (a-z), dígitos (0-9) e "{{additionalChars}}"',
        ),
        self::MODE_NEGATIVE => array(
            self::STANDARD => '{{name}} não deve conter letras (a-z) ou dígitos (0-9)',
            self::EXTRA => '{{name}} não deve conter letras (a-z), dígitos (0-9) ou "{{additionalChars}}"',
        ),
    );
```

meglio:
Should `$defaultTemplates` be added to the documentation as well?

henriquemoody:
I think it's too much exposure to how we work internally.

Of course anyone can hack as much as they want but that can make it hard when upgrading to a new version. I'd rather keep working on a better way to handle translations in the next version than changing the documentation for the time being

Issue #28, opened 2011-09-25
I18n


jeanpimentel:
Is there any plan to this?
How could I translate errors?

``` php
v::email()->assert('invalid@'); 
# "invalid@" must be valid email.
```

to

``` php
v::email()->assert('invalid@'); 
# "invalid@" deve ser um email válido.
```

PR #320, opened 2015-03-30
Some improvements including localization

maliayas:
I've implemented below:

1. Quick and a little dirty localization. Definitely it can be improved, but works fine in current state.
2. Make `between()`, `max()` and `min()` inclusive by default
3. Do not allow whitespace in `alnum()`, `alpha()` and `digit()`
4. Update README for above changes
5. Added Turkish translation (not all the messages are translated).

Localization example:

```php
v::setLocale('en_US');

try {
    v::int()->check("hello");

} catch ($exception) {
    echo $exception->getMainMessage(); // automatically localized message
}
```

If no locale is set, English is the default. So it won't break existing code.

\# 2 and # 3 however will break existing code. It would be good to get feedback from users if they agree with those changes.

Notes:

* I'm not able to update tests at the moment.
* I've based my changes onto 0.8 tag. This PR is through master and it seems it gonna give conflicts.

/cc #119, #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -1380,10 +1380,10 @@ since length accepts many types:
 v::arr()->length(1, 5)->validate(array('foo', 'bar')); //true
 ```
 
-A third parameter may be passed to validate the passed values inclusive:
+A third parameter may be passed to validate the passed values exclusive:
 
 ```php
-v::string()->length(1, 5, true)->validate('a'); //true
+v::string()->length(1, 5, false)->validate('a'); //false
 ```
 
 Message template for this validator includes `{{minValue}}` and `{{maxValue}}`.
--- a/README.md
+++ b/README.md
@@ -1414,14 +1414,14 @@ v::macAddress()->validate('af-AA-22-33-44-55'); //true
 ```
 
 #### v::max(mixed $maxValue)
-#### v::max(mixed $maxValue, boolean $inclusive = false)
+#### v::max(mixed $maxValue, boolean $inclusive = true)
 
 Validates if the input doesn't exceed the maximum value.
 
 ```php
 v::int()->max(15)->validate(20); //false
-v::int()->max(20)->validate(20); //false
-v::int()->max(20, true)->validate(20); //true
+v::int()->max(20)->validate(20); //true
+v::int()->max(20, false)->validate(20); //false
 ```
 
 Also accepts dates:
@@ -1437,8 +1437,8 @@ Also date intervals:
 v::date()->max('-18 years')->validate('1988-09-09'); //true
 ```
 
-`true` may be passed as a parameter to indicate that inclusive
-values must be used.
+`false` may be passed as a parameter to indicate that the max
+value is not inclusive.
 
 Message template for this validator includes `{{maxValue}}`.
 
@@ -1448,14 +1448,14 @@ See also:
   * [v::between()](#vbetweenmixed-start-mixed-end)
 
 #### v::min(mixed $minValue)
-#### v::min(mixed $minValue, boolean $inclusive = false)
+#### v::min(mixed $minValue, boolean $inclusive = true)
 
-Validates if the input is greater than the minimum value.
+Validates if the input is not less than the minimum value.
 
 ```php
 v::int()->min(15)->validate(5); //false
-v::int()->min(5)->validate(5); //false
-v::int()->min(5, true)->validate(5); //true
+v::int()->min(5)->validate(5); //true
+v::int()->min(5, false)->validate(5); //false
 ```
 
 Also accepts dates:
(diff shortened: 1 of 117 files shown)


augustohp:
Holy mother of god

That is **a lot work**! Thanks a lot for the input and time to develop those things! I :heart: the localization feature.

What do you think on splitting this *pull request** into smaller ones? I can help you do these if you don't, your commits are **great** so it should be easy:

1. Change rules to *inclusive* by default should be a pull request.
1. Removing *white spaces* from rules should be another pull request.
1. Localization should be in another.

I agree on the feedback of the rule changing, separating them would make it easier to merge localization stuff. :wink:

library/Messages/tr_TR.php
```diff
@@ -0,0 +1,954 @@
...
+use Respect\Validation\Exceptions\BetweenException;
+use Respect\Validation\Exceptions\DateException;
+use Respect\Validation\Exceptions\EqualsException;
+use Respect\Validation\Exceptions\IpException;
+use Respect\Validation\Exceptions\LengthException;
+use Respect\Validation\Exceptions\MaxException;
+use Respect\Validation\Exceptions\MinException;
+use Respect\Validation\Exceptions\NotEmptyException;
+
+$messages = array(
```
henriquemoody: You could just return the array instead of creating a variable it would be much more clean.

library/Exceptions/ValidationException.php
```diff
@@ -173,7 +172,7 @@ protected function buildTemplate()
     {
         $templateKey = $this->chooseTemplate();
 
-        return static::$defaultTemplates[$this->mode][$templateKey];
+        return @ Validator::$messages[basename(get_called_class())][$this->mode][$templateKey];
```
henriquemoody: This is also for the changes above: don't use shut up operator and make the verification instead, you can throw an exception, for example.

README.md
```diff
@@ -761,9 +769,9 @@ v::consonant()->validate('xkcd'); //true
 
 See also:
 
-  * [v::alnum()](#valnum)  - a-z0-9, empty or whitespace only
-  * [v::digit()](#vdigit) - 0-9, empty or whitespace only
-  * [v::alpha()](#valpha)  - a-Z, empty or whitespace only
+  * [v::alnum()](#valnum)  - a-z0-9 or empty
+  * [v::digit()](#vdigit) - 0-9 or empty
+  * [v::alpha()](#valpha)  - a-z or empty
```
henriquemoody: We appreciate these changes, but you should sent a separated pull request for this. 

Each pull request must be one only subject, so it can easily reviewed and it increases the chance of merge; if this change on README.md was send separately it was already merged.

henriquemoody:
It's really too much changes for just one pull request, I totally agree with @augustohp:

> Make between(), max() and min() inclusive by default

See #291, it's already available on `1.0` branch


> Do not allow whitespace in alnum(), alpha() and digit()

You could send a pull request just for this.


> Quick and a little dirty localization. Definitely it can be improved, but works fine in current state.
> Update README for above changes
> Added Turkish translation (not all the messages are translated).

TBH, I didn't like the idea. The original (english) messages must be in the exception classes, if we accept your changes we will not able to created custom rules because the exceptions messages will not be in the messages' files.
You could just translate the messages like `'This is a message` => 'Isto é uma mensagem'` in the messages files instead, for example.

library/Messages/tr_TR.php
augustohp: > You could just return the array instead of creating a variable it would be much more clean.

I would argue that the best approach would be keeping the messages inside the rule and translation them while creating the exceptions.

henriquemoody: :+1:

README.md
maliayas: README changes are not standalone. They're with PHP changes.

henriquemoody: Yes, the same file, but two kinds of changes: 
- Removing whitespace only
- Translation changes

library/Messages/tr_TR.php
maliayas: Probably you're right. On the other hand, I think usage should still be like this:

```php
v::setLocale('en_US');

try {
    v::int()->check("hello");

} catch ($exception) {
    echo $exception->getMainMessage(); // automatically localized message
}
```

maliayas:
> TBH, I didn't like the idea. The original (english) messages must be in the exception classes

I guess you are right. It was dfficult for me when I wanted to implement a "isbn" validatior. You may forget to write error messages in my scenario.

However localization in this PR is just a quick & dirty implementation as I've already stated :) There was short time and I had to do it. I'd be pretty happy if you grab `v::setLocale('en_US');` approach and do the actual implementation however you want.

I'd like to divide this PR into multiple PR's but that won't be easy. How about cherry-picking my useful commits (if any)? I may not find time to deal with dividing the PR.

Thanks for the interest.

henriquemoody:
@maliayas, I closed this because it's very old and need a rebase which will be really a PITA for you.

Also, I'm not sure if you're aware, but message localization is available since 2011, see #57.

However, thanks a lot for contributing! :panda_face:

PR #323, opened 2015-04-12
Date-validation fails when system is set to non-UTC timezone

qrazi:
The validation of a string with timezone information failed on systems with a different timezone set then UTC. Changing the formatting to the DateTime-object keeps to original Timezone instead of adjusting to the system's Timezone information.

Found this issue when running the testsuite with php.ini setting "date.timezone = Europe/Amsterdam". Expanded the DateTest to simulate different timezone's on the system.

### DIFF
--- a/library/Rules/Date.php
+++ b/library/Rules/Date.php
@@ -34,6 +34,6 @@ class Date extends AbstractRule
         $dateFromFormat = DateTime::createFromFormat($this->format, $input);
 
         return $dateFromFormat
-               && $input === date($this->format, $dateFromFormat->getTimestamp());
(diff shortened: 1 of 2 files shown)


henriquemoody:
@qrazi, I think its a [PATCH](http://semver.org/) change, can you please close this pull request and send another against 0.7 branch? You probably have to rebase to do this.

Also, can you create some `assertFalse()` tests to make clear what you just fixed here?

qrazi:
@henriquemoody OK, no problem. I should be about to do that tomorrow.

henriquemoody:
Take your time, @qrazi :smile:

PR #325, opened 2015-04-16
Adds Turkish Character check rule

ahmetgunes:


henriquemoody:
Looks like we have a problem with HHVM, and it doesn't looks like out problem.

henriquemoody:
@ahmetgunes, can you give be an usage example for this validator?

Looks like it's too much specific to be in the on our code base

ahmetgunes:
Hello @henriquemoody we are using these kind of checks for Turkish SSN controlls. We are applying different combinations for the given input (usually users firstname and surname) , if it includes Turkish Characters.

henriquemoody:
@ahmetgunes, I will accept your pull request. I did some search on the Internet and looks like this is a very common  problem.

To make it acceptable you need to send this pull request against the `master` branch because it's a new feature.

About the `Locale` namespace, you're in the right one, but `Validator` cannot see this namespace... You can create a `Character()` rule working like `Bank()` rule.

Also you need to create documentation for your rule and also add it do the `Validator` dockblock.

ahmetgunes:
@henriquemoody Thank you, I will fix those and will try to add more language specific rules next PR's

PR #327, opened 2015-04-16
Make AlwaysInvalid::assert('') fail

### LINKED ISSUE #326: AlwaysInvalid::assert() passes for empty input
As `AlwaysInvalid` uses the methods from its base class, where this case is allowed, it will let an empty input pass. Using `validate()` will fail, as it returns false.
`MostOfTheTimesInvalid` is not the same as `AlwaysInvalid`, and empty input is not a really rare edge case, so in my eyes this should be fixed to always fail.

I see a point in the general exception that empty input is passing, but for this special case I would rather stick to the very name of the validator instead of the general rule.

andreaswolf:
Resolves #326

### DIFF
--- a/library/Rules/AlwaysInvalid.php
+++ b/library/Rules/AlwaysInvalid.php
@@ -3,6 +3,11 @@ namespace Respect\Validation\Rules;
 
 class AlwaysInvalid extends AbstractRule
 {
+    public function __invoke($input)
+    {
+        return $this->validate($input);
+    }
+
     public function validate($input)
     {
         return false;
--- a/tests/Rules/AlwaysInvalidTest.php
+++ b/tests/Rules/AlwaysInvalidTest.php
@@ -1,0 +1,17 @@
+<?php
+namespace Respect\Validation\Rules;
+
+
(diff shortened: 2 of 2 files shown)


henriquemoody:
Thanks for contributing. Your commit was cherry-picked and it's available on **0.7.8** and **0.8.4** versions.

augustohp:
That is a feature of validation, right? All "empty" input should be valid, exception if `notEmpty` rule is explicitly used.

This was made with forms in mind, but if this gets changed it should be considered a BC break and not Patch version.

andreaswolf:
Am 17.04.2015 um 04:15 schrieb Augusto Pascutti:
> That is a feature of validation, right? All "empty" input should be
> valid, exception if |notEmpty| rule is explicitly used.
> 
> This was made with forms in mind, but if this gets changed it should be
> considered a BC break and not Patch version.

On the one hand, I agree – it changes the behaviour. OTOH, "always
invalid" should really *always* be invalid IMHO. Therefore, I considered
this a bug that should be fixed.

PR #328, opened 2015-04-16
travis: PHP 7.0 nightly added

TomasVotruba:
- `--dev` is by default for a long time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
   - hhvm
   - hhvm-nightly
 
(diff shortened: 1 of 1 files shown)


henriquemoody:
Hi @TomasVotruba, thanks for contributing but it's already on our master version: https://github.com/Respect/Validation/blob/master/.travis.yml

We keep 0.8 as default branch because it's the last stable version but looks like it's causing some confusing situations.

I think I must change it back to master, probably.

TomasVotruba:
> We keep 0.8 as default branch because it's the last stable version but looks like it's causing some confusing situations.

It is confusing. The developing branch should be available as first.
For releases tags are usually enough.

PR #336, opened 2015-04-22
New NotOption-rule to fix unexpected NotEmpty-rule

qrazi:
Added a not-optional-rule that behaves according to the documentation, and updated the documentation.

### DIFF
--- a/library/Exceptions/NotOptionalException.php
+++ b/library/Exceptions/NotOptionalException.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Respect\Validation\Exceptions;
+
+class NotOptionalException extends ValidationException
+{
+    const STANDARD = 0;
+    const NAMED = 1;
+    public static $defaultTemplates = array(
+        self::MODE_DEFAULT => array(
+            self::STANDARD => 'The value is not optional but required',
+            self::NAMED => '{{name}} is not optional but required',
+        ),
+        self::MODE_NEGATIVE => array(
+            self::STANDARD => 'The value must be an empty string',
+            self::NAMED => '{{name}} must be an empty string',
+        ),
+    );
+
+    public function chooseTemplate()
+    {
+        return $this->getName() == "" ? static::STANDARD : static::NAMED;
+    }
+}
--- a/library/RequiredValidatable.php
+++ b/library/RequiredValidatable.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Respect\Validation;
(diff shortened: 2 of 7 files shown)


library/RequiredValidatable.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Respect\Validation;
+
+/**
+ * Interface for Rules that should always have their validate() called.
+ */
+interface RequiredValidatable
```
henriquemoody: Good idea! :+1:

library/Exceptions/NotOptionalException.php
```diff
@@ -0,0 +1,24 @@
...
+    const STANDARD = 0;
+    const NAMED = 1;
+    public static $defaultTemplates = array(
+        self::MODE_DEFAULT => array(
+            self::STANDARD => 'The value is not optional but required',
+            self::NAMED => '{{name}} is not optional but required',
+        ),
+        self::MODE_NEGATIVE => array(
+            self::STANDARD => 'The value must be an empty string',
+            self::NAMED => '{{name}} must be an empty string',
```
henriquemoody: I like the `NAMED` idea.

library/Rules/AbstractRule.php
```diff
@@ -18,7 +19,7 @@ public function __construct()
 
     public function __invoke($input)
     {
-        return !is_a($this, __NAMESPACE__.'\\NotEmpty')
+        return !$this instanceof RequiredValidatable
             && $input === '' || $this->validate($input);
```
henriquemoody: In the future I would like to use the `NotOptional()` instead of this hard coded checking (`$input === ''`).

library/Exceptions/NotOptionalException.php
qrazi: It's actually based on the NotEmptyException.... :)

library/Rules/AbstractRule.php
qrazi: Your right, that makes a lot of sense.

henriquemoody: The problem is, we have to inject this rule here, otherwise we will create one more rule to each rule we have in a chain.

Probably we can do this after the API change.

henriquemoody:
As I already said, there are some things we can improve but since 1.0 is a work in progress we can change it later.

Thank you very much for contributing, @qrazi and I'm really sorry by my late.

PR #339, opened 2015-04-24
 Adds Turkish Character check rule

ahmetgunes:
Made changes according to feedbacks in https://github.com/Respect/Validation/pull/325 PR.

### DIFF
--- a/docs/Character.md
+++ b/docs/Character.md
@@ -1,0 +1,14 @@
+# Character
+
+- `v::character(string $countryCode)`
+
+Validates if an input includes language specific characters.
+
+```php
+v::character('TR')->validate("Mustafa Kemal Atatürk"); //true
+v::character('TR')->validate("Ahmet"); //false
+```
+
+These country codes are supported:
+
+ * "TR" (Turkey)
--- a/library/Rules/Locale/TurkishCharacter.php
+++ b/library/Rules/Locale/TurkishCharacter.php
@@ -1,0 +1,24 @@
+<?php
+namespace Respect\Validation\Rules\Locale;
+
+use Respect\Validation\Rules\AbstractRule;
+
+/**
+ * Validates if a text includes Turkish Characters.
+ *
+ * @author Ahmet Güneş <ahmetgunes@mail.com>
+ */
+class TurkishCharacter extends AbstractRule
+{
+    const TURKISH_CHARACTER_REGEX = '([çığöşüÇİĞÖŞÜ]+)';
+
+    /**
+     * Iterates over Turkish Characters array to detect turkishCharacters
+     * @param $input
+     * @return bool
+     */
+    public function validate($input)
+    {
+        return preg_match(self::TURKISH_CHARACTER_REGEX, $input) ? true : false;
+    }
+}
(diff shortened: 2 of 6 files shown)


henriquemoody:
@ahmetgunes, you missed CharacterException, that's why your build is failing.

docs/Character.md
```diff
@@ -0,0 +1,14 @@
+# Character
+
+- `v::character(string $countryCode)`
+
+Validates if an input includes language specific characters.
+
+```php
+v::character('TR')->validate("Mustafa Kemal Atatürk"); //true
+v::character('TR')->validate("Ahmet"); //false
+```
```
henriquemoody: I will prefer some [ISO 639](http://en.wikipedia.org/wiki/ISO_639) code ([ISO 639-1](http://en.wikipedia.org/wiki/ISO_639-1) and/or [ISO 639-5](http://en.wikipedia.org/wiki/ISO_639-5)], since the relationship between language and countries is m:n.

tests/Rules/Locale/TurkishCharacterTest.php
```diff
@@ -0,0 +1,19 @@
...
+namespace Respect\Validation\Rules\Locale;
+
+use Respect\Validation\Rules\Locale;
+
+class TurkishCharacterTest extends \PHPUnit_Framework_TestCase
+{
+    public function testShouldAcceptTurkish()
+    {
+        $rule = new TurkishCharacter;
+        $this->assertTrue($rule->validate('Mustafa Kemal ATATÜRK'));
```
henriquemoody: Could you use some data providers? Then you can test more situations, like camel case, for example.

library/Rules/Locale/TurkishCharacter.php
```diff
@@ -0,0 +1,24 @@
...
+    const TURKISH_CHARACTER_REGEX = '([çığöşüÇİĞÖŞÜ]+)';
+
+    /**
+     * Iterates over Turkish Characters array to detect turkishCharacters
+     * @param $input
+     * @return bool
+     */
+    public function validate($input)
+    {
+        return preg_match(self::TURKISH_CHARACTER_REGEX, $input) ? true : false;
```
henriquemoody: We have [AbstractRegexRule](https://github.com/Respect/Validation/blob/master/library/Rules/AbstractRegexRule.php), check it out!

henriquemoody:
I feel like I've put you and this code in a bad situation when I told you on #325 to use [Bank()](https://github.com/Respect/Validation/blob/master/library/Rules/Bank.php) rule as reference since it's much simpler, looks like it can be done as the same of [PostalCode()](https://github.com/Respect/Validation/blob/master/library/Rules/PostalCode.php) instead.

ahmetgunes:
@henriquemoody Hello, sorry for late answer. I will fix them and push again, as soon as I can. Thank you.

henriquemoody:
Take your time @ahmetgunes.

BTW, I think it's better you create a pull request on [Respect/ValidationLocale](https://github.com/Respect/ValidationLocale) repository instead of this one.

henriquemoody:
Hey @ahmetgunes, is there any progress on this pull request?

ahmetgunes:
@henriquemoody Sorry for the late response. I will close the pull request now. I want to open a new one when it's ready.

PR #342, opened 2015-04-24
Date-validation fails when system is set to non-UTC timezone

qrazi:
The creation of a DateTime-object from the input-string keeps the given timezone information. However, in the comparison the DateTime-object is
first output to a timestamp, which is then converted to a string with the
date()-function. But because a timestamp does not include timezone
information, date() will assume the system's timezone. So if the system
timezone set in the PHP settings is UTC, a string with another timezone,
e.g. 2015-04-24T21:11:00+02:00, will fail to be validated. The result from
the date()-function in this example is 2015-04-24T19:11:00+00:00, which is
a different string then the input.
The DateTime-class has also an option to create a string from a format,
DateTime->format(). So it is possible to skip the use of the
date()-function with the added benefit that using format() does have
access to the timezone information and thus produces the expected
2015-04-24T21:11:00+02:00 to compare with the given input.

This commit changes the Rule to accommodate this and expands the tests. If
these added tests are run before applying the fix to the Date-rule, the
tests will fail.

### DIFF
--- a/tests/library/Respect/Validation/Rules/DateTest.php
+++ b/tests/library/Respect/Validation/Rules/DateTest.php
@@ -72,5 +72,35 @@ class DateTest extends \PHPUnit_Framework_TestCase
         $this->dateValidator = new Date('r');
         $this->assertTrue($this->dateValidator->assert('Thu, 29 Dec 2005 01:02:03 +0000'));
     }
+
+    /**
+     * Test that datetime strings with timezone information are valid independent on the system's timezone setting.
+     *
+     * @param string $systemTimezone
+     * @param string $input
+     * @dataProvider providerForDateTimeTimezoneStrings
+     */
+    public function testDateTimeSystemTimezoneIndependent($systemTimezone, $format, $input)
+    {
+        date_default_timezone_set($systemTimezone);
+        $this->dateValidator = new Date($format);
+        $this->assertTrue($this->dateValidator->assert($input));
+    }
+
+    /**
+     *
+     * @return array
+     */
+    public function providerForDateTimeTimezoneStrings(){
+        return array(
+                array('UTC', 'c', '2005-12-30T01:02:03+01:00',),
+                array('UTC', 'c', '2004-02-12T15:19:21+00:00',),
+                array('UTC', 'r', 'Thu, 29 Dec 2005 01:02:03 +0000',),
+                array('Europe/Amsterdam', 'c', '2005-12-30T01:02:03+01:00',),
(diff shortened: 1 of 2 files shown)


tests/library/Respect/Validation/Rules/DateTest.php
```diff
@@ -72,5 +72,20 @@ public function testDateTimeExceptionalFormatsThatShouldBeValid()
...
+        date_default_timezone_set('UTC');
+        $this->dateValidator = new Date('c');
+        $this->assertTrue($this->dateValidator->assert('2005-12-30T01:02:03+01:00'));
+        $this->dateValidator = new Date('c');
+        $this->assertTrue($this->dateValidator->assert('2004-02-12T16:19:21+01:00'));
+        date_default_timezone_set('Europe/Amsterdam');
+        $this->dateValidator = new Date('c');
+        $this->assertTrue($this->dateValidator->assert('2005-12-30T01:02:03+01:00'));
+        $this->dateValidator = new Date('c');
+        $this->assertTrue($this->dateValidator->assert('2004-02-12T16:19:21+01:00'));
```
henriquemoody: Can you create some `assertFalse()` tests to make clear what you just fixed here?

Also, could you please use dataProviders instead?

qrazi: @henriquemoody I'm not totally sure how an assertFalse would make it more clear what I fixed? Maybe little pointer in the right direction? Dataprovider I will implement now... :)

henriquemoody:
I'm sorry to bother you again, but could you [squash](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/) these commits at once? There are 4  (four) commits for this change, 2 (two) of them are merges.

henriquemoody:
What happens @qrazi ?

qrazi:
Messing to much with Git.... :-/ I'm trying to figure out why the rebase
HEAD~4 shows 5 commits, but three are not mine, even though git log shows
last four commits are mine....

henriquemoody:
Try to create a new branch, based on 0.7, then just `git cherry-pick ec80ab258b39fe06216a45eb2872a83f68bd7648`, I think it's going to work.

qrazi:
Ok, I think I have cleaned up my branch now... But because I pushed a hard
reset earlier, the pull request was closed. I think I have to open a new
one? Perhaps have a look at my current 0.7 branch before I open a new PR?

By the way, sorry for the inconvenience. And I appreciate your patience! :)

henriquemoody:
Thanks for contributing!

Issue #346, opened 2015-04-30
noWhitespace validator causes warning on array input


bmorg:
Using the `noWhitespace` validator on an array produces a warning (because of the `preg_match` usage):

```php
v::digit()->noWhitespace()->assert(array());
```

The warning:

```
PHP Warning:  preg_match() expects parameter 2 to be string, array given in /.../respect/validation/library/Rules/NoWhitespace.php on line 8
```

Strangely the warning is not shown when using `validate()` instead of `assert()`.

henriquemoody:
This was fixed on versions [0.8.7](https://github.com/Respect/Validation/releases/tag/0.8.7) and [0.9.1](https://github.com/Respect/Validation/releases/tag/0.9.1).

bmorg:
Thanks for the quick fix!

henriquemoody:
Kudos to @augustohp :beer:

PR #347, opened 2015-05-03
Prevents noWithespace from failing validating arrays

augustohp:
A warning was thrown because of the preg_match call, as not every input
is a string we prevent calls from arrays going further into checking.

Issue #346

### DIFF
--- a/library/Rules/NoWhitespace.php
+++ b/library/Rules/NoWhitespace.php
@@ -5,6 +5,14 @@ class NoWhitespace extends AbstractRule
 {
     public function validate($input)
     {
-        return is_null($input) || !preg_match('#\s#', $input);
+        if (is_null($input)) {
+            return true;
+        }
+
+        if (false === is_scalar($input)) {
+            return false;
+        }
+
+        return !preg_match('#\s#', $input);
     }
 }
--- a/tests/Rules/NoWhitespaceTest.php
+++ b/tests/Rules/NoWhitespaceTest.php
@@ -42,6 +42,7 @@ class NoWhitespaceTest extends \PHPUnit_Framework_TestCase
     {
         return array(
             array(''),
+            array(null),
             array(0),
             array('wpoiur'),
             array('Foo'),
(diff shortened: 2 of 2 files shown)


library/Rules/NoWhitespace.php
```diff
@@ -5,6 +5,10 @@ class NoWhitespace extends AbstractRule
 {
     public function validate($input)
     {
+        if (is_array($input)) {
```
henriquemoody: I would use `!is_scalar()` instead of `is_array()`. What you think?

henriquemoody: I guess the `is_null()` below is there to cover optional input only but let's keep it in order to avoid BC breaks.
The `is_scalar()` function does not consider NULL to be scalar.

Even with this fix the rule will fail when raise errors with non-scalar values.

library/Rules/NoWhitespace.php
```diff
@@ -5,6 +5,10 @@ class NoWhitespace extends AbstractRule
...
+            return false;
+        }
+
         return is_null($input) || !preg_match('#\s#', $input);
```
henriquemoody: I guess it's better to use `(is_scalar($input) && preg_match(..` instead of the checking above. What you think?

augustohp: You are right, I will make the change and update this pull request.

augustohp: > I guess it's better to use (is_scalar($input) && preg_match(.. instead of the checking above. What you think?

I think preserving behavior is better, I didn't quite understood you comment before. I am sorry.
Changing the rule not to accept `null` breaks BC which is never a good thing without a good reason.

henriquemoody: Actually what I did mean was to just replace the `preg_match(..` for `(is_scalar($input) && preg_match(..`.

library/Rules/NoWhitespace.php
```diff
@@ -5,6 +5,10 @@ class NoWhitespace extends AbstractRule
 {
     public function validate($input)
     {
+        if (false === is_scalar($input)) {
```
henriquemoody: That's the problem, with this change when `$input` is `null` it returns `false` instead of `true`, which is the current behaviour. That's because `null` is not a scalar type.

henriquemoody:
Thanks, man!

PR #355, opened 2015-06-06
Pin "symfony/validator" version on 2.6.x

henriquemoody:


composer.json
```diff
@@ -16,7 +16,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
-        "symfony/validator": "~2.6",
+        "symfony/validator": "~2.6.9",
```
augustohp: Any reason for this change?

henriquemoody: I'm sorry, I should have explain that on the commit message, at least.

The tests was failing:

```
1) Respect\Validation\Rules\SfTest::testValidationWithAnExistingValidationConstraint
The Symfony\Component\Validator\Validator\RecursiveValidator::validateValue method is deprecated in version 2.5 and will be removed in version 3.0. Use the Symfony\Component\Validator\Validator\ValidatorInterface::validate method instead.
```

I didn't have enough time to fix this issue.

Issue #356, opened 2015-06-10
Filesize and FileType Validator?


joeporterme:
Sorry, ill preface this by saying im a newb and finally have this working after some hours of messing around.  i have my validations working but i have a file input in my form.  Not sure if i am missing it or not but is there a File Type Validator and File size? 

Example: Making sure it is a jpg or png file that is less than 1MB being inputed.

henriquemoody:
Hey, @joeporterme, unfortunately these rules do not exist. I will not have time for create them so soon, but if you want, please send a pull request, then me or someone else can review and merge.

maxw3st:
Not sure what language you're using but, in Java this would take two validation steps on  an input field. 
String FileInput;
double filebytes = File.length( FileInput );
double fileMB = ( filebytes / 1024 ) / 1024
if (FileInput == *.jpg || *.png) // checks for a file type that is either jpg or png
    if ( fileMB < 1 );  // checks to see if the file is less than 1MB in size

joeporterme:
Yeah. I have it done in php right now. Just didn't know if it came with RV. Be nice to have all in one :). Thanks for the replies guys.




On Sat, Jun 13, 2015 at 11:57 AM -0700, "Max West" <notifications@github.com<mailto:notifications@github.com>> wrote:


Not sure what language you're using but, in Java this would take two validation steps on an input field.
String FileInput;
double filebytes = File.length( FileInput );
double fileMB = ( filebytes / 1024 ) / 1024
if (FileInput == *.jpg || *.png) // checks for a file type that is either jpg or png
if ( fileMB < 1 ); // checks to see if the file is less than 1MB in size

maxw3st:
It looks like the only real difference between Java and PHP is that in PHP the method is filesize() instead of File.length. Both return a value in bytes, so divide that returned value by 1024 twice. Also, you can put both conditions into one line:

if ( ( $InputFile == *.jpg || $InputFile == *.png ) && ( filesize( $InputFile ) < 1000000 ) )

joeporterme:
This is how i have it working for myself in case anybody is wondering

if (!empty($_FILES['image']['type'])) :
			//This is the directory where images will be saved 
			$target = "/var/www/pathtosebsite/httpdocs/img/"; 
			$target = $target . time() . basename( $_FILES['image']['name']); 
			$allowed = array("image/jpeg", "image/png");
			$file_type = $_FILES['image']['type'];
			$filename  = basename($_FILES['image']['name']);
[...]

PR #359, opened 2015-06-16
Create "Size" rule

henriquemoody:
- `v::size(string $minSize)`
- `v::size(string $minSize, string $maxSize)`
- `v::size(null, string $maxSize)`

Validates file sizes:

```php
v::size('1KB')->validate($filename); // Must have at least 1KB size
v::size('1MB', '2MB')->validate($filename); // Must have the size between 1MB and 2MB
v::size(null, '1GB')->validate($filename); // Must not be greater than 1GB
```

Sizes are not case-sensitive and the accepted values are:

- B
- KB
- MB
- GB
- TB
- PB
- EB
- ZB
- YB

This validator will consider `SplFileInfo` instances, like:

```php
$fileInfo = new SplFileInfo($filename);
v::size('1.5mb')->validate($fileInfo); // Will return true or false
```

Message template for this validator includes `{{minSize}}` and `{{maxSize}}`.

***
Related to #356.


henriquemoody:
Closed by cc6da3a.

PR #360, opened 2015-06-16
Create "Extension" rule

henriquemoody:
- `v::extension(string $extension)`

Validates if the file extension matches the expected one:

```php
v::extension('png')->validate('image.png'); //true
```

This rule is case-sensitive.

***
Updates required PHP version to 5.3.6 since `SplFileInfo::getExtension()` was created on this version.

Related to #356.


henriquemoody:
Closed by 96f4fc2.

PR #361, opened 2015-06-16
Create "Mimetype" rule

henriquemoody:
- `v::mimetype(string $mimetype)`

Validates if the file mimetype matches the expected one:

```php
v::mimetype('image/png')->validate('image.png'); //true
```

This rule is case-sensitive and requires [fileinfo](http://php.net/fileinfo) PHP extension.

***
Related to #356.


henriquemoody:
Closed by fb74237

PR #365, opened 2015-06-30
AbstractRelated should set names for child rule

### LINKED ISSUE #86: Why are validation shown against value rather than field name?

henriquemoody:
A code like this:
```php
try {
    v::key('username', v::alnum()->noWhitespace()->length(4,22))->check($_POST);
} catch (ValidationExceptionInterface $exception) {
    echo $exception->getMainMessage().PHP_EOL;
}
```

Should display messages like this:

```
username must not contain whitespace
```

Fix #86

### DIFF
--- a/library/Rules/AbstractRelated.php
+++ b/library/Rules/AbstractRelated.php
@@ -26,10 +26,22 @@ abstract class AbstractRelated extends AbstractRule
 
     public function __construct($reference, Validatable $validator = null, $mandatory = true)
     {
-        $this->setName($reference);
         $this->reference = $reference;
         $this->validator = $validator;
         $this->mandatory = $mandatory;
+
+        $this->setName($reference);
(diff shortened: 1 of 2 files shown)


henriquemoody:
I'm not sure if it's a bug fix or a behaviour change. 

If it is really a bug fix, it's necessary to cherry-pick this commit to 0.8 and 0.9 branches, then create a PATCH version.

Issue #354, opened 2015-06-02
New Concrete API


henriquemoody:
Hey, guys! The new Concrete API for Respect\Validation is almost ready then we can release our first major version (see #258). This issue was created only just to explain and discuss about what I'm working on.

There will be no changes on our Fluent API, I took care to never break the agility we have creating validation chains.

First of all I wanna be clear about the comments below, Validation is, IMHO, the best validation library ever written in PHP and I always say, @alganet is a genius for creating such an amazing library! The reason I wanna create a new Concrete API is simple, the requirements changed since the library was created, eventually code rusts and that's the normal cycle of software development.
### All rules depend on `AbstractRule`

Yes, we have the `Validatable` interface, but let's be honest, we never create a rule from scratch and instead we just extend `AbstractRule` to create new rules.

That's a problem because all Rules are based on an implementation, if we have to change `AbstractRule` all rules will be changed as well.
### `Validatable` interface has too many methods

That's the reason why we extend `AbstractRule` instead, this interface has 7 methods and all rule must do the same in some of them.

The methods in `Validatable` are really necessary to do what Validation does, I fell like all methods where created really by necessity:
1. `assert()`
2. `check($input)`
3. `getName()`
4. `reportError($input, array $relatedExceptions = [])`
5. `setName($name)`
6. `setTemplate($template)`
7. `validate($input)`
### Decentralized validation

Each rule performs the validation to the input with no intermediate and this is a problem we are facing for more than one year (see #205 and #235).

Since we cannot intercept the input to check if it is or isn't optional, the only way to check optional inputs on all rules we have is changing all rules we have.

Currently the methods `assert()`, `check()` and `__invoke()` accept optional inputs as valid but the `validate()` method doesn't. Each rule has its own way to validate the input, so it can or cannot accept optional values and interpret whatever input as optional or valid.
### NotEmpty rule

That's rule as optional value checking has been a PITA for a long time (see #244, #154, #196, #173). Also the name _Empty_ should imply on using only the PHP `empty()` language constructor but it doesn't do that.
### Nested exceptions

Also a problem. When using `check()` method it works like a charm, but wen you use `assert()` with `key()`, for example, it doesn't work properly (see #179).
I've tried several times to fix this issue on the current code-base but the fixes were always a poor workaround. Of course, working with recursion is almost always ugly, but I think we can do better on the new Concrete API.
### Nested exception messages

[...]

PR #367, opened 2015-07-01
Allow `null` to pass nullable rules

maliayas:
I think the following is an inconsistency:

    v::someRule()->validate(''); // passes
    v::someRule()->validate(null); // doesn't pass

This commit fixes this.


maliayas:
Tests failed. Is my approach wrong?

augustohp:
> Tests failed. Is my approach wrong?

You will see that all tests which failed were against that particular case. This behavior has to do with optional input.

If you think it should be different, give a ping about it on #354

PR #368, opened 2015-07-04
New generic tlds

LucasKauz:
Source: https://www.icann.org/resources/pages/tlds-2012-02-25-en


henriquemoody:
I'm ok with that, but I remember some conversation with @alganet and if I'm not wrong he is against that. @alganet?

alganet:
I'm actually OK with this change!

PR #370, opened 2015-07-12
Prevents vim swap files to be found as rules

augustohp:
Exception testing searched for all files inside the rules directory and
tried to come up with a class name to it. VIM swap files always start
with a dot (since they are hidden files) and fail to be a valid class
name, provoking an error while testing.

Redesign the tests to use that "rule list" as a data provider instead
of being an algorithmn inside a test. Getting rid of the itertion inside
the test we can test all exception properties more easily while being
able to fail just one specific test (for an especific exception rule)
instead of preventing any more tests on exceptions from being executed.

While the test has fewer lines now, coverage and asserts are the same as
before.

### DIFF
--- a/tests/Exceptions/CheckExceptionsTest.php
+++ b/tests/Exceptions/CheckExceptionsTest.php
@@ -22,7 +22,7 @@ class CheckExceptionsTest extends \PHPUnit_Framework_TestCase
         'Vowels',
     );
 
-    private function getAllRuleNames()
+    public function provideListOfRuleNames()
     {
         $rulesDirectory = __DIR__.'/../../library/Rules';
         $rulesDirectoryIterator = new DirectoryIterator($rulesDirectory);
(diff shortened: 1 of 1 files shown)


alganet:
Looks ok. I'm not sure what is a vim swap file though. Are there any specific steps to test this change?

augustohp:
> Looks ok. I'm not sure what is a vim swap file though. Are there any specific steps to test this change?

1. Edit a file on VIM, inside the `Rules` namespace, while *swap-files* are enabled (this should be default behavior).
2. Don't save the file, put vim on *background*.

You should now have a `.RuleName.php.swp` file on the same directory of the rule. Without this commit, the mentioned test fails when trying to use that file as a valid rule.

VIM uses swap files to store unsaved changes. I confidently didn't use that option until I lost everything on my git repository after an HD crash causing a kernel panic.

alganet:
Oh I see! Many editors use these temporary files (including gedit). As I understood, your change would blacklist any rule names containing a dot.

A possible improvement would be to check for the extension itself:

```php
$ruleName = substr($fileInfo->getBasename(), 0, -4);
$ruleIsDeprecated = in_array($ruleName, $this->deprecateds);
$isNotARule = $fileInfo->getExtension() !== 'php';
```

By checking only for the dot, there is a possibility of a rule named "SomeRule.x.y" being caught. The `substr` should strip the entire `.x.y` (4 chars) suffix, leaving `SomeRule` clean and passing incorrectly. By checking the extension first, we prevent this and and future-proof support for other types of temporary editor files and variable-length file extensions.

augustohp:
> By checking only for the dot, there is a possibility of a rule named "SomeRule.x.y" being caught. The substr should strip the entire .x.y (4 chars) suffix, leaving SomeRule clean and passing incorrectly. By checking the extension first, we prevent this and and future-proof support for other types of temporary editor files and variable-length file extensions.

Nice siggestion, I've updated the test.

henriquemoody:
LGTM

PR #374, opened 2015-07-20
Create KeySet rule

### LINKED ISSUE #350: Validate array structure
I am sure this must be possible, but I am having a hard time figuring out how to validate array structure with this library. By this, I mean that the keys of the array are within a set of key values _and_ that each value associated with each key is of the right type. I initially thought this might be possible with each, but I don't see a way to ensure that the value is the right type for the key. For instance, if I have arrays of the form `['username' => 'name', 'user_id' => 102]`, how do I validate that the array only contains `username` and `user_id` and that `username` always points to a string while `user_id` points to an integer?

henriquemoody:
- `v::keySet(Key $rule...)`

Validates a keys in a defined structure.

```php
$dict = array('foo' => 42);

v::keySet(
    v::key('foo', v::int())
)->validate($dict); //true
```

Extra keys are not allowed:
```php
$dict = array('foo' => 42, 'bar' => 'String');

v::keySet(
    v::key('foo', v::int())
)->validate($dict); //false
```

Missing required keys are not allowed:
```php
$dict = array('foo' => 42, 'bar' => 'String');

v::keySet(
    v::key('foo', v::int()),
    v::key('bar', v::string()),
    v::key('baz', v::bool())
)->validate($dict); //false
```

Missing non-required keys are allowed:
```php
$dict = array('foo' => 42, 'bar' => 'String');

v::keySet(
    v::key('foo', v::int()),
    v::key('bar', v::string()),
    v::key('baz', v::bool(), false)
)->validate($dict); //true
```

The keys' order is not considered in the validation.

Fix #350


alganet:
I like it!

Viky-zhang:
Can `Extra keys are not allowed`  optional ?  
I have many params in my `$_POST`,  I will only use some of them.

henriquemoody:
Then just use a chain of `key()` rules, @Viky-zhang :)

KeySet exists only to validate structures, or else, just use nested `key()`.

PR #375, opened 2015-07-20
Fix wrong behaviour on Size rule

henriquemoody:
The validation was made only on `validate()` method, but it was
necessary to overwrite `assert()` and `check()` to replace the filename
by the size.
The `Size` rule now doesn't extends `Between` class, and instead it
extends `AbstractRule` and performs the validation only on `validate()`
method like many other rules.


henriquemoody:
Manual cherry-pick 4f8faab.

PR #384, opened 2015-08-08
Make inputs optional

### LINKED ISSUE #235: Different results for validate(), assert() and check()

henriquemoody:
Fix #235

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,18 +36,28 @@ validator that validates if a string is equal "Hello World".
 
 The rule itself needs to implement the `Validatable` interface.
 Also, it is convenient to extend the `AbstractRule`.
-Doing that, you'll only need to declare one method: `validate($input)`.
+Doing that, you'll only need to declare one method: `validateConcrete($input)`.
 This method must return `true` or `false`.
 
 If your validator class is `HelloWorld`, it will be available as `v::helloWorld()`
 and will natively have support for chaining and everything else.
 
 ```php
+
+/*
+ * This file is part of Respect/Validation.
+ *
+ * (c) Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
+ *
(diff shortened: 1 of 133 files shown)


henriquemoody:
Hey @Respect/troubleshooters, I'm really considering doing that for 1.0.

It's the only BC break I would consider for the first major version (1.0). I really need your attention here.

augustohp:
It is a pretty big BC break is it not? I'd rather have current API as is.

henriquemoody:
Yes it is, in the other hand that's how the library should work...

augustohp:
This defeats the purpose of keeping a compatible API doesn't it?




Although I am not against the idea, I am inclined to believe it defeats the purpose of keeping an old compatible API.

henriquemoody:
I don't see like that.

It's really a huge BC break, but this PR keeps the same _Concrete API_ we already (discussed in #354), but it changes the way to create rules indeed.

henriquemoody:
Since there was no agreement on that, I closed this pull request.

Any way, I'm not sure how we can proceed with the 1.0 version. I'm not really comfortable to release it as is specially because of the _optional input_ problem.

This pull request was only a alternative to the problem, but it's a very poor solution. Perhaps we should remove the _optional input_ and just create the `optional()` rule to handle this situation, I don't know...

What I do know is that we have to release 1.0 on Oct 24 and there is a lot to talk about it until them and that also we have 2.0 to release but we don't even finished the discussions about the new Concrete API and I don't wanna make the decisions by myself, it's not how I wanna do that.

PR #386, opened 2015-08-08
Display possible values when using in() with array

### LINKED ISSUE #385: Display possible values when using in() with array
The exception message for `in` with an array of strings will display `Array` instead of the haystack values.

`in('a', 'b')->validate('x')`
currently results in
`"x" must be in (Array)`
but should result in
`"x" must be in ('foo', 'bar')`

nochso:
Fixes #385

### DIFF
--- a/library/Exceptions/ValidationException.php
+++ b/library/Exceptions/ValidationException.php
@@ -39,16 +39,34 @@ class ValidationException extends InvalidArgumentException implements Validation
         if (is_string($value)) {
             return $value;
         } elseif (is_array($value)) {
-            $strings = array_filter($value, 'is_string');
-            if (count($strings) > 0) {
-                return "'" . implode("', '", $strings) . "'";
-            }
-            return 'Array';
+            return self::stringifyArray($value);
         } elseif (is_object($value)) {
             return static::stringifyObject($value);
         } else {
             return (string) $value;
         }
+    }
+
+    /**
+     * @param $value
+     * @return string
+     */
+    public static function stringifyArray($value)
+    {
+        $items = array();
+        foreach ($value as $val) {
+            if (is_object($val)) {
+                $items[] = self::stringifyObject($val);
+            } elseif (is_string($val)) {
+                $items[] = "'$val'";
+            } else {
+                $items[] = (string) $val;
+            }
+        }
+        if (count($items) > 0) {
+            return implode(", ", $items);
+        }
+        return 'Empty array';
     }
 
     public static function stringifyObject($value)
--- a/library/Exceptions/ValidationException.php
+++ b/library/Exceptions/ValidationException.php
@@ -39,7 +39,11 @@ class ValidationException extends InvalidArgumentException implements Validation
         if (is_string($value)) {
             return $value;
         } elseif (is_array($value)) {
-            return 'Array'; //FIXME
+            $strings = array_filter($value, 'is_string');
+            if (count($strings) > 0) {
+                return "'" . implode("', '", $strings) . "'";
+            }
+            return 'Array';
         } elseif (is_object($value)) {
             return static::stringifyObject($value);
         } else {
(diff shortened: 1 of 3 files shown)


library/Exceptions/ValidationException.php
```diff
@@ -39,7 +39,11 @@ public static function stringify($value)
         if (is_string($value)) {
             return $value;
         } elseif (is_array($value)) {
-            return 'Array'; //FIXME
+            $strings = array_filter($value, 'is_string');
+            if (count($strings) > 0) {
```
henriquemoody: That doesn't mean all values are strings.

library/Exceptions/ValidationException.php
```diff
@@ -39,7 +39,11 @@ public static function stringify($value)
...
+                return "'" . implode("', '", $strings) . "'";
```
henriquemoody: A more effective way to do this should probably use `json_encode()`.

tests/Exceptions/ValidationExceptionTest.php
```diff
@@ -84,7 +84,10 @@ public function providerForStringify()
         return array(
             array('foo', 'foo'),
             array(123, '123'),
-            array(array(), "Array"),
+            array(array(), "Empty array"),
```
henriquemoody: I'm :-1: for that, an empty list would be better.

nochso:
Moved it into a new function stringifyArray(). See the unit tests for the behaviour.

Is this ok? Also, which branch should this be based on? Accidentally based it off 0.9 but PR is on master :/ Should I close and open another PR?

tests/Exceptions/ValidationExceptionTest.php
nochso: So an empty string?

henriquemoody: Your list is "(1, 2, 3)" then "()".

henriquemoody:
You can just rebase to the current master

library/Exceptions/ValidationException.php
```diff
@@ -47,6 +47,28 @@ public static function stringify($value)
...
+    public static function stringifyArray($value)
+    {
+        $items = array();
+        foreach ($value as $val) {
+            if (is_object($val)) {
+                $items[] = self::stringifyObject($val);
+            } elseif (is_string($val)) {
+                $items[] = "'$val'";
+            } else {
+                $items[] = (string) $val;
```
henriquemoody: What if it is an array of arrays?

cloudson: Maybe : 

```php 
$array = [    
    1,
    "respect",
    [1,2 , [ "respect" ]]
]
```
could be printed as 

```
(1, "respect", Array) 
```
or we use recursion here with some way to define depth of recursion.

nochso: Either way is fine, I'll open another PR based on master

cloudson: What do you guys thinks about the solution ? @henriquemoody

nochso: Please see #387 for the recursive version

henriquemoody: I don't think it's a good idea because maybe what we what is an array of arrays:

```php
v::in([
    [0,0],
    [0,1],
    [1,1],
    [1,0],
])->assert([1, 2]);
```

In this case the message will have `(Array, Array, Array, Array)`.

Some checking of recursion level is necessary, but I think it's also needed, since we're changing it now.

I think the best option is to use json_encode and make a normalization before. A good example is the [NormalizerFormatter](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Formatter/NormalizerFormatter.php#L58-L139) from Monolog, but maybe I can implement that later, unless @nochso wants to do it.

nochso: I've already implemented a recursive version in #387 that should handle arrays of arrays:
```
+            array(array(1, array(2, array(3))), "1, (2, (3))"),
+            array(array(1, array(2, array(3, array(4)))), "1, (2, (3, ...))"),
```
The maximum depth is a public static property.

PR #387, opened 2015-08-08
Display possible values when using in() with array

### LINKED ISSUE #385: Display possible values when using in() with array

nochso:
Re-submitting PR #386 based off master
Fixes #385

### DIFF
--- a/library/Exceptions/ValidationException.php
+++ b/library/Exceptions/ValidationException.php
@@ -27,6 +27,10 @@ class ValidationException extends InvalidArgumentException implements Validation
             self::STANDARD => 'Data validation failed for %s',
         ),
     );
+    /**
+     * @var int Maximum depth when stringifying nested arrays
+     */
+    private static $maxDepthStringify = 3;
     protected $id = 'validation';
     protected $mode = self::MODE_DEFAULT;
     protected $name = '';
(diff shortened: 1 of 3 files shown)


nochso:
```
$array = [    
    1,
    "respect",
    [1,2 , [ "respect" ]]
]
```
`(1, 'respect', Array)`
or
`(1, 'respect', (1, 2, ('respect')))` with a maximum depth, indicated with `, ... `

Keep in mind that stringify() itself does not add the surround `()` so this might have to be reworked.

library/Exceptions/ValidationException.php
```diff
@@ -27,6 +27,10 @@ class ValidationException extends InvalidArgumentException implements Validation
             self::STANDARD => 'Data validation failed for %s',
         ),
     );
+    /**
+     * @var int Maximum depth when stringifying nested arrays
+     */
+    public static $maxDepthStringify = 2;
```
henriquemoody: Make it private, and increase that to 3

library/Exceptions/ValidationException.php
```diff
@@ -57,6 +61,31 @@ public static function stringify($value)
         }
     }
 
+    /**
+     * @param $value
+     * @return string
+     */
+    public static function stringifyArray($value, $depth = 0)
```
henriquemoody: Make it private and also fix the wrong docblock.

library/Exceptions/ValidationException.php
```diff
@@ -57,6 +61,31 @@ public static function stringify($value)
         }
     }
 
+    /**
+     * @param $value
+     * @return string
```
henriquemoody: New line before `@return`

library/Exceptions/ValidationException.php
```diff
@@ -57,6 +61,31 @@ public static function stringify($value)
         }
     }
 
+    /**
+     * @param $value
```
henriquemoody: Missing type and missing `$depth`

henriquemoody:
After the changes you also need to squash all commits at once.

nochso:
Ok, I've made the changes and squashed it into a single commit.

henriquemoody:
I did it manually on commit 2de6e013, because I also fixed some coding standards in your commit (to not bother you).

Thank you for contributing! :panda_face:

nochso:
Gladly. I just learned you could rebase / squash like that even though I already pushed on that branch, so thank you too :)

PR #402, opened 2015-09-05
Added PHPLabel rule

### LINKED ISSUE #401: New rule: String that is a valid PHP label (variable/function names, etc.)
Would be good to be able to validate that a string is a valid PHP label - useful when writing something that will take input to help create output that we expect to execute later.

From the PHP docs:

http://php.net/manual/en/functions.user-defined.php
http://www.php.net/manual/en/language.variables.basics.php

````
As a regular expression, it would be expressed thus: [a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*
````

thedavidmeister:
resolves #401

### DIFF
--- a/library/Exceptions/PHPLabelException.php
+++ b/library/Exceptions/PHPLabelException.php
@@ -1,0 +1,24 @@
+<?php
+
+/*
+ * This file is part of Respect/Validation.
+ *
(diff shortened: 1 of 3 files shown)


thedavidmeister:
see #403

PR #403, opened 2015-09-06
401 php label

### LINKED ISSUE #401: New rule: String that is a valid PHP label (variable/function names, etc.)

thedavidmeister:
Exactly the same as #402 but against a dedicated branch. Fixes #401

### DIFF
--- a/tests/unit/Rules/PHPLabelTest.php
+++ b/tests/unit/Rules/PHPLabelTest.php
@@ -78,7 +78,7 @@ class PHPLabelTest extends \PHPUnit_Framework_TestCase
             mt_rand(),
             0,
             1,
-            [],
+            array(),
             new \StdClass(),
             new \DateTime(),
         ));
--- a/tests/unit/Rules/PHPLabelTest.php
+++ b/tests/unit/Rules/PHPLabelTest.php
@@ -1,0 +1,79 @@
+<?php
+
+/*
+ * This file is part of Respect/Validation.
+ *
+ * (c) Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
+ *
+ * For the full copyright and license information, please view the "LICENSE.md"
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Rules;
+
+/**
+ * @group  rule
+ * @covers Respect\Validation\Rules\PHPLabel
+ * @covers Respect\Validation\Exceptions\PHPLabelException
+ */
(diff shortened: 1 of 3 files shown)


henriquemoody:
Could you please create PHPLabel.md on "docs" directory?

tests/unit/Rules/PHPLabelTest.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * @dataProvider providerForNotPHPLabel
+     * @expectedException Respect\Validation\Exceptions\PHPLabelException
+     */
+    public function testInvalidPHPLabelShouldThrowPhoneException($input)
+    {
+        $this->assertFalse($this->phpLabelValidator->__invoke($input));
+        $this->assertFalse($this->phpLabelValidator->assert($input));
```
luispabon: Just to point out, all assert methods on PHPUnit are static. This might stop working at some point in the future.

henriquemoody: I think it's not going to happen, see https://github.com/sebastianbergmann/phpunit/issues/1914#issuecomment-148407321

luispabon: Doesn't depend on him, but rather on PHP itself becoming stricter over time and hacks like this being removed; which admittedly won't happen just yet. Regardless of what Sebastian Bergmann says, $this-> to access a static method is wrong and might stop working in the future. 

PHPUnit's hardly an example of good php code.

henriquemoody: When/if it happens a single line of shell script can change the entire test suite and, since it's not BC break, because it's on the test suite, it can be changed in any version.

I think you're right (on many points), but until it happens I think we should stick with what is a common and widely accepted standard, don't you think?

luispabon: You're right indeed :+1: . Just trying to raise awareness as lots of people haven't realised never having looked at PHPUnit assertions code.

henriquemoody:
Closed by f40eb63d7b640697f612d513ea47b61c6a3f6293

Issue #379, opened 2015-07-29
Tests refactoring


henriquemoody:
We have to refactor the tests before we launch version 2.0.

- [x] Move exception testes (`assert()` and/or `check()`) methods to PHPT
- [x] Remove duplicated tests of `assert()` and `check()` of classes that are children of `AbstractRule`
- [x] Move all integration tests to PHPT tests
- [x] Create `RuleTestCase` class in order to make rule tests easier
- [x] Create tests for `AbstractRule` class
- [x] Create tests for abstract classes
- [x] Create tests for untested rules
- [x] Create integration test suite
- [x] Move unit tests to `tests/unit` directory
- [x] Use `@covers` annotation on rule tests (757c377)
- [x] Use `@group rule ` annotation on rule tests (757c377)

Most of this issue was discussed by @alganet, @augustohp and me on Hangouts.

alganet:
I'm working on the tests for the AbstractRule class.

batusa:
I'm working on the tests for abstract classes (:

PR #405, opened 2015-09-06
added FactorOf

### LINKED ISSUE #404: New Rule: FactorOf
Should validate as `true` if `$input` is a factor of a given integer.

thedavidmeister:
This is a WIP on a new FactorOf rule and will resolve #404 

Mostly, it just needs more tests for the exceptions.

### DIFF
--- a/library/Rules/FactorOf.php
+++ b/library/Rules/FactorOf.php
@@ -11,21 +11,31 @@
 
 namespace Respect\Validation\Rules;
 
+use Respect\Validation\Exceptions\ValidationException;
 use Respect\Validation\Exceptions\FactorOfException;
+use Respect\Validation\Exceptions\ComponentException;
 
 class FactorOf extends AbstractRule
 {
+    public $dividend;
+
     public function __construct($dividend)
     {
-        if (!is_numeric($dividend) || (int) $dividend != $dividend || $dividend == 0) {
-            throw new FactorOfException('Dividend must be an integer greater than 0');
+        if (!is_numeric($dividend) || (int) $dividend != $dividend) {
+            throw new ComponentException('Dividend ' . ValidationException::stringify($dividend) . ' must be an integer');
         }
 
-        $this->dividend = $dividend;
+        $this->dividend = (int) $dividend;
     }
 
     public function validate($input)
     {
+        // Every integer is a factor of zero, and zero is the only integer that
+        // has zero for a factor.
+        if ($this->dividend === 0) {
+            return true;
+        }
+
         // Factors must be integers that are not zero.
         if (!is_numeric($input) || (int) $input != $input || $input == 0) {
             return false;
--- a/library/Rules/FactorOf.php
+++ b/library/Rules/FactorOf.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of Respect/Validation.
+ *
+ * (c) Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
+ *
+ * For the full copyright and license information, please view the "LICENSE.md"
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Rules;
+
+use Respect\Validation\Exceptions\FactorOfException;
+
+class FactorOf extends AbstractRule
+{
+    public function __construct($dividend)
+    {
+        if (!is_numeric($dividend) || (int) $dividend != $dividend || $dividend == 0) {
+            throw new FactorOfException('Dividend must be an integer greater than 0');
+        }
+
+        $this->dividend = $dividend;
+    }
+
+    public function validate($input)
+    {
+        // Factors must be integers that are not zero.
+        if (!is_numeric($input) || (int) $input != $input || $input == 0) {
+            return false;
+        }
+
+        $input = (int) abs($input);
+        $dividend = (int) abs($this->dividend);
+
+        // The composite divided by the input must be an integer if input is a
+        // factor of the composite.
+        return is_integer($dividend / $input);
+    }
+}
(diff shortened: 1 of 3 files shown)


library/Rules/FactorOf.php
```diff
@@ -0,0 +1,41 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Rules;
+
+use Respect\Validation\Exceptions\FactorOfException;
+
+class FactorOf extends AbstractRule
+{
+    public function __construct($dividend)
```
henriquemoody: You must define $dividend as a valid public property.

library/Rules/FactorOf.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        if (!is_numeric($dividend) || (int) $dividend != $dividend || $dividend == 0) {
+            throw new FactorOfException('Dividend must be an integer greater than 0');
```
henriquemoody: In this case you have to use `ComponentException` instead. That's because the `FactorOfException` must be used only when you call `assert()` or `check()` methods.

henriquemoody: Thus, every integer (which are also negative numbers) is a factor of zero.

tests/unit/Rules/FactorOfTest.php
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public function providerForInvalidFactorOfDividend()
+    {
+        return array(
+            // Invalid dividend, valid input.
+            array(0, 0),
+            array(0, 1),
+            array(0.5, mt_rand()),
+            array(1.5, mt_rand()),
```
henriquemoody: I would prefer if you don't use random values on tests for two reasons:

1. It's better to read and understand
2. It's not unpredictable

Talking a little bit more about 2, the indention of the testing, as many other intentions as well, is to ensure the feature works as expected, having unpredictable values only makes the suite unstable, if it breaks in some point, it may be hard to reproduce the exact same error.

thedavidmeister: 1. I disagree, a singular, native function call is hardly difficult to read or understand
2. That's precisely the point, using randomness forces us to write code that is valid for all integers that could be generated and not just the very small subset of integers that I could write by hand

"if it breaks in some point"

yes, there may be hard to reach bugs in places we have not thought of, that only a random test would catch.

"it may be hard to reproduce the exact same error"

Not true, the value that fails should be in the PHPUnit logs, we can see exactly which integer broke and then add that as a special/edge case explicitly in the tests at that time.

library/Rules/FactorOf.php
thedavidmeister: Sure, I'll clean this up next time I'm working on this :)

library/Rules/FactorOf.php
thedavidmeister: yup, np

henriquemoody:
@thedavidmeister, could you also squash your commits at once, please?

tests/unit/Rules/FactorOfTest.php
thedavidmeister: FYI, other people think that some degree of generative testing _alongside_ explicit, hand written unit tests is not a bad thing:

- https://en.wikipedia.org/wiki/QuickCheck
- https://github.com/steos/php-quickcheck
- http://www.pivotaltracker.com/community/tracker-blog/generative-testing

thedavidmeister:
@henriquemoody hang on, still getting it right >.<

henriquemoody:
Soon we will drop support for 5.3, then we're not going to have these problems anymore. :smile:

thedavidmeister:
yeah the `array()` stuff is super spammy syntax :( glad to be rid of it.

thedavidmeister:
@henriquemoody kk, how do you feel about this now?

henriquemoody:
Merged manually on f14e539, also renames it to _factory_ because that's how most of rules are named on the current version.

I still disagree with your tests, there are a lot of kinds of tests, I think unit tests must be as simpler as possible, these tests have too many logic stuff, they gotta be simpler, I will probably refactor them when I have time.

thedavidmeister:
@henriquemoody thanks! This is great, I was looking forward to using this in my own projects. Apologies for not getting the naming conventions right, and it looked like I missed a few other details too :(

On the testing, feel free to refactor the logic, but I believe a validation library is a _great_ use case for generative testing and you might want to even consider using a library like Quickcheck. Maybe think some more about this in your own time and see if you still disagree with me later.

The reproducibility problem in unit testing is about functions that are non-deterministic because they either have or rely on side effects (the functions are not pure).

Examples:
- Relying on the state of a database
- Relying on a response from a third party API
- Relying on a specific time and/or date period to work
- A function that writes to disk and reports the results
- A function that *internally* generates a random value and generates an output based on that

The reproducibility problem isn't about randomness in the data set, provided it is auditable and shares a common property we want to verify (e.g. "positive integers", "strings that begin with the letter B", "phone numbers").

In a validation library all the rules *really* should be referentially transparent anyway (only one possible output for any given input), so the goal for testing should be to flush out as many edge cases as possible that we may not have thought of that confound our rule logic (or even future logic we haven't written yet) - but in the case that we do find an edge case, it is trivial to add it to the explicit part of the test suite - there is no problem reproducing any failures in the tests, and so there is no fragility.

To highlight that any potential failures are perfectly auditable in the current setup, here is a screenshot of a failing test:

screen shot 2015-09-11 at 1 20 16 am

You can see it failed on data set `(0, 1259392607)`, the random integer didn't impact my ability to fix this bug at all!

henriquemoody:
I'm not against that, I'm just against to have untested logic on the tests, but using a library, which has its own tests, sounds good to me; to be honest I thought about using Faker as data provider once.

Perhaps you should share your thoughts on #379, it will be awesome!

PR #410, opened 2015-09-15
Add Youtube, Vimeo, Video (both) validator

ricardogobbosouza:


henriquemoody:
Cool! Thanks for contributing! :panda_face: 

I'm just confusing about the names, because a _video_ is not the same of a _video URL_, a _Youtube video_ is not the same of _Youtube_ and so on... Could you rename them?

library/Rules/Video.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Respect\Validation\Rules;
+
+class Video extends AbstractRule
+{
+    public function validate($input)
+    {
+        return !empty($input) && (
+            preg_match('/^.*((youtu.be\/)|(v\/)|(\/u\/\w\/)|(embed\/)|(watch\?))\??v?=?([^#\&\?]*).*/i', $input) ||
+            preg_match('/^.*(vimeo\.com\/)((channels\/[A-z]+\/)|(groups\/[A-z]+\/videos\/)|(video\/))?([0-9]+)/i', $input)
```
henriquemoody: Here and in other patterns you allow any protocol, it that what you indented to do?

ricardogobbosouza: They are validators URL:

Youtube - Accepts URL Youtuve
Vimeo - Accepts URL Vimeo
Video - Supports both URL

I would like to move to UrlYoutube?

Sorry my bad engilish.

henriquemoody: I would probably go for only one rule, called `videoUrl()` which accepts argument on its construtor, that can be `"youtube"` or `"vimeo"`, them we can implement more, like `"ted"`, `"viki"`... You can create a key => value array with the patterns to match for the defined video provider (youtube, vimeo...) or try all of them if no provider is defined.

ricardogobbosouza: Ok. I'm changing ..
Thanks

henriquemoody:
Merged manually by a0261245904b8dc48ec3c6a23648d0d41e6575b6

ricardogobbosouza:
Thanks for help!

Agora que vi que você é de São Paulo.
Desculpa meu péssimo inglês!
Obrigado

henriquemoody:
Sem problemas, man! :smile:

PR #419, opened 2015-10-03
Upgrade PHPUnit version

henriquemoody:


henriquemoody:
It cannot be done because PHPUnit 5.0 dropped support for PHP < 5.6

PR #420, opened 2015-10-05
Make all rules optional

### LINKED ISSUE #235: Different results for validate(), assert() and check()

henriquemoody:
This is the hard-coded solution I was avoiding but seems the best thing to do, due the created issues related to the _optional input_ feature.

All rules should accept optional values (`''`) and that behaviour must be the same for `assert()`, `check()` and `validate()` methods.

Create `notOptional()` rule to handle optional values. This rule uses the defined optional values for the library by using the method `Validator::getOptionalValues()`. This commit also allows the user to customize optional values by `Validator::setOptionalValues()`.

Fix #235, Closes #422

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,6 +49,10 @@ class HelloWorld extends AbstractRule
 {
     public function validate($input)
     {
+        if ($this->isOptional($input)) { // Handles optional inputs
+            return true;
+        }
+
         return $input === 'Hello World';
     }
 }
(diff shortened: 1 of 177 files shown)


henriquemoody:
ping @Respect/troubleshooters

alganet:
I'm :+1: for making all of the rules mandatory on the next major.

henriquemoody:
@alganet, thanks for the feedback! 

I also created the `optional()` rule, then we can still handle optional inputs.

PR #421, opened 2015-10-05
Nested key validator

zinovyev:
Key validator for nested keys: 'foo.bar.baz' => array('foo' => array('bar' => 'baz' => ...
Inspired by Yii2 ArrayHelper class.

### DIFF
--- a/docs/NestedKey.md
+++ b/docs/NestedKey.md
@@ -1,0 +1,38 @@
+# NestedKey
+
+- `v::nestedKey(string $name)`
+- `v::nestedKey(string $name, v $validator)`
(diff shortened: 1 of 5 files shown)


zinovyev:
I've added an additional check for the array key containing a dot sign, e.g. `array( 'bar.baz' => 'foo' )`.

henriquemoody:
I made some changes you this rule

1. Named is as KeyNested to keep the "key" prefix, since we have `key()` and `keySet()` rules
2. If you use `foo.bar` as reference and you also have a `foo.bar` it doesn't validate, because that's the behaviour `key()` already have, it may create confusion
3. Improved documentation
4. Fixed some coding standards 

Thank you very much for your contribution! :panda_face:

PR #439, opened 2015-10-08
Method notEmpty - Issue 438

### LINKED ISSUE #438: Isn't possible validate if "0" (zero as string) is a string with "notEmpty" method.
This test will return an exception:

```php
Validator::type('string')
    ->notEmpty()
    ->check("0");
```

The tests in "library/Rules/NotEmpty.php" don't cover this case.

bsmrs:
modified:   library/Rules/NotEmpty.php
modified:   tests/unit/Rules/NotEmptyTest.php

- The test for empty strings was modified.
- A new value for "NotEmptyTest::providerForNotEmpty" was added.

fix #438

### DIFF
--- a/library/Rules/NotEmpty.php
+++ b/library/Rules/NotEmpty.php
@@ -17,6 +17,7 @@ class NotEmpty extends AbstractRule
     {
         if (is_string($input)) {
             $input = trim($input);
+            return isset($input[0]);
         }
 
         return !empty($input);
--- a/tests/unit/Rules/NotEmptyTest.php
+++ b/tests/unit/Rules/NotEmptyTest.php
@@ -47,6 +47,7 @@ class NotEmptyTest extends \PHPUnit_Framework_TestCase
         return array(
             array(1),
             array(' oi'),
+            array('0'),
             array(array(5)),
             array(array(0)),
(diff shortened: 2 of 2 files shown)


library/Rules/NotEmpty.php
```diff
@@ -17,6 +17,7 @@ public function validate($input)
     {
         if (is_string($input)) {
             $input = trim($input);
+            return isset($input[0]);
```
henriquemoody: But that will still invalidate `0` (integer) is that what you expect?

bsmrs: In my opinion this is a side effect of calling a method to validate strings from a integer validation. An integer can't be empty while a string can.

henriquemoody:
Hey, @bsmrs! I really appreciate your pull request, but since [there will be no more optional input](https://github.com/Respect/Validation/blob/b28e97c/docs/README.md#input-optional), unless you use [optional()](https://github.com/Respect/Validation/blob/b28e97c/docs/Optional.md) rule I don't think this change still makes sense.

However, If you do think it still does, fell free to reopen this pull request any time. I have plans to update this rule to follow the same

Thanks for contributing to the Respect\Validation project :panda_face:

Issue #472, opened 2015-10-17
Refactor unit test "NotTest"


reginaldojunior:
Refactor "NotTest" for more test and % coverage in Class

nickl-:
Closed: as per commit Respect/Validation@87708fb

Issue #434, opened 2015-10-07
Methods `assert()` and `check()` should not return value


henriquemoody:
These methods don't return a boolean value, they only return `true` when success or throw a [ValidationExceptionInterface](https://github.com/Respect/Validation/blob/master/library/Exceptions/ValidationExceptionInterface.php).

Since they throw an exception when error, they can just throw it or not, they don't need to return `true` for success.

henriquemoody:
Fixed by c80524b4574b780d71b1c348417f0094aad8b64a and 30dc089565b15a1fcbf24f6b70f2104ee0eea1ea

PR #481, opened 2015-10-17
close issue #472 add more one function for test class

reginaldojunior:
only more a function for test in class Not


henriquemoody:
As we were talking early. That's the kind of tests we're trying to avoid.

First we doesn't matter the return of `assert()` method (#434) and secondly unit tests should not use real classes, like `IntVal` for testing, and instead we should use mocks for that.

I will keep this pull request opened, how about you use mocks for that?

PR #491, opened 2015-10-17
Refactoring of CnpjTest class

batusa:
removed unnecessary asserts and renamed some methods which test for excepted exceptions.

### DIFF
--- a/tests/unit/Rules/CnpjTest.php
+++ b/tests/unit/Rules/CnpjTest.php
@@ -65,33 +65,6 @@ class CnpjTest extends \PHPUnit_Framework_TestCase
         $this->assertFalse($this->cnpjValidator->validate($input));
     }
 
-    /**
-     * @dataProvider providerInvalidFormattedCnpj
-     * @expectedException Respect\Validation\Exceptions\CnpjException
-     */
-    public function testInvalidCnpjShouldThrowCnpjExceptionWhenFormatted($input)
-    {
-        $this->cnpjValidator->assert($input);
-    }
-
-    /**
-     * @dataProvider providerInvalidUnformattedCnpj
-     * @expectedException Respect\Validation\Exceptions\CnpjException
-     */
-    public function testInvalidCnpjShouldThrowCnpjExceptionWhenNotFormatted($input)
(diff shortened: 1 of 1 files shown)


tests/unit/Rules/CnpjTest.php
```diff
@@ -41,38 +39,33 @@ public function testFormattedCnpjsShouldValidate($input)
...
     /**
      * @dataProvider providerInvalidFormattedAndUnformattedCnpjLength
      * @expectedException Respect\Validation\Exceptions\CnpjException
      */
-    public function testCnpjsWithIncorrectLengthShouldThrowCnpjExceptionAndReturnFalse($input)
+    public function testCnpjsWithIncorrectLengthShouldThrowCnpjException($input)
     {
-        $this->assertFalse($this->cnpjValidator->validate($input));
-        $this->assertFalse($this->cnpjValidator->assert($input));
+        $this->cnpjValidator->assert($input);
```
henriquemoody: Thanks the kind of test we're trying to avoid, instead you should validate just the `validate()` method, since it's the [only existing method on Cnpj rule](https://github.com/Respect/Validation/blob/master/library/Rules/Cnpj.php#L16-L39).

batusa: the valildate method won't throw an exception.

henriquemoody: Yes, then you should remove the `@expectedException` annotation.

The exceptions must be tested by PHPT tests.

batusa: so I should remove all "expected exception test methods", right?

batusa: remove only `@expectedException` annotation wouldn't be enough as the method purpose is only test for exceptions.

batusa:
check it out now :)

henriquemoody:
Awesome!

henriquemoody:
Merged manually by dbad46a1c1a295a49adacc2b95f6794f01d28ac3

batusa:
yay! :smiley:

PR #544, opened 2015-10-17
Create "Countable" rule

### LINKED ISSUE #524: Create "Countable" rule that validates if a value is countable or not;

joaotorquato:
Fix #524

### DIFF
--- a/library/Exceptions/CountableException.php
+++ b/library/Exceptions/CountableException.php
@@ -1,0 +1,24 @@
+<?php
+
+/*
+ * This file is part of Respect/Validation.
+ *
+ * (c) Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
+ *
+ * For the full copyright and license information, please view the "LICENSE.md"
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Exceptions;
+
+class CountableException extends ValidationException
+{
+    public static $defaultTemplates = array(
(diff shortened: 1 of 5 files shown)


library/Exceptions/CountableException.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Respect\Validation\Exceptions;
+
+class CountableException extends ValidationException
+{
+    public static $defaultTemplates = array(
+        self::MODE_DEFAULT => array(
+            self::STANDARD => '{{name}} must be an array',
+        ),
+        self::MODE_NEGATIVE => array(
+            self::STANDARD => '{{name}} must not be an array',
```
henriquemoody: The correct text for both message must be "a countable" instead of "an array".

library/Rules/Countable.php
```diff
@@ -0,0 +1,23 @@
...
+ * (c) Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
+ *
+ * For the full copyright and license information, please view the "LICENSE.md"
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Rules;
+
+use ArrayAccess;
+use Traversable;
```
henriquemoody: These two classes have not been used.

henriquemoody:
Merged manually by a3694241f2a846f67c96a0ba00e2312070072bd1 with some addictions and changes.

Issue #563, opened 2015-10-18
Json rule is accepting data which are not json


batusa:
while I was checking out some unit tests, I sported with the [JsonTest](https://github.com/Respect/Validation/blob/master/tests/unit/Rules/JsonTest.php#L52) class, and there's a data provider which includes *strings* such as: 

* 2
* null
* false
* "abc"

as valid json formats.

in my understanding they are not jsons and cannot be validated as a json string.

henriquemoody:
Fixed by dddad80f485dc9307986c59b844740003ab2814b

PR #564, opened 2015-10-18
Json rule is accepting data which are not json

batusa:
That's my suggestion up to now. This would fix the issue #563 .


library/Rules/Json.php
```diff
@@ -15,11 +15,12 @@ class Json extends AbstractRule
 {
     public function validate($input)
     {
-        if (is_string($input)
-            && strtolower($input) == 'null') {
-            return true;
+        if (!is_string($input)) {
```
henriquemoody: :+1:

library/Rules/Json.php
```diff
@@ -15,11 +15,12 @@ class Json extends AbstractRule
...
+            return false;
         }
 
-        return (null !== json_decode($input));
+        $decode = json_decode($input);
+
+        return is_object($decode) || is_array($decode);
```
henriquemoody: I would not do that. According to [json.org](http://www.json.org/), a JSON have three forms:
- An object is an unordered set of name/value pairs
- An array is an ordered collection of values
- A value can be a string in double quotes, or a number, or true or false or null, or an object or an array.

It doesn't need to be an array or an object, see this example:

```
php > var_dump(json_decode("Hello"));
NULL
php > var_dump(json_decode('"Hello"'));
string(5) "Hello"
```

See, that's because `"Hello"` is a valid Javascript notation.

henriquemoody: The proper body for this `validate()` method should be:
```php
        if (is_string($input)) {
            return true;
        }

        json_decode($input);

        return (json_last_error() === JSON_ERROR_NONE);
```

batusa:
```php
json_decode($input);

return (json_last_error() === JSON_ERROR_NONE);
```

I've found this part of code in many places on the web, telling that would validate a json. but in fact many people has also complained about it exactly because it would also validate only numbers or certain strings.

according to json,org actually, a json is built on TWO data structures, AND, inside these data structures there will be a value, and yes, these values can be:

* string
* number
* object
* array
* true
* false
* null

henriquemoody: I'm still not about it, I'm going to read [RFC 7159](https://tools.ietf.org/html/rfc7159) as soon as I can.

batusa: a few things I got from this (taking a quick reading...):

this [RFC 7159](https://tools.ietf.org/html/rfc7159) is an update of [RFC 4627](https://tools.ietf.org/html/rfc4627). in 4627 a json is considered either an object or an array.

in 7149 it says: "Note that certain previous specifications of JSON constrained a JSON text to be an object or an array.". in the [examples section](https://tools.ietf.org/html/rfc7159#page-12) of 7149 it considers the following as three small JSON texts containing only values:

   * "Hello world!"
   * 42
   * true

using some json parsers on the web such as: http://jsonlint.com/, http://json.parser.online.fr/
they consider the three values above as invalid, I think they follow the 4627.

well, this changed my mind about only validating either an object or an array :)

what do we do now? :stuck_out_tongue_winking_eye:

henriquemoody: It that's in the [RFC 7159](https://tools.ietf.org/html/rfc7159) that's what we should do, so let's stick with:

```php
json_decode($input);

return (json_last_error() === JSON_ERROR_NONE);
```



If PHP is able to parse it, that means it's a valid JSON. Can you update this pull request with this change?

PS.: I'm going to read it any way, but I think we should trust PHP for now.

batusa:
Sure! Lemme get home ;) (I'm on the street lol)

library/Rules/Json.php
batusa: even an empty string it's considering as true, I won't accept empty strings, ok?

henriquemoody: As I said, I think we should trust on PHP `json_decode()`

tests/unit/Rules/JsonTest.php
```diff
@@ -11,28 +11,23 @@
 
 namespace Respect\Validation\Rules;
 
+use Respect\Validation\Test\RuleTestCase;
```
henriquemoody: Just remove this line, `RuleTestCase` is in the same namespace of the test (and the rule).

henriquemoody:
Merged manually by dddad80f485dc9307986c59b844740003ab2814b

PR #567, opened 2015-10-18
Add currency code rule.

### LINKED ISSUE #430: Create "CurrencyCode" rule
It must use [ISO 4217](http://en.wikipedia.org/wiki/ISO_4217).

JustinHook:
Closes #430

### DIFF
--- a/docs/CurrencyCode.md
+++ b/docs/CurrencyCode.md
@@ -1,0 +1,9 @@
+# CurrencyCode
+
+- `v::currencyCode()`
+
+Validates an ISO currency code like GBP or EUR.
+
+```php
+v::currencyCode()->validate('GBP'); //true
+```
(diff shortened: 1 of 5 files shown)


henriquemoody:
Merged manually by 4145faab39f7cd76d2f0586c12dba4278ac35fe7

Looks like you just forgot to change the Validator docblock, I have to update the documentation with that step.

Thanks a lot! :panda_face:

batusa:
would you guys consider favor hash lookups over array searches? http://mtdowling.com/blog/2014/03/17/hash-lookups-over-array-search/

henriquemoody:
For an array like this, with 178 items, I'm not sure how much would it make difference, besides I don't see a scenario where validation would be performed for 10000 times, like in the link you sent us, I think this micro-optimization it not worthy to make this code ugly with an associative array which the values will be always the same value.

What you guys think?

batusa:
@henriquemoody I was in doubt about it too. So let's keep it as is.

PR #568, opened 2015-10-18
Creates unit test for CurrencyCode rule

batusa:
:+1:


henriquemoody:
Thanks @batusa, but [RuleTestCase](https://github.com/Respect/Validation/blob/master/tests/library/RuleTestCase.php) does that already!

However, thanks for contributing! :panda_face:

batusa:
oh sorry! I didn't know that!

henriquemoody:
No problem :sunglasses:

PR #575, opened 2015-10-19
Create "composer test" command in order to run phpunit tests

rogeriopradoj:
As the project already is installed via composer, this command
could save a few keystrokes!

Besides that, is just one command for all OS, because will not be
needed to change slashes to backslashes in Windows.

*X      (21 keystrokes): `./vendor/bin/phpunit`
Windows (21 keystrokes): `.\vendor\bin\phpunit`
New (14 keystrokes):     `composer test`

https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -187,11 +187,23 @@ You can test the project using the commands:
 $ vendor/bin/phpunit
 ```
 
+or
+
+```sh
+$ composer test
+```
+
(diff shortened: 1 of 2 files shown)


henriquemoody:
Thanks, @rogeriopradoj!!

Rogério patinete

rogeriopradoj:
\o/

Issue #581, opened 2015-10-21
Create "BoolVal" rule


henriquemoody:
It must use `filter_var($input, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE)` to validate it.

batusa:
I'll do it ;)

henriquemoody:
Awesome, @batusa!

henriquemoody:
Fixed by #581

PR #583, opened 2015-10-21
Create "BoolVal" rule

batusa:
https://github.com/Respect/Validation/issues/581

### DIFF
--- a/tests/integration/boolval_1.phpt
+++ b/tests/integration/boolval_1.phpt
@@ -1,0 +1,23 @@
+--FILE--
+<?php
+require 'vendor/autoload.php';
+
+use Respect\Validation\Validator as v;
+
+v::boolVal()->assert(1);
+v::boolVal()->assert('on');
+v::boolVal()->assert('off');
+v::boolVal()->assert('yes');
+v::boolVal()->assert('no');
+v::boolVal()->assert(true);
+v::boolVal()->assert(false);
+
+v::boolVal()->check(1);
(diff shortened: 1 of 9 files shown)


tests/integration/boolval_1.phpt
```diff
@@ -0,0 +1,15 @@
...
+
+use Respect\Validation\Validator as v;
+
+v::boolVal()->validate(1);
+v::boolVal()->validate('on');
+v::boolVal()->validate('off');
+v::boolVal()->validate('yes');
+v::boolVal()->validate('no');
+v::boolVal()->validate(true);
+v::boolVal()->validate(false);
```
henriquemoody: I think that suppose to be `check()` and/or `assert()` methods, isn't? Otherwise we are just reproducing what was in the unit tests, right?

tests/integration/boolval_2.phpt
```diff
@@ -0,0 +1,30 @@
...
+} catch (BoolValException $e) {
+    echo $e->getMainMessage() . PHP_EOL;
+}
+
+try {
+    v::boolVal()->assert('yep');
+} catch (AllOfException $e) {
+    echo $e->getFullMessage() . PHP_EOL;
+}
+?>
```
henriquemoody: I think you just missed the tests for the `assert()` method with `not()` rule.

henriquemoody:
Perfect! Thanks for contributing! :panda_face:

Issue #593, opened 2015-10-25
MinimunAge rule should not check if age is an integer type


henriquemoody:
On [MinimumAge](https://github.com/Respect/Validation/blob/master/library/Rules/MinimumAge.php#L29-L31) rule, it checks if the input is an integer, but instead is must use `filter_var()` with `FILTER_VALIDATE_INT` to also accept also integer numbers as string types.

PR #594, opened 2015-10-25
Fix minimumAge rule

KennedyTedesco:
This is related to #593

### DIFF
--- a/library/Rules/MinimumAge.php
+++ b/library/Rules/MinimumAge.php
@@ -12,6 +12,7 @@
 namespace Respect\Validation\Rules;
 
 use DateTime;
+use Respect\Validation\Exceptions\ComponentException;
 
 class MinimumAge extends AbstractRule
 {
@@ -20,16 +21,16 @@ class MinimumAge extends AbstractRule
 
     public function __construct($age, $format = null)
     {
+        if (!filter_var($age, FILTER_VALIDATE_INT)) {
+            throw new ComponentException('The age must be a numeric value.');
+        }
+
         $this->age = $age;
         $this->format = $format;
     }
 
     public function validate($input)
     {
-        if (!is_int($this->age)) {
-            return false;
-        }
-
         if ($input instanceof DateTime) {
             $birthday = new \DateTime('now - '.$this->age.' year');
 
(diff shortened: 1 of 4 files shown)


library/Rules/MinimumAge.php
```diff
@@ -20,16 +21,16 @@ class MinimumAge extends AbstractRule
 
     public function __construct($age, $format = null)
     {
+        if (!filter_var($age, FILTER_VALIDATE_INT)) {
+            throw new ComponentException('The age must be a numeric value.');
```
henriquemoody: This is great for version 1.1, but it will not me released on versions 0.8>.

library/Rules/MinimumAge.php
```diff
@@ -20,16 +21,16 @@ class MinimumAge extends AbstractRule
...
+        }
+
         $this->age = $age;
         $this->format = $format;
     }
 
     public function validate($input)
     {
-        if (!is_int($this->age)) {
```
henriquemoody: If you just change this line to `filter_var($age, FILTER_VALIDATE_INT)`, and send a pull request agains 0.8 branch, then we can release a path version with this fix, isn't that what you want?

KennedyTedesco: Yes, this is exactly what I want. But it will also go to 1.0, right?

henriquemoody: Yes, according to our [Release Cycle and Support](https://github.com/Respect/Validation/wiki/Release-Cycle-and-Support) we're still supporting 0.8, 0.9 and 1.0 as  well, of course.

henriquemoody: But, you have to send another pull request against 0.8 branch to fix that on those versions.

KennedyTedesco: Got it. I'll do it soon. :smile:

henriquemoody:
After b98595a5c093bfebd8686646fffb6b441fce6a08 these changes leads with code duplication, can you rebase and change this code to fit the new codebase?

KennedyTedesco:
Sure! I'll do it asap.

henriquemoody:
Merged manually by cca733919c9f272ffe162b75fcc0043f5b93ea1a.

Thanks for contributing! 🐼

PR #595, opened 2015-10-25
Making real tests for `Exists` and integration tests added

KennedyTedesco:
This is just a proposal because it's not so bad the test with a real file. Or this approach is somehow problematic?

### DIFF
--- a/tests/integration/rules/exists_1.phpt
+++ b/tests/integration/rules/exists_1.phpt
@@ -1,0 +1,17 @@
+--FILE--
+<?php
+require 'vendor/autoload.php';
+
+use org\bovigo\vfs\content\LargeFileContent;
+use Respect\Validation\Validator as v;
+use org\bovigo\vfs\vfsStream;
+
+$root = vfsStream::setup();
+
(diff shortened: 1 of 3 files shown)


henriquemoody:
I'm not sure about what is the benefits of this change. I would rather use [vfsStream](http://vfs.bovigo.org/) like we do on [SizeTest](https://github.com/Respect/Validation/blob/master/tests/unit/Rules/SizeTest.php#L102-L111) for that.

KennedyTedesco:
Yeah, that's fair enough. If you want to I can send a new one PR using vfsStream. Or we just maintain the current implementation.

henriquemoody:
If you don't mind, just send thought this pull request

KennedyTedesco:
Now i'm happy with this. :) If there is anything more that I have to do, please let me know.

henriquemoody:
Merged manually by 7d8b18bacae9967912e4acf0fdc3e88be13cf56a

KennedyTedesco:
:beers:

PR #597, opened 2015-10-26
Create "LanguageCode" rule

batusa:
the `LanguageCode` class is very similar to `CountryCode` class. would you consider a parent class called AbstractIsoCode to remove these duplications? @henriquemoody 

ps: as you can see in the `$languageCodeList`, a few languages really don't have alpha-2 code

### DIFF
--- a/docs/LanguageCode.md
+++ b/docs/LanguageCode.md
@@ -1,0 +1,20 @@
+# LanguageCode
+
+- `v::languageCode()`
+
+Validates a language code based on ISO 639:
+
+```php
+v::languageCode()->validate('pt'); // true
+v::languageCode()->validate('en'); // true
+v::languageCode()->validate('it'); // true
+v::languageCode('alpha-3')->validate('ita'); // true
+v::languageCode('alpha-3')->validate('eng'); // true
+```
+
+You can choose between alpha-2 and alpha-3, alpha-2 is set by default.
+
+***
+See also:
+
+  * [CountryCode](CountryCode.md)
--- a/docs/VALIDATORS.md
+++ b/docs/VALIDATORS.md
@@ -139,6 +139,7 @@
 
   * [CountryCode](CountryCode.md)
   * [CurrencyCode](CurrencyCode.md)
+  * [LanguageCode](LanguageCode.md)
   * [PostalCode](PostalCode.md)
   * [SubdivisionCode](SubdivisionCode.md)
   * [Tld](Tld.md)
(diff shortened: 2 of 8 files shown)


docs/VALIDATORS.md
```diff
@@ -139,6 +139,7 @@
 
   * [CountryCode](CountryCode.md)
   * [CurrencyCode](CurrencyCode.md)
+  * [LanguageCode](LanguageCode.md)
```
henriquemoody: Can you add this link in the alphabetically sorted list in this same file?

docs/LanguageCode.md
```diff
@@ -0,0 +1,20 @@
...
+
+- `v::languageCode()`
+
+Validates a language code based on ISO 639:
+
+```php
+v::languageCode()->validate('pt'); // true
+v::languageCode()->validate('en'); // true
+v::languageCode()->validate('it'); // true
+v::languageCode(LanguageCode::ALPHA3)->validate('ita'); // true
```
henriquemoody: I think that would be better just use the strings on the rule's constructor, like `"alpha-3"` instead. That's make a lot easy to use the rule and it's the best example we can give.

library/Rules/LanguageCode.php
```diff
@@ -0,0 +1,564 @@
...
+ * For the full copyright and license information, please view the "LICENSE.md"
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Rules;
+
+use Respect\Validation\Exceptions\ComponentException;
+
+/**
+ * Validates countries in ISO 639.
```
henriquemoody: Validates _languages_ in ISO 639.

PS.: That's one of the reasons why I hate documentation on code... The code must explain itself, but since we're an open source library, that kinda have to do this... Or not?

henriquemoody:
About the `AbstractIsoCode`, that is a great idea but I would be even better if we haven't released 1.0 yet. To create this abstract class we have to break the API, since `$countryCodeList` on `CountryCode` is protected. So, if the change it we must bump the next version to 2.0, so, for now, I'm :+1: to keep it as is but I will create an issue to be considered on the next major version or you can send the pull request with this change agains 2.0. What you think?

library/Rules/LanguageCode.php
batusa: lol

I think it's only for IDE purposes :P

docs/LanguageCode.md
batusa: ok

docs/VALIDATORS.md
batusa: sure!

batusa:
awesome @henriquemoody !

I''ll make the changes  needed here... brb ;)

henriquemoody:

PR #600, opened 2015-10-28
Validate the age to be always be an integer

### LINKED ISSUE #593: MinimunAge rule should not check if age is an integer type

diegoholiveira:
Closes #593

### DIFF
--- a/library/Rules/MinimumAge.php
+++ b/library/Rules/MinimumAge.php
@@ -20,7 +20,7 @@ class MinimumAge extends AbstractRule
 
     public function __construct($age, $format = null)
     {
-        $this->age = $age;
+        $this->age = filter_var($age, FILTER_VALIDATE_INT);
         $this->format = $format;
     }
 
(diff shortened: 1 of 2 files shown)


henriquemoody:
It's fixed already by b98595a5c093bfebd8686646fffb6b441fce6a08

However, thanks for contributing, @diegoholiveira, it's a pleasure to see you around! :panda_face:

diegoholiveira:
@henriquemoody: it would be nice to find a way to avoid two people working on the same issue.

henriquemoody:
That's true, unfortunately GitHub doesn't provide a smart way to do this. I'm going to write a process and put it on CONTRIBUTING.md file.

I'm sorry about the inconvenient. :confused:

diegoholiveira:
There is no problem about this. It's more important to have this issue solved, no matter who did it.

But this can happen again. And with something bigger than just a few lines... and this really concerns me.

henriquemoody:
Do you have any suggestions?

KennedyTedesco:
I guess this is something that we have to look before send a PR. Right? Mostly because doesn't exist a way to avoid this by using a tool, for example.

batusa:
let's take this issue to an issue :stuck_out_tongue: https://github.com/Respect/Validation/issues/605

PR #601, opened 2015-10-29
Create a test to Rule/AbstractSearcher

marcosptf:
Like there ( https://scrutinizer-ci.com/g/Respect/Validation/code-structure/master/class/Respect%5CValidation%5CRules%5CAbstractSearcher ) the coverage about this rule, is zero, then i've create this test to this abstract class.

It's my first contribution here, please, see if my code is correct or if i don't forget something!
Any comments are ever welcome!!!

Thanks
:-)

### DIFF
--- a/tests/unit/Rules/AbstractSearcherTest.php
+++ b/tests/unit/Rules/AbstractSearcherTest.php
@@ -19,31 +19,107 @@ namespace Respect\Validation\Rules;
 class AbstractSearcherTest extends \PHPUnit_Framework_TestCase
 {
 
-    public function testValidateMethodWhenArgsIsNull()
+    public function testValidateMethodWhenArgsIsNullCompareIdenticalFalse()
     {
         $compositeRuleMock = $this->getMockForAbstractClass('Respect\\Validation\\Rules\\AbstractSearcher');
         $compositeRuleMock->expects($this->any())
-             ->method('validate')
-             ->will($this->returnValue(true));
+                ->method('validate')
+                ->will($this->returnValue(true));
+        $compositeRuleMock->compareIdentical = false;
         $this->assertTrue($compositeRuleMock->validate(null));
     }
 
-    public function testValidateMethodWhenArgsIsEmptyWithDoubleQuotes()
+    public function testValidateMethodWhenArgsIsNullCompareIdenticalTrue()
+    {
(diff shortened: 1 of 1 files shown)


tests/unit/Rules/AbstractSearcherTest.php
```diff
@@ -0,0 +1,49 @@
...
+ */ 
+
+class AbstractSearcherTest extends \PHPUnit_Framework_TestCase
+{
+
+    public function testValidateMethodWhenArgsIsNull()
+    {
+        $compositeRuleMock = $this->getMockForAbstractClass('Respect\\Validation\\Rules\\AbstractSearcher');
+        $compositeRuleMock->expects($this->any())
+             ->method('validate')
```
batusa: You shouldn't have mocked the `validate` method as it's the concrete method we want to test. You only need to mock the abstract class (as you did), pass an array to the `$haystack` attribute and do some tests against the conditionals in the `validate` method.

Through the `validate` method, you'll be able to cover `validateEquals` and `validateIdentical`.

batusa: btw, coincidentally, I was working on this test today ^^

marcosptf:
Like this for example?

marcosptf:
ahhhhhhh
now i understand!!!!
thanks so much @batusa !!!!!

:D :D :D :D

henriquemoody:
I'm sorry by my late, guys, I was afk the latest days.

It seems this is duplicated since @batusa was working on it see https://github.com/Respect/Validation/issues/379#issuecomment-152268047, I must write a process and put it on the CONTRIBUTING.md file to avoid problems.

@marcosptf, I think prefer the changes of ff5a85fe06241bcba3c9dbb607cef73a04cbe62f which is already on master now.

However, thanks a lot for contributing to this project, I really appreciate that! :panda_face:

marcosptf:
@henriquemoody 

oohh noooooo!!!
i've work a lot in this branch


kid_crying_blog

henriquemoody:
I'm really really sorry about that, man... I know this feeling it happens to me too.

@batusa have created #605 then we can avoid problems like this.


Again, I'm really sorry

marcosptf:
hahahaahaahahahaa
okey
it's life
kkkkk

PR #602, opened 2015-10-30
Creation of tests for abstract classes

batusa:
I'll leave it open here. any suggestion of improvement is very welcome.


tests/unit/Rules/AbstractCompositeTest.php
```diff
@@ -150,4 +150,77 @@ public function testShouldNotDefineNameForInternalRulesWhenItHasAName()
...
+    }
+
+    public function testHasRuleShouldReturnTrueWhenRuleIsFound()
+    {
+        $simpleRuleMock = $this->getMock('Respect\\Validation\\Validatable');
+
+        $compositeRuleMock = $this->getMockForAbstractClass('Respect\\Validation\\Rules\\AbstractComposite');
+        $compositeRuleMock->addRule($simpleRuleMock);
+
+        $this->assertTrue($compositeRuleMock->hasRule($simpleRuleMock));
```
batusa: I would do here another *assertion* trying to check if there's a rule using a `string`. but I couldn't be able to do that because phpunit mocks are not working so fine with `get_class()` which is used internally in the method (https://github.com/Respect/Validation/blob/master/library/Rules/AbstractComposite.php#L92).

btw it's a known issue on phpunit repo:  https://github.com/sebastianbergmann/phpunit-mock-objects/issues/134

batusa:
I didn't test any protected/private methods that I couldn't reach through a public method. as an example would be `filterWhiteSpaceOption` in `AbstractCtypeRule`. any idea on how I could test it? (or shouln't I test it at all? :P)

henriquemoody:
Merged manually on master!

Thanks for the hard work, @batusa, you rocks! :panda_face:

PR #604, opened 2015-11-04
Use standard phrasing for max/min exception messages

amp343:
Existing exception messages for max/min validators are:

**Max:**
```
must be greater than [or equals]
```
**Min:**
```
must be lower than [or equals]
```

This PR updates those messages to use more standard comparison phrasing:

**Max:**
```
must be greater than [or equal to]
```

**Min:**
```
must be less than [or equal to]
```


henriquemoody:
Merged manually by 347f4c9fc8d4c08eb834622bd72d2a699e05e8d6


Thanks for contributing, much better like this! :panda_face:

PR #607, opened 2015-11-06
Created new Image rule

guilhermesiani:


guilhermesiani:
When I execute tests, the message "'Respect\Validation\Rules\RuleTestCase' not found" is showing for many classes. Is thera a change on test that wasn't reported?

batusa:
try to dump your composer autoload again.

`composer dump-autoload`

guilhermesiani:
It works! thanks. I'll commit again.

library/Exceptions/ImageException.php
```diff
@@ -0,0 +1,24 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Exceptions;
+
+class ImageException extends ValidationException
+{
+    public static $defaultTemplates = [
+        self::MODE_DEFAULT => [
+            self::STANDARD => '{{name}} must be a valid image URL',
```
batusa: is it aimed just for urls? I think you could use it to check local files too.

library/Rules/Image.php
```diff
@@ -0,0 +1,20 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Rules;
+
+class Image extends AbstractRule
+{
+    public function validate($input)
+    {
+        return @getimagesize($input) ? true : false;
```
batusa: to avoid this `@` you could do something like:

```php
if (is_file($input) || filter_var($input, FILTER_VALIDATE_URL)) {
    return is_array(getimagesize($input)); // as it returns an array in case of success...
}

return false;
```

batusa: I've put the the *if* statement this way but maybe it would be better to use Respect rules for both conditions...

[Url](https://github.com/Respect/Validation/blob/master/library/Rules/Url.php)
[File](https://github.com/Respect/Validation/blob/master/library/Rules/File.php)

guilhermesiani:
Thanks, Batusa! I followed your tips. It is really better

williamespindola:
@guilhermesiani can you do the integration tests for this rule?

tests/unit/Rules/ImageTest.php
```diff
@@ -0,0 +1,53 @@
...
+    public function testInvalidImageUrlShouldNotValidate($input)
+    {
+        $instance = new Image();
+        $this->assertFalse($instance->validate($input));
+    }    
+
+    public function providerForValidInput()
+    {
+        return [
+            ['http://lorempixel.com/400/200/'],
```
henriquemoody: You cannot use an external resource on a unit test, event for the PHPT I try to avoid third-party resource, although you can add an image to test it.

library/Rules/Image.php
```diff
@@ -0,0 +1,24 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Rules;
+
+class Image extends AbstractRule
+{
+    public function validate($input)
+    {
+    	if (is_file($input) || filter_var($input, FILTER_VALIDATE_URL)) {
```
henriquemoody: I think this is to specific to be called _Image_, so it cannot accept local messages but only image URL?

library/Rules/Image.php
```diff
@@ -0,0 +1,24 @@
...
+     	   return is_array(getimagesize($input));
```
henriquemoody: I don't think a good idea to require GD installed as dependency just to execute a validation rule, instead you can use [finfo](http://php.net/finfo) which is already in use by [Mimetype](https://github.com/Respect/Validation/blob/master/library/Rules/Mimetype.php) rule.

batusa: according to the manual, `getimagesize` doesn't require GD. http://php.net/manual/en/function.getimagesize.php#refsect1-function.getimagesize-notes

batusa: but anyways, I think your alternative is a valid one ;)

batusa: Even the manual says the same... I hadn't realized it lol:
"Do not use getimagesize() to check that a given file is a valid image. Use a purpose-built solution such as the Fileinfo extension instead. "

henriquemoody: It says that this function doesn't require the GD library but not the GD extension, I'm not sure, though, I have to test it. But if you see the paragraph right below this note you will see this:

> This function expects filename to be a valid image file. If a non-image file is supplied, it may be incorrectly detected as an image and the function will return successfully.
> Do not use getimagesize() to check that a given file is a valid image. Use a purpose-built solution such as the Fileinfo extension instead.

batusa: yes I just saw it! good to know! :)

henriquemoody: I'm sorry, GitHub doesn't update it real time, I was writing already :D

batusa: no problem! :stuck_out_tongue:

guilhermesiani:
Nice to know about these points! Thanks. I'll make the changes as soon as possible

guilhermesiani:
Where can I put the image file just to unit test?

henriquemoody:
`tests/fixtures` would be the proper directory.

guilhermesiani:
Finally, I updated the files. excuse me for delay

henriquemoody:
There were lots of things missing on this pull request, then I send #621 and merged it.

PR #608, opened 2015-11-06
Fix perfect square validate

fefas:
I noted a problem in the [PerfectSquare rule](https://github.com/Respect/Validation/blob/master/library/Rules/PerfectSquare.php) validation logic.

Mathematically, `sqrt(x) * sqrt(x)` will be always equal to `x` even if `x` is a *perfect square* or not. The unit tests do not show any problem because the *floating point comparison problem*.

The right way to check if a given number `x` is perfect square or not is checking if `sqrt(x)` is an integer or not.

I hope this changes can upgrade `Respect\Validation` package :D


tests/unit/Rules/PerfectSquareTest.php
```diff
@@ -57,7 +57,7 @@ public function providerForPerfectSquare()
             ['0'],
             [81],
             [0],
-            [250],
+            [2500],
```
batusa: you could then add 250 to `providerForNotPerfectSquare`

henriquemoody: 250 is not a perfect square because it cannot be expressed as the product of two equal integers

henriquemoody:
Merged manually by e4622df22cfd6822b43e6a8607b98b3c964b3fe3

fefas:
:+1:

PR #614, opened 2015-11-25
Fix PL postal code format and examples

regdos:
http://www.geonames.org/countries/PL/republic-of-poland.html


henriquemoody:
Merged manually and already available on versions 1.0.3, 0.9.6, 0.8.12.

PR #630, opened 2016-01-13
Fix homepage URL in composer config

shadowhand:


henriquemoody:

PR #641, opened 2016-01-29
Recursive SetName for Not

### LINKED ISSUE #640: Recursive SetName for Not
Guys, what do you think about the possibility of recursive **setName** for **Not**?

something like this:

```php
$int = new IntVal();
$not = new Not($int);

$not->setName('Foo');

$int->getName(); // return 'Foo';
```

caiotava:
Fix #640

### DIFF
--- a/library/Rules/Not.php
+++ b/library/Rules/Not.php
@@ -27,7 +27,7 @@ class Not extends AbstractRule
     {
         $this->rule->setName($name);
 
-        parent::setName($name);
+        return parent::setName($name);
     }
 
     public function validate($input)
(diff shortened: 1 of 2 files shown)


library/Rules/Not.php
```diff
@@ -23,6 +23,13 @@ public function __construct(Validatable $rule)
         $this->rule = $rule;
     }
 
+    public function setName($name)
+    {
+        $this->rule->setName($name);
+
+        parent::setName($name);
```
henriquemoody: `setName()` must return self as the same of `AbstractRule`, please, `return $this` in the end of the method.

caiotava:
Sorry, fixed. :smile:

henriquemoody:
Merged manually by 3c75bbc0bde8055e4b0175fe28fe95352b080a40

PS.: I'm so sorry by my late.

PR #643, opened 2016-01-30
Length gets length of integers

Iverum:


library/Rules/Length.php
```diff
@@ -64,6 +64,10 @@ protected function extractLength($input)
         if (is_object($input)) {
             return count(get_object_vars($input));
         }
+        
+        if (is_int($input)) {
+            return mb_strlen((string)$input, mb_detect_encoding((string)$input));
```
henriquemoody: The use mbstring here is not necessary, since integers cannot have different encodings. Just use `strlen()`.

henriquemoody:
Merged manually by 1bd81141500179b47b1b2fdff7b7f20b0c68a7ed

PS.: I'm sorry by my late.

PR #644, opened 2016-02-16
Ensure namespace separator on appended prefixes

augustohp:
Appending a prefix to search new rules under required that the namespace
(prefix) being added always ended with a trailing namespace character so
rules could successfully be found under it. This ensures that the
separator is always present.

Changes a test for a rule which does not implement Respect's interface
to an actual class so we don't need to declare one to use as a stub.

### DIFF
--- a/library/Factory.php
+++ b/library/Factory.php
@@ -25,7 +25,12 @@ class Factory
 
     public function appendRulePrefix($rulePrefix)
     {
-        array_push($this->rulePrefixes, $rulePrefix);
+        $namespaceSeparator = '\\';
+        $rulePrefix = rtrim($rulePrefix, $namespaceSeparator);
+        array_push(
+            $this->rulePrefixes,
+            $rulePrefix . $namespaceSeparator
+        );
     }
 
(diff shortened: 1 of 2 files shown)


henriquemoody:
The name _prefix_ instead of _namespaces_ was used with the intention of allow prefixes (not namespaces), then people would not need to use a namespace.

I'm not against that change, but then I think we should rename it to `appendRuleNamespace()`.

library/Factory.php
```diff
@@ -25,7 +25,12 @@ public function getRulePrefixes()
 
     public function appendRulePrefix($rulePrefix)
     {
-        array_push($this->rulePrefixes, $rulePrefix);
+        $namespaceSeparator = '\\';
+        $rulePrefix = rtrim($rulePrefix, $namespaceSeparator);
```
henriquemoody: If that's the case that will be necessary on `prependRulePrefix()` as well.

henriquemoody:
Merged manually by 9460a4c237818bf7a445202c5ca2acab66f6120c

Thanks for contributing! 🐼

Issue #647, opened 2016-02-26
AbstractWrapper doesn't pass on setName calls to the wrapped Validatable


alnorth:
When using something like `SubdivisionCode` `setName` calls aren't passed on to the particular class that's wrapped (e.g. `UsSubdivisionCode`). This results in the input value being used as the name in error messages.

I'm working on a PR for this and will add a fix with a test.

alnorth:
Ignore the first commit (559d82c). b20bb93 fixes this issue.

henriquemoody:
@alnorth, can you send a pull request against 1.0 branch?

alnorth:
Will do. (Although your docs seemed to say that bugfixes should be submitted against 0.8).

henriquemoody:
Hum... I will try to make it more clear on our documentation.

alnorth:
It was "Pull requests for bug fixes must be based on the oldest supported version's branch" in the contributing doc plus Example 1 here: https://github.com/Respect/Validation/wiki/Release-Cycle-and-Support.

Pull request submitted as #648.

henriquemoody:
Yes, it must be the _latest_ supported version which is 1.0.

henriquemoody:
I'm sorry, you're right. But thanks any way!

alnorth:
No problem :smile:

PR #648, opened 2016-02-26
AbstractWrapper should pass on setName calls to the wrapped Validatable.

alnorth:
When using something like SubdivisionCode setName calls aren't passed on to the particular class that's wrapped (e.g. UsSubdivisionCode). This results in the input value being used as the name in error messages.

Fixes issue #647.


henriquemoody:

Versions 0.8.13, 0.9.7 and 1.0.4 are already available with this fix.

alnorth:
Fast work! Thank you :+1:

henriquemoody:
Thanks to you

Issue #650, opened 2016-02-27
Validation credit card with brand


reginaldojunior:
Create a validator to brand of the cards,

Example

```php
v::creditCardVisa()->validate('4111 1111 1111 1111');
```

or 

```php
v::creditCard('visa')->validate('5376 7473 9720 8720');
```

Is viable or not is a feature good?

PR #651, opened 2016-02-29
fix issue #650

reginaldojunior:


library/Exceptions/CreditCardVisaException.php
```diff
@@ -0,0 +1,24 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Exceptions;
+
+class CreditCardVisaException extends ValidationException
+{
+    public static $defaultTemplates = array(
+        self::MODE_DEFAULT => array(
+            self::STANDARD => '{{name}} must be a valid Credit Card visa number valid',
```
henriquemoody: Credit Card visa => Visa credit card

library/Exceptions/CreditCardVisaException.php
```diff
@@ -0,0 +1,24 @@
...
+        ),
+        self::MODE_NEGATIVE => array(
+            self::STANDARD => '{{name}} must not be a valid Credit Card visa number',
```
henriquemoody: Same of [my last comment](https://github.com/Respect/Validation/pull/651/files#r54555387)

library/Rules/CreditCardVisa.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+namespace Respect\Validation\Rules;
+
+class CreditCardVisa extends CreditCard
+{
+
+    public function validate($input)
+    {
+        $input = preg_replace('([^0-9])', '', $input);
```
henriquemoody: A credit card like `This is a credit card: 1234 1234 1234 1234` will pass.

The best option would be to define the separator in the constructor and just replace the separator for `''`.

library/Rules/CreditCardVisa.php
```diff
@@ -0,0 +1,43 @@
...
+        if (!$this->verifyMod10($input)) {
+            return false;
+        }
+
+        return $this->verifyIfCardIsVisa($input);
+    }
+
+    private function verifyIfCardIsVisa($input)
+    {
+        $pattern = "/^([4]{1})([0-9]{12,15})$/";//Visa
```
henriquemoody: The groups are only necessary if you wanna use them, in this case they're not necessary, you can simply have:

```regex
/^4[0-9]{12,15}$/
```

library/Rules/CreditCardVisa.php
```diff
@@ -0,0 +1,43 @@
...
+
+        if (preg_match($pattern, $input)) {
+            return true;
+        }
+
+        return false;
```
henriquemoody: You can replace all returns by:

```php
return (preg_match($pattern, $input)) > 0);
```

library/Rules/CreditCardVisa.php
batusa: a few lines of code here are actually the same in the method `validate` of the class `CreditCard`.

what do you think about:
```php
return parent::validate($input) && $this->verifyIfCardIsVisa($input);
```
?

henriquemoody: Even better, then we don't need to make `verifyMod10()` protected.

batusa: ;)

batusa: I was thinking about an abstract class with an abstract method where this method would have the code to validate a certain brand, so you would inherit from this abstract class and just code the part about the brand itself. doing so, the validate method would be untouchable. 

what do you think?

reginaldojunior:
@henriquemoody @batusa Fixes done and class abstract implemented :smile:

reginaldojunior:
Fixes done and class abstract implemented 

@henriquemoody @batusa

henriquemoody:
Hey, @reginaldojunior, be sure that your contribution makes the pandas happy! 🐼 

First of all, thanks a lot for contributing and I'm sorry I took so long to comment here. 

I was thinking about this approach and turns out that it doesn't seem the best for us. Since we can have many different credit card brands, having one rule for each of them seems impracticable.

With that in mind I created #661 and I would appreciate very much some feedback from you and @batusa, if possible, of course.

batusa:
@henriquemoody you made it less complex and that's good!!

I think that we should avoid case-sensitive mistakes in the `$brand` parameter by doing some sort of filtering in the parameter so that things like:
```php
v::CreditCard('visa')->validate($input);
``` 

would work too.

What do you think?

henriquemoody:
I decided to use natural language, but my first thought was:

- `amex`: American Express
- `diners`: Diners Club
- `discover`: Discover
- `jcb`: JCB
- `mastercard`: MasterCard
- `visa`: Visa

The case-sensitive may be a problem but there are spaces as well. The best option is just show the available options in case of error, IMHO.

batusa:
natural language would be good for the `self::BRANDED` templates. then in the constructor we could use the way you showed in the list to make it simpler.

reginaldojunior:
Ow! I liked the way you did. :smile: :panda_face:

henriquemoody:
Thanks guys for your feedback! 🐼

Issue #656, opened 2016-03-29
Missing functionality


SimTabi:
Hi,

I think you missed something i am using the latest version. Seems like `languageCode() `is missing.

batusa:
thanks for reporting that @SimTabi , indeed `LanguageCode` rule is not in the branch `1.0`.

henriquemoody:
Actually that's not a but. This rule was introduced after the release of version 1.0, so it's only available on master that will be released in the next version on April 24th.

PR #657, opened 2016-03-29
Adds "LanguageCode" rule to branch 1.0

batusa:
@henriquemoody I don't know if it's really the right way to do that. By #656 I saw that LanguageCode rule was missing in the branch 1.0, so everything I did was just to "cherry-pick" the [commit](https://github.com/Respect/Validation/commit/621b4707b59fe38a6199461b0ce605e7dffd2881) of this rule in the branch 1.0.

### DIFF
--- a/docs/LanguageCode.md
+++ b/docs/LanguageCode.md
@@ -1,0 +1,20 @@
+# LanguageCode
+
+- `v::languageCode()`
+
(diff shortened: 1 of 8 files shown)


henriquemoody:
Hum... I merged it but that was a mistake, I will explain that on  #656

PR #660, opened 2016-04-06
Fix missing change from arr() to arrayVal()

rubenv:


henriquemoody:
Thanks for contributing!! 🐼

PR #663, opened 2016-04-07
Set template for the first rule in the chain

### LINKED ISSUE #662: callback setTemplate works on assert but not with check #bug
Hi 

I found a bug with setTemplate even after https://github.com/Respect/Validation/commit/3ed3114e58b2f516d53d8656efc47c41ce5d824b

This is from unit tests (`assert`)

```php
try { 
  v::callback('is_int')->setTemplate('{{name}} is not tasty')->assert('something'); 
} catch(NestedValidationException $e) { 
  print($e->getFullMessage()); 
}
```
returns: `"something" is not tasty - "something" must be valid`

Which is "good".

But then i do the same validation with `check`

```php
try { 
  v::callback('is_int')->setTemplate('{{name}} is not tasty')->check('something'); 
} catch(NestedValidationException $e) { 
  print($e->getFullMessage()); 
}
```
returns: `- "something" must be valid`

The custom template is gone.

henriquemoody:
When there is just one rule in the chain and the there is a defined template for that, the expected behaviour when using the `check()` method is to see the exception message with the defined template.

***
Fix #662

### DIFF
--- a/library/Validator.php
+++ b/library/Validator.php
@@ -15,6 +15,7 @@ use finfo;
 use ReflectionClass;
 use Respect\Validation\Exceptions\AllOfException;
 use Respect\Validation\Exceptions\ComponentException;
+use Respect\Validation\Exceptions\ValidationException;
 use Respect\Validation\Rules\AllOf;
 use Respect\Validation\Rules\Key;
 
@@ -182,6 +183,19 @@ class Validator extends AllOf
             self::getFactory()->appendRulePrefix($rulePrefix);
         } else {
             self::getFactory()->prependRulePrefix($rulePrefix);
+        }
+    }
+
+    public function check($input)
+    {
+        try {
+            return parent::check($input);
+        } catch (ValidationException $exception) {
+            if (count($this->getRules()) == 1 && $this->template) {
+                $exception->setTemplate($this->template);
+            }
+
+            throw $exception;
         }
     }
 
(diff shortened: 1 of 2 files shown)


batusa:
why when there's only just *one*?

I'm asking this because of the following tests:

[set_template_with_multiple_validators_should_use_template_as_full_message](https://github.com/Respect/Validation/blob/master/tests/integration/set_template_with_multiple_validators_should_use_template_as_full_message.phpt)
[set_template_with_multiple_validators_should_use_template_as_main_message](https://github.com/Respect/Validation/blob/master/tests/integration/set_template_with_multiple_validators_should_use_template_as_main_message.phpt)

henriquemoody:
This is kinda tricky.

When you have `v::something()` you may thing: _I am just using the "something" rule_. But actually you have a `something` rule in a `allOf` rule because `Validator` extends  `AllOf` and `Validator::__callStatic()` creates a new `Validator` instance.

When you execute the `setTemplate()` method, like `v::something()->setTemplate('Yo, man!')`, you're actually setting the template for the `allOf` rule not for the `something()`, but it still feels like _I am just using the "something" rule_, thus the expected behavior when using the `check()` method with an invalid input is to catch an exception using the template you defined, IMO.

However this change is too ugly for my taste, I'm still not convinced to merge it yet. 😛

batusa:
> thus the expected behavior when using the check() method with an invalid input is to throw an exception using the template you defined, IMO.

I agree!

I'm only wondering why you're checking if there's only one rule. Setting a template when using more than one rule would have any problem? That's why I showed you [this](https://github.com/Respect/Validation/blob/master/tests/integration/set_template_with_multiple_validators_should_use_template_as_main_message.phpt).

henriquemoody:
When the chain has more than one rule it's okey to set the template for the main rule instead.

If someone uses a chain with multiple rules and uses the `assert()` method he/she is going to use `getFullMessage()`, otherwise using the `check()` method would be better and then  `getMainMessage()`.

PR #668, opened 2016-04-25
Drop support for PHP 5 and HHVM

henriquemoody:
Ping @Respect/core, @Respect/respect-validation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,7 @@ language:
   php
 
 php:
-  - 5.4
-  - 5.5
-  - 5.6
   - 7
-  - hhvm
 
 before_script:
   - composer self-update
@@ -19,8 +15,8 @@ script:
   - vendor/bin/phpunit --configuration phpunit.xml.dist --colors --coverage-clover=coverage.clover
 
 after_script:
-  - test "5.6" == "${TRAVIS_PHP_VERSION}" && wget https://scrutinizer-ci.com/ocular.phar
-  - test -f ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover
+  - wget https://scrutinizer-ci.com/ocular.phar
+  - php ocular.phar code-coverage:upload --format=php-clover coverage.clover
 
 notifications:
   irc:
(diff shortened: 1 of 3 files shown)


augustohp:
I have no opinion about HHVM, it is a nice to have but I am not familiar with the trouble to maintain it.

Regarding PHP 5.6, its supported as a stable release will go until [2019](http://php.net/supported-versions.php). Although I see some benefit from certain PHP 7 features, I would keep support for PHP 5.6 as long as we can and tie the drop of support to the commit that uses a feature only present on 7.

henriquemoody:
Actually if we keep the support for 5.6 we're able to keep the support for HHVM, otherwise we might have problems with that. I wish we could mesure the number of users who use HHVM and these other PHP versions. What I'm saying is that makes sense to keep support for PHP 5.6 but I badly want to start doing PHP 7 code on Validation, specially type declarations, maybe something with `AssertionError` and so on.

Although, from the user's perspective it's better to keep as much compatibility as we can.

lcobucci:
Considering that its a major release it would be really good to adopt the PHP 7 features. The PHP 5 users will still be able to use the old API with no harm.

augustohp:
> What I'm saying is that makes sense to keep support for PHP 5.6 but I badly want to start doing PHP 7 code on Validation, specially type declarations, maybe something with AssertionError and so on.

I don't see type hints with good eyes, specially strict enforcing them. There are many rules which benefit from handling multiple type of data.

I would go PHP >= 7 if there is something that would really make our life easier and/or improve the library to a whole new level. Although I do see benefits from Anonymous Classes for some ideas I have, I don't see them as reason enough to adopt >= 7.

> Although, from the user's perspective it's better to keep as much compatibility as we can.

That is my main concern. Keeping compatibility with a no longer supported version is one thing, droping a stable one is another.

> Considering that its a major release it would be really good to adopt the PHP 7 features.

Adopting features just by doing so is not a reasonable thing to do, I am also eager to do so but I think we need to see the real benefits behind it.

> The PHP 5 users will still be able to use the old API with no harm.

Who is going to keep it and fix bugs? Having two code bases is not a solution to the problem.

PR #669, opened 2016-04-26
Validate month format using UNIX epoch

augustohp:
When using `date` rule to validate only the month portion of a date,
depending of *when* you validate something you might [have a strange
behavior][1].

If you want to validate `v::date('m')->validate('02')` for instance, if
you are on the 30th or 31th of a month you will get a `false` as an
answer. This happens because we use `DateTime::createFromFormat` and it
uses the current time for portions of the format which are omited.
So, the following happens internally:

    php > echo date('Y-m-d', strtotime('2015-02-30'));
    2015-03-02

As the month given by the `DateTime` object will be "03", the date check
now fails.
This fixes the problem by using `!` in the format (if it is only related
trying to validate a month) which forces the `createFromFormat` method
to use the Unix Epoch as the rest of the date instead of the current
time. Making this happen:

    php > echo date('Y-m-d', strtotime('1970-02-01'));
    1970-02-01

The test made will only break on 30th and 31th (and 29th depending on
the year), so it is not perfect.

  [1]: http://stackoverflow.com/questions/29346953

### DIFF
--- a/library/Rules/Date.php
+++ b/library/Rules/Date.php
@@ -35,6 +35,10 @@ class Date extends AbstractRule
         $exceptionalFormats = [
             'c' => 'Y-m-d\TH:i:sP',
             'r' => 'D, d M Y H:i:s O',
+            'm' => '!m',
+            'n' => '!n',
+            'F' => '!F',
+            'M' => '!M'
         ];
 
         if (in_array($this->format, array_keys($exceptionalFormats))) {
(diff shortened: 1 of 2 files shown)


batusa:
I thought @henriquemoody had fixed that https://github.com/Respect/Validation/commit/eecc696792ea7ced66d6bbdbe08247e31d6ba951

I warned about it once when he found it https://twitter.com/batusa/status/715631696343908352

it was exactly a Mar 31th :P

augustohp:
Oh my! I did the rebase before pushing it and did not bother to see the changes between my HEAD and the upstream HEAD. Sorry about that.

PR #671, opened 2016-04-27
Added Method validation

micmania1:
This adds the ability to validate a class instance based on method return values. Its based off the `Attribute` rule.

The use case for this is classes which must implement an interface. In these cases, you can't rely on properties existing so the attribute rule can't be used.

Example:
```
interface MyInterface
{
    public function getMyValue();
}

class MyClass implements MyInterface
{
    protected $myValue = 'My value';
    public function getMyValue()
    {
        return $this->myValue;
    }
}

class MyOtherClass implements MyInterface
{
    protected $dataSource;
    public function getMyValue()
    {
        return $this->dataSource->get('myValue').
     }
}
````

I'd be happy to write some docs for this but would like to hear some feedback before moving forward.

Thanks.

### DIFF
--- a/library/Exceptions/MethodException.php
+++ b/library/Exceptions/MethodException.php
@@ -1,0 +1,33 @@
+<?php
+
+/*
+ * This file is part of Respect/Validation.
+ *
+ * (c) Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
+ *
+ * For the full copyright and license information, please view the "LICENSE.md"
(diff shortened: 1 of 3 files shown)


henriquemoody:
You can archive the same thing with  [Call](https://github.com/Respect/Validation/blob/master/docs/Call.md).

If you still want to contribute, you can create more docs on the document I just sent!

However, thanks for contributing! 🐼

Issue #670, opened 2016-04-27
KeySet error if input is not an array


pengu-fr:
Using KeySet rule to validate anything else than an array with raise an error :

`v::keySet(v::key('test'))->assert(null);`
> array_key_exists() expects parameter 2 to be array, null given

I think hasValidStructure function should simply return false if input is not an array :

```
private function hasValidStructure($input)
    {
        if (!is_array($input)){
            return false;
        }
        foreach ($this->getRules() as $keyRule) {
            if (!array_key_exists($keyRule->reference, $input) && $keyRule->mandatory) {
                return false;
            }

            unset($input[$keyRule->reference]);
        }

        return (count($input) == 0);
    }
```

batusa:
Hi @pengu-fr ,

thank you for reporting this issue! a patch is on the way!

henriquemoody:
Fixed on versions 1.0.6 and 1.1.1. 

Thanks for reporting! 🐼

PR #672, opened 2016-04-27
Fixes 'KeySet' rule when input is not array type

batusa:
Reported on #670

### DIFF
--- a/library/Rules/KeySet.php
+++ b/library/Rules/KeySet.php
@@ -87,6 +87,10 @@ class KeySet extends AllOf
      */
     private function hasValidStructure($input)
     {
+        if (!is_array($input)) {
+            return false;
+        }
+
         foreach ($this->getRules() as $keyRule) {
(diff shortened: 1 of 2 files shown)


library/Rules/KeySet.php
```diff
@@ -87,6 +87,10 @@ public function getKeys()
      */
     private function hasValidStructure($input)
     {
+        if (!is_array($input)) {
+            return false;
```
henriquemoody: I didn't think about it, but indeed this is a fair approach. The `Key` also only accepts arrays, so doesn't make sense to accept `ArrayAccess` or `Iterator` instances on this one.

But maybe this is something we could think about in the future, what you think?

henriquemoody:
Merged manually by d9a4c78b163079f2210f7560156b3924e7e909f1 on branches `1.0`, `1.1` and `master`.

Thanks for contributing! 🐼

library/Rules/KeySet.php
batusa: yeah, sure!

PR #679, opened 2016-05-13
replacing cnh algorithm validation

gpedro:
The current algorithm is wrong, i made a [PoC](https://github.com/gpedro/Validation/commit/ee2311d1fdcbdb70df3b25bbad5adc8296f72058) for testing. Changing the check digit for a wrong value, the tests passed. therefore, the algorithm is not correct. For the test fails, it is necessary that one digit is different from the result obtained by the algorithm. 

---

the old algorithm not working with some cnh values, e.g 00641829409. so i've found this algorithm at [iMasters](http://forum.imasters.com.br/topic/400293-validate-br/)


gpedro:
@henriquemoody PR ready.

tests/unit/Rules/CnhTest.php
```diff
@@ -98,30 +98,30 @@ public function validCnhProvider()
...
-               ['14324384312'],
-               ['14339482949'],
-               ['11136521151'],
-               ['11612581127'],
-               ['11613454741'],
-               ['14129251992'],
-               ['13411741211'],
-               ['13417248311'],
-               ['11671431345'],
-               ['13292694415'],
```
henriquemoody: Thanks for contributing! 

First of all, I'm sorry by my late. Secondly, I didn't get exactly why the current algorithm is not correct, can you elaborate?

Also, the numbers removed from here, should all them be valid? If so, they must be put on the `validCnhProvider()` but I need some clarification about why.

gpedro: @henriquemoody I will explain to you in more detail via email, because I need to attach documents to prove it.

henriquemoody:
Merged by #679

Thanks for contributing! 🐼

Issue #683, opened 2016-05-30
Date validator cannot handle valid timestamps unless they are cast to string


chetzof:
Does not pass validation
```
Validator::date('U')->validate(time());
```

Passes validation
```
Validator::date('U')->validate((string)time());
```

batusa:
thanks for reporting that @chetzof ! I'll have a look at it soon.

henriquemoody:
Fixed on versions 1.0.13 and 1.1.8.

Thanks for reporting! 🐼

PR #684, opened 2016-05-31
Date rule accepting also integer type as input argument

batusa:
It solves the issue #683


library/Rules/Date.php
```diff
@@ -26,7 +26,7 @@ public function validate($input)
     {
         if ($input instanceof DateTime) {
```
batusa: we got an issue here.

how could we compare if the given DateTime is compatible with the rule format (if provided)?

henriquemoody:
Closed by #731.

Thanks a lot, man! 🐼

PR #691, opened 2016-06-13
Fix NL postal code regex.

HDsign:
The dutch postal code can either be with of without a space between the numbers and the letters.


batusa:
lowercase letters should be accepted too?

could you please provide some tests to cover your change?

thank you!

HDsign:
lowercases are aloud. Most of the time one will format the postal code with uppercase letters. Validation should be okay. I have added some tests.

PR #692, opened 2016-06-13
Fix NL postal code regex. (Tests)

HDsign:
Added some tests for: The dutch postal code can either be with of without a space between the numbers and the letters.


batusa:
your tests failed because you created them without changing the rule code (as you did in #691). could you send your change and tests in the same pull request?

HDsign:
Oh no, stupid! 👍 I will create a new one.

PR #694, opened 2016-06-13
Fix for the dutch postal codes

HDsign:
Dammit! Github sure making me look good


library/Rules/PostalCode.php
```diff
@@ -113,7 +113,7 @@ class PostalCode extends Regex
         "NF" => "/^(\d{4})$/",
         "NG" => "/^(\d{6})$/",
         "NI" => "/^(\d{7})$/",
-        "NL" => "/^(\d{4}[A-Z]{2})$/",
+        'NL' => "/^(\A[1-9]\d{3} ?[a-zA-Z]{2}\z)$/",
```
henriquemoody: Is there any document that explains that lowercase are allowed?

HDsign: I haven't been able to find such information yet. The only official thing I found is this: http://www.postnl.nl/klantenservice/veelgestelde-vragen/versturen/postcodes-adressen/hoe-is-een-postcode-in-nederland-opgebouwd.html
For the rest of it, there's only Wikipedia's on the dutch zipcode. I've requested some more information on the matter by the official dutch postal company.

I can only say from experience most websites allow this regex on forms and stuff. For presentation they most of the time use this format '0000 AA'. But for form validation purposes we think it shouldn't fail.

So I'm still waiting on this official information.

henriquemoody: Hey @HDsign, do you have any news on it? At least the company I work in The Netherlands we don't allow lower case chars.

HDsign: It's PostNL, they don't provide any information on this. Never mind the change then.

PR #696, opened 2016-06-22
Fix Rules Roman

moura137:


tests/unit/Rules/RomanTest.php
```diff
@@ -60,7 +60,7 @@ public function providerForRoman()
             ['MCCXXXIV'],
             ['MMXXIV'],
             ['MCMLXXV'],
-            ['MMMMCMXCIX'],
```
henriquemoody: MMMMCMXCIX is 4999

***
https://www.quora.com/If-the-Roman-numeral-MMMMCMXCIX-is-4999-isnt-MMMMIM-just-as-valid-as-4999

PR #697, opened 2016-06-22
Fix Rules Cnpj

moura137:


tests/unit/Rules/CnpjTest.php
```diff
@@ -99,12 +99,13 @@ public function providerInvalidFormattedCnpj()
     public function providerInvalidUnformattedCnpj()
     {
         return [
-            ['11111111111'],
-            ['22222222222'],
-            ['12345678900'],
-            ['99299929384'],
-            ['84434895894'],
-            ['44242340000'],
```
henriquemoody: Those number are still invalid you shouldn't not remove then, just add the new ones.

moura137:
Hi @henriquemoody,
I changed because the previous numbers were 11 characters this way the numbers are unnecessary

library/Rules/Cnpj.php
```diff
@@ -19,7 +19,7 @@ public function validate($input)
         $c = preg_replace('/\D/', '', $input);
         $b = [6, 5, 4, 3, 2, 9, 8, 7, 6, 5, 4, 3, 2];
 
-        if (strlen($c) != 14) {
+        if (strlen($c) != 14 || preg_match("/^{$c[0]}{14}$/", $c)) {
```
henriquemoody: There is no way that of a single char (the first one, `$c[0]`) be 14 digits.

henriquemoody:
I didn't get the real intent of this PR, can you elaborate, @moura137?

library/Rules/Cnpj.php
malukenho: @henriquemoody He's trying to avoid repeated strings of only one digit like `11111111111111`, so `/^1{14}$/` is correct, btw, implementation sucks IMHO.

henriquemoody:
Closed by #780

Thanks for contributing! 🐼

PR #699, opened 2016-07-01
Fix Pesel validator when first char is zero

regdos:


henriquemoody:
Merged manually by 73a9ff40eb12e88ecee4ebf911ebe228c390c4d3.

Thanks for contributing! 🐼

PR #707, opened 2016-07-21
Added NAN validation

iamraccoon:
This adds the ability to validate NAN values.


library/Rules/NanType.php
```diff
@@ -0,0 +1,26 @@
...
+
+class NanType extends AbstractRule
+{
+    public function validate($input)
+    {
+        if (is_numeric($input)) {
+            return is_nan($input);
+        }
+
+        throw new ComponentException('The value must be a numeric.');
```
henriquemoody: The `validate()` method doesn't throw exceptions.

henriquemoody:
Why not use `v::not(v::numeric())`?

henriquemoody:
Actually after taking a close look at this I realized what it really means. A _numeric_ value may also be an unrepresentable number, that's why `is_nan()` is for. 

However, this rule would be more useful if called _Number_, if you revert the operation. Can you provide this change among with the other request?

wcbrown11:
Please post a job for free at careerduh.byethost31.com/

Skip the payment form.

Thank you,
William - manager 
CareerDuh

henriquemoody:
Thanks for contributing! 🐼

PR #709, opened 2016-07-30
replace deprecated getMock with createMock in test cases

royallthefourth:
It's better to have those unit tests pass without warnings :sunglasses:


henriquemoody:
Thanks for contributing! 🐼

Issue #717, opened 2016-08-19
Please Help! `Fatal error: Uncaught Error: Call to undefined function Respect\Validation\Rules\mb_strlen() `


CasperHK:
When I run the following method, I got error `Fatal error: Uncaught Error: Call to undefined function Respect\Validation\Rules\mb_strlen() `.

``` php
<?php
    require_once("./vendor/autoload.php");
    use Respect\Validation\Validator as v;

    class InputValidator {
        public static function isUsernameCorrect($username) {
            $usernameValidator = v::alnum()->noWhitespace()->length(1, 15);
            $result = $usernameValidator->validate($username);
            return $result;
        }
    }
?>
```

I called it in this way:

``` php
var_dump(InputValidator::isUsernameCorrect("casper1234"));
```

### Error Message

> Fatal error: Uncaught Error: Call to undefined function Respect\Validation\Rules\mb_strlen() in /var/www/html/PHPSimVlidator/vendor/respect/validation/library/Rules/Length.php:57 Stack trace: #0 /var/www/html/PHPSimVlidator/vendor/respect/validation/library/Rules/Length.php(49): Respect\Validation\Rules\Length->extractLength('casper1234') #1 /var/www/html/PHPSimVlidator/vendor/respect/validation/library/Rules/AllOf.php(45): Respect\Validation\Rules\Length->validate('casper1234') #2 /var/www/html/PHPSimVlidator/InputValidator.php(17): Respect\Validation\Rules\AllOf->validate('casper1234') #3 /var/www/html/PHPSimVlidator/InputValidator.php(102): InputValidator::isUsernameCorrect('casper1234') #4 {main} thrown in /var/www/html/PHPSimVlidator/vendor/respect/validation/library/Rules/Length.php on line 57

How to solve it?

auburus:
It means that you are missing the required extension `mbstring`. I think it's enabled by default on some Linux distributions, but I'm not sure. 

Please, use a version of PHP that includes this extension (or [install it](http://php.net/manual/en/mbstring.installation.php) yourself :smile: )

CasperHK:
I am using Ubuntu 16.04 with php7.0 running on the service apache2

| Key     | Value              |
|:---------|:-------------------|
| OS      | Ubuntu 16.04 |
| PHP    | php7.0           |
| Server | apache2        |

auburus:
[...]

PR #718, opened 2016-08-20
Make PHP extension mbstring required

auburus:
Issues like [issue 717](https://github.com/Respect/Validation/issues/717) can detected before, just letting composer stop the install if this extension is missing.

I haven't check if any other validator apart from `LengthValidator` use `mb_*` functions, though.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,8 @@
         }
     ],
     "require": {
-        "php": ">=5.6"
+        "php": ">=5.6",
+        "ext-mbstring": "*"
     },
     "require-dev": {
         "egulias/email-validator": "~1.2",


composer.json
```diff
@@ -12,7 +12,8 @@
         }
     ],
     "require": {
-        "php": ">=5.6"
+        "php": ">=5.6",
+        "ext-mbstring": "*"
```
henriquemoody: My only concern about it is that there are 144 rules and just because 8 rules we insert this dependency.

auburus:
In fact... thats a very valid concern, and my opinion have shifted to suggest NOT merging this pull request.
I think that the email validator also needs an externsl library, but it's not a hard dependency but a suggestion. Probably should be the same here (although I don't know if thats possible with extensions).

What I do suggest, and if someone agrees with me I'll do it, is updating the docs for those validators to make clear that they require this extension.

If thats okay I'll close this pull request.

prisis:
why you dont use symfony/polyfill-mbstring and add mbstring as suggest to composer?

henriquemoody:
@prisis, beacause it only supports UTF-8, don't know if it's a good idea...

prisis:
@henriquemoody UTF-8 is better than nothing and if mbstring is missing you getting are fatal error

composer.json
batusa: We could then check if the mb-string is enabled in these rules and throw a `RuntimeException` telling it's required. I think it's better than a Fatal Error...

What do you guys think?

prisis:
can be closed too?

henriquemoody:
Those bugs were finally fixed on versions [1.0.12](https://packagist.org/packages/respect/validation#1.0.12) and [1.1.7](https://packagist.org/packages/respect/validation#1.1.7) by #729.

Thanks for reporting! :panda_face:

Issue #620, opened 2015-12-14
setName doesn't work for attribute


strange-v:
Hi guys.
I have some troubles with setName and attributes.
My code looks like this:
```php
$validator = v::attribute('email', v::notEmpty()->email())->setName('Email Field')
	->attribute('password', v::notEmpty()->noWhitespace())->setName('Password Field');
$this->validator->assert($entity);

$errors = $ex->findMessages([
    'email' => 'Error: {{name}}',
    'notEmpty' => 'Error: {{name}}',
    'noWhitespace' => 'Error: {{name}}'
]);
```
I've expected to see error
> Error: Email Field
> Error: Password Field

But I've got
> Error: email
> Error: password

Why so? How I can set name for attribute correctly?
Thank you.

henriquemoody:
Thanks for reporting, seem it's really a bug.

Although, you're using `setName()` incorrectly, because you're defining the name of the _chain_ two times, what you really want to do it:

```php
v::attribute('email', v::notEmpty()->email()->setName('Email Field'))
    ->attribute('password', v::notEmpty()->noWhitespace()->setName('Password Field'))
    ->assert($entity);
```

I will take a closer look to this bug as soon as possible.

francoism90:
This still doesn't work:
```
$user = new \stdClass;
$user->email = 'Ale@@@xandre';
$user->password = 'Ale@@@xandre';

$username = v::attribute('email', v::notEmpty()->email()->setName('Email Field'))
    ->attribute('password', v::notEmpty()->noWhitespace()->setName('Password Field'))
    ->assert($user);`
```

Result:
`Array ( [0] => email must be valid email )`

Thanks. :)

henriquemoody:
Oh, I got it. I will fix it as soon as possible.

batusa:
Should the proper behavior be the name of the chain of rules for `attribute`, change `attribute`'s name?

I was thinking about something like `KeySet` but for attributes, it would be something like:

```php
v::attributeSet(
    v::attribute('email', v::notEmpty()->email())->setName('Email Field'),
    v::attribute('password', v::notEmpty()->noWhitespace())->setName('Password Field')
)->assert($entity);
```

I don't know if I took it too far...

What do you think?

henriquemoody:
Unfortunately the `AttributeSet` rule won't fix the problem, but it's a nice idea. Can you create a issue for that, @batusa?

henriquemoody:
[...]

PR #725, opened 2016-09-03
Added test for issue #620

auburus:
Created test case for detecting issue #620 . I won't be able to try to fix it until next weekend (although I'm not familiar with the codebase), so I'm leaving this pull request here just in case someone wants to jump in...

### DIFF
--- a/tests/integration/issue-620.phpt
+++ b/tests/integration/issue-620.phpt
@@ -1,0 +1,31 @@
+--FILE--
+<?php
+
+require 'vendor/autoload.php';
+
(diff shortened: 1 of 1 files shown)


henriquemoody:
This bug was finally fixed on versions [1.0.11](https://packagist.org/packages/respect/validation#1.0.11) and [1.1.6](https://packagist.org/packages/respect/validation#1.1.6) by #675.

I'm very sorry about my delay (I talked about that already https://github.com/Respect/Validation/issues/620#issuecomment-237474435).

Thanks for reporting! :panda_face:

PR #735, opened 2016-09-26
Add Spain's NIF

juliangut:
Spanish NIF validation in all its forms

* DNI
* NIE
* CIF


henriquemoody:
Fixed by #738.

Can you give a review before merge, @juliangut?

PR #745, opened 2016-10-19
performance improvement in Factory

buggedcom:
It is better to statically cache the reflections for performance reasons. 

Since using ReflectionClass is not very performant we statically cache the instantized reflection objects so that subsequent requests for the same rule to not have to do the expensive reflection startup process once again.

### DIFF
--- a/library/Factory.php
+++ b/library/Factory.php
@@ -47,8 +47,20 @@ class Factory
             return $ruleName;
         }
 
+        // statically cache the reflections for performance reasons.
+        // Since using RelectionClass is not very performant we statically cache
+        // the initantized reflect objects so that subsequent requests for the
+        // same rule to not have to do the expensive reflection startup process
+        // once again.
+        static $reflections = array();
+
         foreach ($this->getRulePrefixes() as $prefix) {
             $className = $prefix.ucfirst($ruleName);
+
+            if(isset($reflections[$className]) === true) {
+                return $reflections[$className]->newInstanceArgs($arguments);
+            }
+
             if (!class_exists($className)) {
                 continue;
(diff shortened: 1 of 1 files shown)


henriquemoody:
One of the problems with this change is that if we add a new rule prefix we're still going to get the old rule and I couldn't a performance improvement the would pay for having a stateful Factory:

```php
$start = microtime(true);
for ($i=0; $i < 1000000; $i++) {
    v::floatVal();
}
echo microtime(true)-$start.PHP_EOL;
```

- `master`: 2.3628811836243
- `pr/745`: 1.7847011089325

However, thanks a lot for contributing! 🐼

buggedcom:
> One of the problems with this change is that if we add a new rule prefix we're still going to get the old rule

I don't see how this is possible unless you are concerned about having the same named rule under two different namespaces - and if that is the case isn't this already an issue anyway?

henriquemoody:
It may be possible having two different rules with the same name into different namespaces. The Factory was created also with this purpose, so it's not an issue but it is an expected behaviour as long as the one does it consciously.

For exemplo, if you find a bug on some of the rules we have you can simply overwrite it while waiting for the new release.

buggedcom:
But again, my patch does not affect this functionality. In fact it operates exactly as your code currently does. The only issue I see here would be that the cache is static inside the function and perhaps there needs to be a way to clear it but given that the setup of respect validation is only really going to happen once per cycle I don't see this being an issue.

If I re-write Factory completely to take this into consideration would you consider this again? We really need this in out product since we make heavy use of respect/validation for both validation request parameters and function argument validation.

Issue #761, opened 2016-11-14
String validator and utf-8 symbols


rzmz:
Hi, awesome package, I just cannot understand, why does the string validator not work with utf-8 symbols. For an example: this does not validate as correct string:

v::alnum()->validate('müstika');

but this does: 

v::alnum()->validate('mystika');

can I somehow make the first string validate or is this something that cannot be done with this package?

rzmz:
PS: I have set the correct locale but my problem is that sometimes users with a specific locale want to input characters from another locale. Anyhow, apreciate your help.

henriquemoody:
I created #762 in order to fix this issue.

However, before I merge it I would like to understand in which scenario do you intent to use this validator?

henriquemoody:
Now is more clear what is the intention here.

According to [Wikipedia](https://en.wikipedia.org/wiki/Alphanumeric):
> Alphanumeric is a combination of alphabetic and numeric characters, and is used to describe the collection of Latin letters and Arabic digits or a text constructed from this collection.

That said, the one we have is correct, and changing it would make it wrong. What is needed is a new rule, with a new name and a new purpose. I'm planning to finish it this week, and I'm gonna make it more clear on the documentation. The rule will be probably called _Name_ but I'm very open for suggestions.

Hirbod:
Any news on this? "Name" would be perfect on this, I also reverted to "stringType" which is actually not really correct for this purpose.

nickl-:
Closed: as per #762 Won't fix

zefj:
So how are we supposed to do that?

henriquemoody:
I wrote a comment describing better what's the problem here: https://github.com/Respect/Validation/pull/817#issuecomment-424098285

I'm afraid that you need to build your custom validation, for example:

```php
v::stringType()->notEmoji()->not(v::punct())->validate()
```

But as I said, that depends a lot on what do **you** consider as a valid name.

zefj:
Resorted to regex. I don't particularly like it because of difficulties in communicating the validation error to the front-end, but then again that's the case with pretty much this entire library so I guess I'll just deal with it.

W dniu śr., 6.03.2019 o 16:50 Henrique Moody <notifications@github.com<mailto:notifications@github.com>> napisał(a):

I wrote a comment describing better what's the problem here: #817 (comment)<https://github.com/Respect/Validation/pull/817#issuecomment-424098285>

I'm afraid that you need to build your custom validation, for example:

[...]

Issue #645, opened 2016-02-24
How i can use russian language?..


musaev-haybulla:
In this library function of validate alnum, alpha and other cannot support russian language. How can fix it?..

henriquemoody:
It depends on the locale that is set. Try to define it in your bootstrap:

```php
setlocale(LC_ALL, 'ru_RUS.utf8');
```

I'm gonna wait for you feedback

musaev-haybulla:
I did as You wrote. Function ctype_alnum, like other ctype functions, not vosprinimaut changing the system locale to Cyrillic in utf-8. Neither on my server nor on a commercial server, which I configured not. Any thoughts on this?

Savsan:
Hi! I've a same problem. Russians characters can't be validate, even after i changed locale on 1251, UTF-8. How can fix it?

musaev-haybulla:
I solved this problem without making code changes. But this is not correct, because if I forget and load the composer om new version of the library, as immediately erased all what I wrote in the code. Knowledgeable guys suggested to make a fork of this library and to scratch code. Namely, to replace all the ctype functions (the benefit there is not a lot) on regular expressions.

on Russiain
Я не решил эту проблему, не внося изменений в код. А это не правильно, так как стоит мне забыться и подгрузить composer'ом новые версии библиотеки, как тут же сотрется все то, что я писал в коде. Знающие ребята подсказали сделать fork от этой библиотеки и поковырять код. А именно - заменить все ctype функции (благо их там не много), на регулярные выражения.

daffoxdev:
I have created rule anyLetter that check for regexp for any utf-8 letter. You can create same only for russian. Read how to create own rules for this library.

```php
namespace YourProjectNamespace\Validation\Rules;

use Respect\Validation\Rules\AbstractRule;

class AnyLetter extends AbstractRule
{
    public function validate($input)
    {
        if (preg_match('/^\pL+$/u', $input)) {
            return true;
        } else {
            return false;
        }
    }
}
```


and Exception
```php
namespace YourProjectNamespace\Validation\Exceptions;

use Respect\Validation\Exceptions\ValidationException;

class AnyLetterException extends ValidationException
{
    public static $defaultTemplates = [
        self::MODE_DEFAULT => [
            self::STANDARD => '{{name}} must contain only letters',
        ],
        self::MODE_NEGATIVE => [
            self::STANDARD => '{{name}} must not contain only letters',
        ],
    ];
}
```

[...]

Issue #627, opened 2016-01-05
What about chars like ÄÖÜßàéè in Alnum/Alpha?


ghost:
Hello!
I've some trouble with validating some user input using the framework. The problem is, that some languages allow chars like **Ä Ö Ü ß à è é ú** but the Alnum and Alpha validators do not allow them. Any idea? Or is there any validator available that allows all available letters?

Regards and thanks for your great work!

alganet:
Unfortunately, we don't have rules to validate such chars yet. 

A previous discussion about this started on issue #92 and it was suggested that the native `setlocale()` function might affect the behavior of the `ctype_alpha` and `ctype_alnum` functions (that we use as the basis for these rules), so calling `setlocale(LC_ALL, 'es_ES'); (changing es_ES to the appropriate language) might work. I never used it this way though.

alganet:
Reading #92 again I've remembered the Charset rule. Might be handy: http://respect.github.io/Validation/docs/charset.html

7rust:
Alright! In my opinion Respect needs at least one Validator that allows all or the most common letters (Latin + normal letters). This rule might could use the ASCII table and just don't allow any chars that aren't letters.

pultho:
What about a rule like `alphaLatin($pAdditional= "");`

Taking a look at [the following UTF basic latin table](http://www.utf8-zeichentabelle.de/unicode-utf8-table.pl?number=1024&utf8=dec) we can see that all latin chars are between:



* **65 - 90**
* **97 - 122**
* **195 - 207** (most latin special chars)

Now we just need to check whether **each char** is within this range by using `ord("Ä");`.

This wouldn't be a perfect solution but a step forward on supporting latin chars (which is definitely required in most European countries).

Maybe something like:
```
   public function validate($input)
   {
        foreach(str_split($input) as $c)
        {
            if(!(
                (in_array(ord($c), range(65, 90)))
                || (in_array(ord($c), range(97, 122)))
                || (in_array(ord($c), range(195, 207))) )
            )return false;
        }
        return true;
   }

```

henriquemoody:
I created #762 in order to fix this issue.

However, before I merge it I would like to understand in which scenario do you intent to use this validator?

pultho:
I think it's helpful for validating texts or names so that you can make sure that your input doesn't contain any special characters or symbols.

henriquemoody:
Now is more clear what is the intention here.

[...]

PR #762, opened 2016-11-15
Allow special chars on Alnum and Alpha rules

### LINKED ISSUE #92: Validate strings with latin characters like Ñ
I've used Validation mainly in English, but because of a project I'm doing I have to validate strings in spanish, meaning that the strings may include accents (áéíóúÁÉÍÓÚ), the letter Ñ (both lower and uppercase) and once in a while üÜ.

So my question is: Is there a way to validate this type of strings?

Besides using v::alpha or v::alnum and having to declare every additional chars.

``` php
v::alnum('áéíóúÁÉÍÓÚñÑ')->validate('La piñata de la nación'); 
```

henriquemoody:
Fix #92, #627, #645, #761

### DIFF
--- a/docs/Alnum.md
+++ b/docs/Alnum.md
@@ -3,10 +3,13 @@
 - `v::alnum()`
 - `v::alnum(string $additionalChars)`
 
-Validates alphanumeric characters from a-Z and 0-9.
+Validates alphanumeric characters.
 
 ```php
 v::alnum()->validate('foo 123'); // true
+v::alnum()->validate('Eso que ni qué'); // true
+v::alnum()->validate('Vai filhão'); // true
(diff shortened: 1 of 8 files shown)


DgeDge63:
Hello, this dev has never been merged, so how to resolve accent validation ?

Thanks

henriquemoody:
By definition they're both correct and this change would actually make them wrong.

After thinking a lot about it I realize that what those issues are really about are _Names_, but a _Name_ is so abstract (and hard to implement) that I couldn't come up with something really relevant. I have a work in progress on a rule called _Name_ but so  far I didn't get the proper way of doing it: https://github.com/Respect/Validation/compare/master...henriquemoody:name?expand=1

PR #764, opened 2016-11-29
Czech and slovakia postal code patterns fix.

mablo:
According to wikipedia zip code Czech Republic and Slovakia can also take a value in the format XXX XX.

Links to wikipedia:
- https://en.wikipedia.org/wiki/List_of_postal_codes_in_the_Czech_Republic
- https://en.wikipedia.org/wiki/Postal_codes_in_Slovakia


henriquemoody:
Merged manually by fa61315079c7634d52109d4cd6ef67609f7f674c

Thanks for contributing! 🐼

PR #768, opened 2016-12-05
Update MasterCard BIN Range

andysnell:
MasterCard has announced that they are increasing the number of potentially valid credit card numbers by adding new BINs (bank identification numbers, i.e., the first six digits of the credit card number). Effective January 2017, banks may begin to issue MasterCard credit cards starting with "2" instead of the usual "5". This major change was announced last year, and MasterCard is expecting acceptance compliance by June 2017.

This PR updates the valid MasterCard card number ranges, modifying the regular expression to support 16-digit card numbers starting with 22-27. Additionally, it adds a published test card number to the CreditCard unit test. No updates to the Exception are necessary.

Additional information on the BIN range update can be found at:
https://www.mastercard.us/en-us/issuers/get-support/2-series-bin-expansion.html


henriquemoody:
Thanks for contributing! 🐼

PR #777, opened 2017-01-18
Updated provinces of Cuba

yosmanyga:
Updated provinces of Cuba according to the change of 2011


henriquemoody:
Merged manually by 32414451b0dd6dd96fb5558819c5e61b8a7a38f2

Thanks for contributing! 🐼

PR #795, opened 2017-02-28
update phpdoc annotations for variadic methods

jstewmon:
Accurate signature annotations on the Validator class improve the editing experience in editors that make use of phpdoc annotations. Without the variadic signature, editors like PHPStorm will complain that the usage does not match the signature.


henriquemoody:
Thanks for contributing! 🐼

PR #798, opened 2017-03-02
do not skip `AttributeException` during recursive iteration

### LINKED ISSUE #796: NestedValidationException::getFullMessage is missing messages when one nested validator fails
If only one nested validator fails, the list of messages does not contain the message indicating which parent validator failed.

Here's an updated version of `assert-with-keys.phpt` that demonstrates the issue:

```php
--FILE--
<?php

try {
    v::create()
        ->key(
            'mysql',
            v::create()
                ->key('host', v::stringType(), true)
                ->key('user', v::stringType(), true)
                ->key('password', v::stringType(), true)
                ->key('schema', v::stringType(), true),
            true
        )
        ->key(
            'postgresql',
            v::create()
                ->key('host', v::stringType(), true)
                ->key('user', v::stringType(), true)
                ->key('password', v::stringType(), true)
                ->key('schema', v::stringType(), true),
            true
        )
        ->setName('the given data')
        ->assert([
            'mysql' => [
                'host' => 42,
                'user' => 'user',
                'password' => 'password',
                'schema' => 'schema',
            ],
            'postgresql' => [
                'host' => 'host',
                'user' => 42,
                'password' => 'password',
                'schema' => 'schema',
            ],
        ]);
} catch (NestedValidationException $exception) {
}
?>
--EXPECTF--
- All of the required rules must pass for the given data
  - All of the required rules must pass for mysql
    - host must be a string
  - All of the required rules must pass for postgresql
    - user must be a string
```

```diff
Failed asserting that format description matches text.
--- Expected
+++ Actual
@@ @@
 - All of the required rules must pass for the given data
-  - All of the required rules must pass for mysql
-    - host must be a string
-  - All of the required rules must pass for postgresql
-    - user must be a string
+  - host must be a string
+  - user must be a string
```

I think the problem comes from this block in NestedValidationException::getIterator:

```php
            if ($childException instanceof self
                && $childException->getRelated()->count() > 0
                && $childException->getRelated()->count() < 2) {
                continue;
            }
```

Removing that block results in more verbose message nesting being reported, which breaks some current tests. I haven't been able to figure out what the right solution is to preserve the context when there is a single failure.

I'm willing to work up a PR to fix the issue, but I haven't had luck fixing it such that the hoisting expected by existing tests do not break, so if you have any pointers, I could use a little help.

jstewmon:
`AttributeException` instances always indicate an interesting nested
context, which should not be omitted from the final result.

Fixes #796

### DIFF
--- a/library/Exceptions/NestedValidationException.php
+++ b/library/Exceptions/NestedValidationException.php
@@ -137,11 +137,12 @@ class NestedValidationException extends ValidationException implements IteratorA
         $knownDepths = [];
         foreach ($recursiveIteratorIterator as $childException) {
             if ($childException instanceof self
-                && $childException->getRelated()->count() > 0
-                && $childException->getRelated()->count() < 2) {
-                continue;
+                && $childException->getRelated()->count() === 1) {
+                $relatedException = iterator_to_array($childException->getRelated())[0];
+                if (! $relatedException instanceof  AbstractRelatedExceptionInterface) {
+                    continue;
+                }
             }
-
             $currentDepth = $lastDepth;
             $currentDepthOriginal = $recursiveIteratorIterator->getDepth() + 1;
 
--- a/library/Exceptions/AbstractRelatedExceptionInterface.php
+++ b/library/Exceptions/AbstractRelatedExceptionInterface.php
@@ -11,6 +11,7 @@
 
 namespace Respect\Validation\Exceptions;
 
-class CallException extends GroupedValidationException implements AbstractRelatedExceptionInterface
+interface AbstractRelatedExceptionInterface extends ExceptionInterface
 {
+
 }
--- a/library/Exceptions/AttributeException.php
+++ b/library/Exceptions/AttributeException.php
@@ -11,7 +11,7 @@
 
 namespace Respect\Validation\Exceptions;
 
-class AttributeException extends NestedValidationException
+class AttributeException extends NestedValidationException implements AbstractRelatedExceptionInterface
 {
     const NOT_PRESENT = 0;
     const INVALID = 1;
--- a/tests/integration/assert-with-attributes.phpt
+++ b/tests/integration/assert-with-attributes.phpt
@@ -1,0 +1,53 @@
+--FILE--
+<?php
+require 'vendor/autoload.php';
+use Respect\Validation\Exceptions\NestedValidationException;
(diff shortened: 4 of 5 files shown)


library/Exceptions/NestedValidationException.php
```diff
@@ -137,11 +137,12 @@ public function getIterator()
         $knownDepths = [];
         foreach ($recursiveIteratorIterator as $childException) {
             if ($childException instanceof self
-                && $childException->getRelated()->count() > 0
-                && $childException->getRelated()->count() < 2) {
-                continue;
+                && $childException->getRelated()->count() === 1) {
+                $relatedException = iterator_to_array($childException->getRelated())[0];
+                if (! $relatedException instanceof  AttributeException) {
```
henriquemoody: Can you check `AbstractRelated` instead of `AttributeException`?

jstewmon: Hi, thanks for reviewing!

`AbstractRelated` is a `Validatable` type, not an `Exception`, so they cannot be interchanged here. There isn't a `AbstractRelatedException` - `AttributeException` is derived from `NestedValidationException`.

I think the closest equivalent to checking for `AbstractRelate` is by checking if the related exception has a `validator` param, but that doesn't seem like a very reliable / accurate check.

We could add an `AbstractRelatedException` class and make all `AbstractRelated` rules' exceptions inherit from it, but we'd run into another problem:
- These types inherit `AbstractRelated`: `Attribute`, `Call`, `Key`, `KeyNested`
- `Key` and `KeyNested` inherit `AttributeException` which inherits `NestedValidationException`
- But, `CallException` inherits `GroupedValidationException`

So, I think we'd want to make all of those rule / exception types to have congruent type ancestry.

What do you think?

jstewmon: @henriquemoody any thoughts on my last comment?

henriquemoody: Sorry about my late response. The alternative would create an interface, implement it on those exception classes, and checking for that interface.

jstewmon: Thanks for the suggestion! I think that's a reasonable approach. I've updated the PR accordingly.

henriquemoody:
Thanks for contributing! 🐼

PR #803, opened 2017-03-15
Added support for Symfony Validator >= 3.0

blackness90:
Version 1.1 is not compatible with symfony/validator >=3.0, it's using a deprecated method `validateValue()`, which was renamed to `validate()`.

I guess it should also work also with the symfony/validator ~2.6


henriquemoody:
Symfony Validator >= 3.0 is already supported on master and cannot be supported on version 1.1, since it breaks the API.

However, thanks for contributing! 🐼

PR #812, opened 2017-04-03
Create "Unique" rule

gronostajo:


docs/Unique.md
```diff
@@ -0,0 +1,11 @@
...
+
+- `Unique()`
+
+Validates whether the input array contains only unique values.
+
+```php
+v::unique()->validate([]); // true
+v::unique()->validate([1, 2, 3]); // true
+v::unique()->validate([1, 2, 2, 3]); // false
+```
```
henriquemoody: Could you also add some related validator here?

library/Exceptions/UniqueException.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Respect\Validation\Exceptions;
+
+class UniqueException extends ValidationException
+{
+    public static $defaultTemplates = [
+        self::MODE_DEFAULT => [
+            self::STANDARD => '{{name}} must not contain duplicates',
+        ],
+        self::MODE_NEGATIVE => [
+            self::STANDARD => '{{name}} must contain at least one duplicate',
```
henriquemoody: I think this is too specific, how about _{{name}} must contain duplicates_?

***
Not a change request here

library/Rules/Unique.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Respect\Validation\Rules;
+
+class Unique extends AbstractRule
+{
+    public function validate($input)
+    {
+        if (!is_array($input)) {
+            return false;
+        }
+        return count($input) == count(array_unique($input));
```
henriquemoody: `$input == array_unique($input)` is enough. Also, please, add a new line before `return`. 

***
As we [suggest](https://github.com/Respect/Validation/blob/master/CONTRIBUTING.md#coding-style-and-standards), run PHP Coding Standards Fixer to be sure.

henriquemoody:
https://git-scm.com/book/en/v2/Git-Branching-Rebasing 

If you need any help, let me know 😄

tests/unit/Rules/UniqueTest.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+class UniqueTest extends RuleTestCase
+{
+    public function providerForValidInput()
+    {
+        $rule = new Unique();
+
+        return [
+            [$rule, []],
+            [$rule, [1, 2, 3]],
```
henriquemoody: Would be nice to create tests with different kinds of data (arrays, booleans, objects, floats, strings)

gronostajo:
Oh, I wasn't sure how GitHub's PRs work with rebases. Will fix that in a moment :)

By the way, php-cs-fixer doesn't work for me, it's not in `composer.json` too. I tried to install it, but it still didn't work afterwards: `[Symfony\Component\Debug\Exception\FatalThrowableError] Class 'Symfony\CS\Finder\DefaultFinder' not found`. I've also tried to install it globally, it throws the same error but more verbose.

PR #815, opened 2017-04-11
Fix Nif Validator

senenh:
Output of str_split is an array of strings. It's necessary to cast the output to "int" to be able to compare $lastDigit to 0 with an identical operator.


henriquemoody:
Nice catch! Thanks for contributing! 🐼

PR #817, opened 2017-04-13
Create "Name" rule

henriquemoody:


henriquemoody:
I think this is a lost battle. Names don't standardize. An "John" can be a valid name as "🙃". Many companies standardize the names they save in their database, but that's _their_ convention, it's impossible to get a convention that will meet everyone's expectations.

Issue #345, opened 2015-04-29
domain containing -- should be allowed


JayQ1982:
v::domain()->validate($domain) returns false (instead of true), if $domain = 'std--a.com'.

Is there a reason for new Not(new Contains('--')) in Rules\Domain.php or can I just comment this part out?

augustohp:
It seems you can comment that out, but I am really not familiar with all the domain naming rules which needs to be revisited to ensure we are not breaking any formal rules.
The '--' part is related to unicode domains using punnycode to convert unicode characters to ascii, at least that I am aware of. But I am very inclined to believe they are valid for regular domains also.

I really don't known if domains with two dashes are valid, I will take a look at the following RFCs: [1034](http://www.ietf.org/rfc/rfc1034.txt), [1123](http://www.ietf.org/rfc/rfc1123.txt), [2396](https://www.ietf.org/rfc/rfc2396.txt) and [3492](https://www.ietf.org/rfc/rfc2396.txt).

Any help you can provide, by checking if these kind of domains is really valid, would be greatly appreciated.

PS: Thanks on pointing that out btw. :wink:

alganet:
I didn't knew domains with two hyphens were possible. I thought they were reserved for [punycodes](http://en.wikipedia.org/wiki/Punycode), which we partially support.

I was careless implementing this part of [RFC3696](https://tools.ietf.org/html/rfc3696):

>   One quick local test however, may be reasonable: as of the time of
>   this writing, there should be no instances of labels in the DNS that
>   start with two characters, followed by two hyphens, where the two
>   characters are not "xn" (in, of course, either upper or lower case).
>   Such label strings, if they appear, are probably erroneous or
>   obsolete, and it may be reasonable to at least warn the user about
>   them.

It appears that `std--a` has three characters before the two hyphens, so our domain rule is indeed invalid. We should allow two hyphens if there are more than two letters between them and the start of the domain.

augustohp:
Great, will work on that today. Btw, we also do not support hostnames when we are not checking TLDs, which is a wrong behavior also: `localhost` should be a valid domain when TLD check is disabled.

JayQ1982:
so for the moment it worked for me by just changing Contains('--') to new StartsWith('xn--') in Rules/Domain.php

r--w:
+1 for this issue, my domain r--w.com is not validated correctly;)

henriquemoody:
[...]

PR #825, opened 2017-04-28
Fix Issue #345 Changed domain validation condition 

clauculo:
Changed condition and added test to validate the new condition.

Bug:
https://github.com/Respect/Validation/issues/345


library/Rules/Domain.php
```diff
@@ -27,18 +27,7 @@ public function __construct($tldCheck = true)
...
-                    new Contains('--')
-                ),
-                new AllOf(
-                    new StartsWith('xn--'),
-                    new Callback(function ($str) {
-                        return mb_substr_count($str, '--') == 1;
-                    })
-                )
-            )
+            new Not(new StartsWith('-'))
```
henriquemoody: You should also add `new Not(new EndsWith('-'))`, because `domain-.com` is also not valid

clauculo:
@henriquemoody Ok! I've included also that piece (and added the unit-test)

henriquemoody:
Fixed on [1.0.20](https://github.com/Respect/Validation/releases/tag/1.0.20) and [1.1.15](https://github.com/Respect/Validation/releases/tag/1.1.15).

PR #838, opened 2017-07-11
Add base64 rule

jenssegers:
This PR adds a base64 rule to check if a value is base64 encoded.

### DIFF
--- a/docs/Base64.md
+++ b/docs/Base64.md
@@ -1,0 +1,21 @@
+# Base
+
+- `Base64()`
+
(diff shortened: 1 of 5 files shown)


Singwai:
https://en.wikipedia.org/wiki/Base64

jenssegers:
Any updates?

Singwai:
I think there are a few improvements can be done in the regex. I will work on it this weekend.

henriquemoody:
Merged by #865

Thanks for contributing! 🐼

PR #839, opened 2017-07-13
CurrencyCode Validator - BYR should be BYN

tstrijdhorst:
References:
https://en.wikipedia.org/wiki/ISO_4217
www.xe.com/iso4217.php


henriquemoody:
Thanks for contributing! 🐼 

This fix will be available on versions 1.0.18 and 1.1.13.

PR #841, opened 2017-07-19
Fix regex

RCooLeR:
Fix regex to match +1 (555) 555 5555


henriquemoody:
Merged by #870.

It is available on versions 1.0.19 and 1.1.14.

Thanks for contributing! 🐼

Issue #805, opened 2017-03-21
How can I set custom error message inside the KEY validation ?


BaNdErOzZz:
`try{`
`    Validator::key('email', Validator::email()->setTemplate('WRONG EMAIL !!!!!!'))->assert(['email' => 'qwe']);`
 `}catch (NestedValidationException $e){`
`            var_dump($e->getMessages()); die();`
`        }`

Shows: "_email must be valid email_" instead of "_WRONG EMAIL !!!!!!_" .

But if I use  `Validator::email()->setTemplate('WRONG EMAIL !!!!!!')->assert('qwe')` it's working !

I use 1.1.12 version.

BaNdErOzZz:
Any suggestions ?

fracz:
Seems like a dup of #619

henriquemoody:
It should be fixed in the version 1.1.19 which I intent to release very soon.

Thank you very much for reporting! 🐼

PR #854, opened 2017-09-14
Set the custom message template in all children rules

### LINKED ISSUE #619: Problems with setTemplate
This code does not change exception message

``` php
    Validator::instance(\stdClass::class)->setTemplate('invalid object')->check('test');
```

This works fine

``` php
(new Instance(\stdClass::class))->setTemplate('invalid object')->check('test');
```

How to set custom template when I use fluent interface of Validator object?

fracz:
This fixes #619 #805

### DIFF
--- a/library/Rules/AbstractComposite.php
+++ b/library/Rules/AbstractComposite.php
@@ -39,6 +39,21 @@ abstract class AbstractComposite extends AbstractRule
         return parent::setName($name);
     }
 
+    public function setTemplate($template)
+    {
+        $parentTemplate = $this->getTemplate();
+        foreach ($this->rules as $rule) {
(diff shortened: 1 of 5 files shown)


fracz:
One possibly breaking change - the templates now should be set ASAP, because setting them after defining all validators will overwrite the default messages.

See the change in unit test [here](https://github.com/Respect/Validation/pull/854/commits/b83f55e3e97d7103b6eb2eea69e7c4bf66750a0b#diff-a58342dd33b6c4abefdfbc2c0e131f0d) to know what I'm talking about.

henriquemoody:
I sent #1047 with a few changes. I'm sorry by my delay, it took some time to think about something that would fix everything.

Thanks for contributing! 🐼

PR #678, opened 2016-05-08
New Concrete API

henriquemoody:
## Motivation

- We don't have enough granularity control. Therefore rules like `not()` cannot work properly;
- `Validatable` interface has too many methods;
- The current API makes it hard to handle messages and translations;
- All rules depend on `AbstractRule`;.

***
Related to #354

PR #857, opened 2017-09-30
PHPDoc, parameter types, and return types for main classes

thadbryson:
I added the following to:

interface: Respect\Validation\Validatable
classes: 
- Respect\Validation\Validator
- Respect\Validation\Factory

* @PHPDoc for parameters, Exceptions thrown, and return types.
* Parameter type hinting on methods where possible (PHP 7.0).
* Return type hinting on all methods where possible (PHP 7.0). 

I did not type hint the interface because I'd have to type hint all the Rules too. I don't have time for that right now.

### DIFF
--- a/library/Factory.php
+++ b/library/Factory.php
@@ -16,14 +16,23 @@ use Respect\Validation\Exceptions\ComponentException;
 
 class Factory
 {
+    /** @var string[] */
     protected $rulePrefixes = ['Respect\\Validation\\Rules\\'];
 
-    public function getRulePrefixes()
+    /**
+     * @return string[]
+     */
+    public function getRulePrefixes(): array
     {
         return $this->rulePrefixes;
     }
 
-    private function filterRulePrefix($rulePrefix)
+    /**
+     * @param string $rulePrefix
+     *
+     * @return string
+     */
+    private function filterRulePrefix(string $rulePrefix): string
     {
         $namespaceSeparator = '\\';
         $rulePrefix = rtrim($rulePrefix, $namespaceSeparator);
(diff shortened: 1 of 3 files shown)


nickl-:
Wow nice work! I didn't even know PHP got type aware lol...

But it's true what you say, we'll have to do all of it and just committing a piece job will only create inconsistencies. Besides @henriquemoody has already been working on a large rewrite at #678 and this would probably fit better pushed against that branch... we'll have to hear what he has to say.

Perhaps you can rally up some support and maybe share this branch with some friends or see if you get some more free time to fill out the rest. It is best to stick to a single goal though, do the type hinting with doc comments and avoiding getting off topic by modifying design patterns. These will just invite more scrutiny, debate and not to forget... testing! Don't let distractions get in the way of your original intentions and lets just get this done. What do you say?

thadbryson:
I'm glad you liked my work. It's nice to be appreciated. I really like the project and want to help. 

I'm fine with merging into whatever branch. Whatever the plan of the project is. 

I think the code should use type hinting wherever possible. It's a feature of PHP 7.0. Really the PHPDoc should be a requirement of the project one way or the other. Use type hinting and PHPDoc or not. 

Although we could have the argument not to use PHPDoc unless necessary. IDEs will pick up the type hinting if it's set which makes PHPDoc unnecessary.

henriquemoody:
Thanks for contributing, @thadbryson! 

I don't think it's worth to update those docblocks once there is this big refactor on #678. 

However, I would really appreciate if you could help me on that PR as well!

🐼

thadbryson:
Thanks. I'll look at that PR.

PR #871, opened 2017-10-17
Changed domain validation condition

### LINKED ISSUE #345: domain containing -- should be allowed

henriquemoody:
Close #825
Fix #345


henriquemoody:
Hey, @williamespindola! Just saw your message! I'm preparing to merge it, apparently there is problem at all. Domains can have `--` as long as they don't start or end with `-` in its parts.

PR #874, opened 2017-10-28
Add Docker configs

brunoluiz:
Using `Docker` and `docker-compose` makes easier for any developer to implement new features, as it will be a controlled environment. The only issue is that the tests are not passing inside the composer container (which seems strange).


henriquemoody:
Thanks for contributing, @brunoluiz! 🐼 

I love Docker (and Docker-Compose) however, I think I have to decline this PR, because the client's environment should not be our responsibility. We event used to have a `Makefile` which was useful, but was removed because how you use the library should not be responsibility.

Issue #826, opened 2017-05-01
Create "Date" rule


henriquemoody:
Request for a new feature to validate date supporting the following formats:

Format  | Description                                                           | Values
--------|-----------------------------------------------------------------------|-------------------------
`d`     | Day of the month, 2 digits with leading zeros                         | 01 to 31
`j`     | Day of the month without leading zeros                                | 1 to 31
`S`     | English ordinal suffix for the day of the month, 2 characters         | st, nd, rd or th
`F`     | A full textual representation of a month, such as January or March    | January to December
`m`     | Numeric representation of a month, with leading zeros                 | 01 to 12
`M`     | A short textual representation of a month, three letters              | Jan to Dec
`n`     | Numeric representation of a month, without leading zeros              | 1 to 12
`Y`     | A full numeric representation of a year, 4 digits                     | Examples: 1988 or 2017
`y`     | A two digit representation of a year                                  | Examples: 88 or 17

The rule should accept only those formats plus spaces and any non-alphanumeric character, so we can do something like: `F jS, Y` (e.g. May 1st, 2017).
***
Depends on #821

DaniloBenevides:
Hello, still need help on this issue ?

henriquemoody:
@brunoluiz was working on it, but it would be fine by me if you would like to take over. See #875. Or you can also help on #810

DaniloBenevides:
I will start with #810 if when i finish it #875 still undone i will help there too.

henriquemoody:
Fixed by #911. Do you have any update about the `Time` rule, @DaniloBenevides?

henriquemoody:
Hey, @DaniloBenevides! Don't worry about it, I took care of it, see #911 and #912.

However, if you still want to contribute, you can check some "Easy pick" issues: https://github.com/Respect/Validation/labels/Easy%20Pick

🐼

DaniloBenevides:
Thanks @henriquemoody , I will search one.

PR #875, opened 2017-10-28
Create "Date" rule

brunoluiz:
Creates rule for #826

### DIFF
--- a/library/Rules/Date.php
+++ b/library/Rules/Date.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ * This file is part of Respect/Validation.
+ *
+ * (c) Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
+ *
+ * For the full copyright and license information, please view the "LICENSE.md"
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Rules;
+
+use DateTimeInterface;
+
+class Date extends AbstractRule
+{
+    public $format = null;
+
+    public function __construct($format = null)
+    {
+        $this->format = $format;
+    }
+
+    public function validate($input)
+    {
+        var_dump($input);
+        if ($input instanceof DateTimeInterface) {
+            return true;
+        }
+
+        if (!is_scalar($input)) {
+            return false;
+        }
+
+        $allowedFormat = '/[djSFmMnYy\W]+$/';
+        if(!preg_match($allowedFormat, $this->format)) {
+            return false;
+        }
+
+        $inputString = (string) $input;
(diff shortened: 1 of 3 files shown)


library/Rules/Date.php
```diff
@@ -0,0 +1,49 @@
...
+
+use DateTimeInterface;
+
+class Date extends AbstractRule
+{
+    public $format = null;
+
+    public function __construct($format = null)
+    {
+        $this->format = $format;
```
henriquemoody: This property should not be even set if `$format` doesn't match `/[djSFmMnYy\W]+$/`

library/Rules/Date.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public function validate($input)
+    {
+        if ($input instanceof DateTimeInterface) {
```
henriquemoody: `DateTimeInterface` is not only a _date_ therefore it should be false

library/Rules/Date.php
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        $inputString = (string) $input;
+        if (is_null($this->format)) {
+            return false !== strtotime($inputString);
+        }
+
+        $info = date_parse_from_format($this->format, $inputString);
+
+        return $info['error_count'] === 0 && $info['warning_count'] === 0;
```
henriquemoody: In order to avoid duplication, I would rather:

```php
$rule = new DateTime($this->format);

return $rule->validate($input);
```

tests/unit/Rules/DateTest.php
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * For the full copyright and license information, please view the "LICENSE.md"
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Rules;
+
+use DateTimeImmutable;
+
+    /**
```
henriquemoody: Remove this extra space

tests/unit/Rules/DateTest.php
```diff
@@ -0,0 +1,162 @@
...
+ * @group  rule
+ * @covers \Respect\Validation\Rules\Date
+ * @covers \Respect\Validation\Exceptions\DateException
+ */
+class DateTest extends \PHPUnit_Framework_TestCase
```
henriquemoody: It's better to extend `RuleTestCase`

brunoluiz:
I guess I missed the docs as well

henriquemoody:
Thanks for contributing! 🐼 

I made some changes and merged on #911!

PR #879, opened 2017-11-04
Updated to PHPUnit 6

gabriel-caruso:
For a feature release of this package, I've updated the suitcase of tests to version 6 of `PHPUnit`.

The biggest change was namespaces. Previous, the tests extended `PHPUnit_Framework_TestCase`, and now `PHPUnit\Framework\TestCase`.

`setExpectedException` method was also changed to `expectException`.

I used [this article](https://thephp.cc/news/2017/02/migrating-to-phpunit-6) from @sebastianbergmann to help me in this update.

PS: There's only [one Travis CI test](https://travis-ci.org/Respect/Validation/jobs/297167988) that is not passing, probably because of this section in the article:

> Global and super-global variables are no longer backed up before and restored after each test by default

What should we do to fix it? :cry:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     },
     "require-dev": {
         "egulias/email-validator": "^2.0",
-        "malkusch/bav": "^1.1",
+        "malkusch/bav": "^1.2",
         "malukenho/docheader": "^0.1.4",
         "mikey179/vfsStream": "^1.6",
         "phpunit/phpunit": "^6.4",
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "malkusch/bav": "^1.1",
         "malukenho/docheader": "^0.1.4",
         "mikey179/vfsStream": "^1.6",
-        "phpunit/phpunit": "^5.6",
+        "phpunit/phpunit": "^6.4",
         "symfony/validator": "^3.0",
         "zendframework/zend-validator": "^2.0"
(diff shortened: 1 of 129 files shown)


williamespindola:
@gabriel-caruso in the same article he said how to enable globals take a look:
> Simply use the --globals-backup commandline option or set backupGlobals="true" in your XML configuration file to enable the backup of global and super-global variables before each test and their restoring after each test.

Did you try something like that?

gabriel-caruso:
@williamespindola Yep, I've tried to add `backupGlobals="true"` in XML configuration file, but nothing changed. I have a guess:
The test that is failing is using `--prefer-lowest --prefer-stable` options, and reading the [PHPUnit 6 Changelog](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md), some bugs related to global variables was only solved in early versions. The prove is that all our other tests are passing. What do you think? :grimacing:

henriquemoody:
This is failing because of a dependency that was removed. #883 will fix it

gabriel-caruso:
@henriquemoody Oh, okay. As soon as #883 is merged, I rebase this PR, run the tests again, and we should be ready to go.

henriquemoody:
I merged it on #884, also adding fa030637cc38c252362e343289465e9e2bd1038d

Thanks for contributing! 🐼

PR #882, opened 2017-11-09
Fix NL postal code

pontoffeltier:
Postal code for NL allows for a space between the digits and the letters


library/Rules/PostalCode.php
```diff
@@ -122,7 +122,7 @@ class PostalCode extends Regex
         'NF' => "/^(\d{4})$/",
         'NG' => "/^(\d{6})$/",
         'NI' => "/^(\d{7})$/",
-        'NL' => "/^(\d{4}[A-Z]{2})$/",
+        'NL' => "/^(\d{4}\s?[A-Z]{2})$/",
```
henriquemoody: `\s` stands for `\t` and `\n` as  well. I believe only a single space is what you want

PR #891, opened 2017-12-01
Test against PHP 7.2

carusogabriel:
With the release of [`PHP 7.2`](http://php.net/archive/2017.php#id2017-11-30-1), would be interesting to test it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ php:
   - 5.6
   - 7.0
   - 7.1
+  - 7.2
   - hhvm
 
 before_script:


henriquemoody:
In that case you also need to upgrade PHPUnit to version 5

carusogabriel:
I’ll add support to it in another PR 👨🏻‍💻

carusogabriel:
Actually, as we already test PHP 7.2 on master, should we do it to 1.1?

henriquemoody:
That would be good, but we need to upgrade PHPUnit fist since PHPUnit 4 fails on PHP 7.2.

carusogabriel:
Upgrade or add support? As PHPUnit 5 requires PHP 5.6 :sweat_smile:

henriquemoody:
I'm sorry that I didn't reply, sometimes I get lost with the GitHub notifications. 

I don't think we should add support to a new version of PHPUnit in version that was already released. `master` has support for PHP 7.2, though, tests are running fine. :)

PR #892, opened 2017-12-02
Create "SuperSet" rule

Singwai:


Singwai:
👀  will fix the issue in a moment

library/Rules/SuperSet.php
```diff
@@ -0,0 +1,32 @@
...
+class SuperSet extends AbstractRule
+{
+
+    public $compareTo;
+
+    /**
+     * SuperSet constructor.
+     * @param $compareTo
+     */
+    public function __construct($compareTo)
```
henriquemoody: `array $compareTo`

library/Rules/SuperSet.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        $this->compareTo = $compareTo;
+    }
+
+    public function validate($input)
+    {
+        return !array_diff($input, $this->compareTo);
```
henriquemoody:
```php
if (!is_array($input)) {
    return false;
}
```

tests/unit/Rules/SuperSetTest.php
```diff
@@ -0,0 +1,61 @@
...
+namespace Respect\Validation\Rules;
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @group  rule
+ * @covers \Respect\Validation\Rules\SuperSet
+ * @covers \Respect\Validation\Exceptions\SuperSetException
+ */
+class SuperSetTest extends TestCase
```
henriquemoody: Could you extend [RuleTestCase](https://github.com/Respect/Validation/blob/1.1/tests/library/Rules/RuleTestCase.php) instead? Check this out: https://github.com/Respect/Validation/blob/1.1/CONTRIBUTING.md#creating-unit-tests

library/Rules/SuperSet.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of Respect/Validation.
+ *
+ * (c) Singwai Chan<c.singwai@gmail.com>
```
henriquemoody: This header should be the same for all files.

library/Rules/SuperSet.php
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * For the full copyright and license information, please view the "LICENSE.md"
+ * file that was distributed with this source code.
+ */
+
+namespace Respect\Validation\Rules;
+
+class SuperSet extends AbstractRule
```
henriquemoody: You can add here a:
```php
/**
 * @author Singwai Chan <c.singwai@gmail.com>
 */
```

henriquemoody:
I sent #1045 with the changes that I request. Also while refactoring that, I noticed that the name of the rule didn't indicate what it was validating. So I renamed it to **Subset** instead.

Thanks for contributing! 🐼

PR #894, opened 2018-01-02
Add "respect/stringifier" as dependency

henriquemoody:


henriquemoody:
Added to #678 instead

PR #902, opened 2018-01-14
Apply contribution guidelines to "Equals" rule

henriquemoody:


henriquemoody:
I just realised that the RuleTestCase introduces a problem: because the SUT is in the data provider and it is loaded way before the test itself, the constructor of the rule has no code coverage. I'm wondering weather using RuleTestCase is a good idea or not.

tests/unit/Rules/EqualsTest.php
```diff
@@ -13,61 +13,53 @@
...
+        ];
     }
 
     /**
-     * @expectedException \Respect\Validation\Exceptions\EqualsException
-     * @expectedExceptionMessage "24" must equal 42
+     * @test
      */
-    public function testShouldThrowTheProperExceptionWhenFailure(): void
+    public function shouldPassCompareToParameterToException(): void
```
henriquemoody: When the rule passes parameters to exceptions a test like that should be created.

PR #904, opened 2018-01-18
Add nullable rule

jenssegers:
Currently, the `optional` check allows both `null` and `""`. This validation rule behaves similar to the `optional` one, but only allows `null`. This is especially helpful when validating json input.


mrnnn:
So sexy <3

library/Rules/Nullable.php
```diff
@@ -0,0 +1,49 @@
...
+        if ($input === null) {
+            return true;
+        }
+
+        return parent::check($input);
+    }
+
+    public function validate($input)
+    {
+        if ($this->isOptional($input)) {
```
WouterSioen: I guess this should also be `$input === null` instead of the `isOptional` method call?

henriquemoody:
Thanks for contributing, @jenssegers!

These value is what I recently called `undefined`, I recommend you to check #905 and use the `UndefinedHelper`.

library/Exceptions/NullableException.php
```diff
@@ -0,0 +1,34 @@
...
+        ],
+        self::MODE_NEGATIVE => [
+            self::STANDARD => 'The value must not be null',
+            self::NAMED => '{{name}} must not be null',
+        ],
+    ];
+
+    public function chooseTemplate()
+    {
+        return $this->getName() == '' ? static::STANDARD : static::NAMED;
```
henriquemoody:
```diff
-- return $this->getName() == '' ? static::STANDARD : static::NAMED;
++ return $this->hasName() ? static::NAMED : static::STANDARD;
```

henriquemoody:
Could you rebase it to `master` and take a look at the [CONTRIBUTING.md](https://github.com/Respect/Validation/blob/master/CONTRIBUTING.md)?

jenssegers:
I applied the suggested changes and added tests.

WouterSioen:
👏

henriquemoody:
I made some changes and I created the PR #909, although I'm not really sure about the name.

How about something like `v::allowNull(v::stringType())->validate($input)`?

henriquemoody:
Closed by #909

Thanks for contributing! 🐼

PR #923, opened 2018-02-26
Test enhancement

peter279k:
# Changed log

- add more tests.


henriquemoody:
Thank you very much for your PR, but the tests you wrote are a bit out of how we want to write our tests. 🐼 

If you still want to help (and I hope you do), take a look at the [Easy Pick](https://github.com/Respect/Validation/issues?q=is%3Aissue+is%3Aopen+label%3A%22Easy+Pick%22) issues we have!

henriquemoody:
I was not disconcerting important of your tests, I closed because I thought it would make more sense to close this one and that you would send new ones in the standard we have. If you would to keep contributing (and as I said, I hope you do!), here how it goes:

* Rules are tested using RuleTestCase (see [ArrayValTest](https://github.com/Respect/Validation/blob/master/tests/unit/Rules/ArrayValTest.php) or [EqualsTest](https://github.com/Respect/Validation/blob/master/tests/unit/Rules/EqualsTest.php))
* Rule exceptions are tested in integration tests using PHPT (See [arrayVal.phpt](https://github.com/Respect/Validation/blob/master/tests/integration/rules/arrayVal.phpt))

Since we're talking, I also would like to understand what was your criteria to create tests, why those classes, for example?

peter279k:
@henriquemoody, thank you for your reply.
I knew that and I should follow your test rules.
And I think that it's the big work for this change into yours.
In that moment, I will take the previous comment that you write to take a look at the issue with "Easy pick" tag :+1: .

henriquemoody:
Cool, @peter279k! Let me know whether I should close this PR or not 🐼

PR #939, opened 2018-03-26
Apply contribution guidelines to "IntType" rule

henriquemoody:


library/Exceptions/IntTypeException.php
```diff
@@ -13,8 +13,16 @@
...
+ * @author Henrique Moody <henriquemoody@gmail.com>
+ */
+final class IntTypeException extends ValidationException
 {
+    /**
+     * {@inheritdoc}
+     */
     public static $defaultTemplates = [
         self::MODE_DEFAULT => [
             self::STANDARD => '{{name}} must be of the type integer',
```
batusa: `{{name}} must be of type integer`

batusa: the same for the negative mode.

PR #940, opened 2018-03-26
Apply contribution guidelines to "NullType" rule

henriquemoody:


tests/unit/Rules/NullTypeTest.php
```diff
@@ -13,47 +13,47 @@
...
-            [0],
-            ['w poiur'],
-            [' '],
-            ['Foo'],
+            [$rule, ''],
+            [$rule, false],
+            [$rule, []],
+            [$rule, 0],
+            [$rule, 'w poiur'],
+            [$rule, ' '],
```
batusa: these last two are not necessary.

henriquemoody: I just kept the same as we had before. Do you think we should remove it?

batusa: not a big deal but jus to keep it cleaner. so I think that yes.

PR #942, opened 2018-03-26
Apply contribution guidelines to "ResourceType" rule

henriquemoody:


library/Exceptions/ResourceTypeException.php
```diff
@@ -14,12 +14,14 @@
...
+ * Exception class for ResourceType.
+ *
  * @author Henrique Moody <henriquemoody@gmail.com>
  */
-class ResourceTypeException extends ValidationException
+final class ResourceTypeException extends ValidationException
 {
     /**
-     * @var array
+     * {@inheritdoc}
```
batusa: maybe better use `{{name}} must be of type resource` to keep consistent with the others?

henriquemoody: I kept it because for the others make sense to have "type" and "value", but for resources only "type" makes sense, so keeping as is sounds better to me. Same will happen to callable, null, iterable...

batusa: I see what you mean.

For resource and null is ok. But for `iterable`, `callable` and `object` (in 7.2), since we can type hint them, it might make sense to call them type indeed.

henriquemoody: Good point.

PR #950, opened 2018-04-25
Update type parameters for Validator doc blocks

cookieguru:
The [`@param`](http://docs.phpdoc.org/references/phpdoc/tags/param.html) tag uses [phpdoc types](http://docs.phpdoc.org/references/phpdoc/types.html).  However the [`@method`](http://docs.phpdoc.org/references/phpdoc/tags/method.html) tag uses PHP's [type declarations](http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration).  Therefore, `mixed` refers to a class whose literal name is mixed and not the [`mixed` keyword](http://docs.phpdoc.org/references/phpdoc/types.html#keyword).  Since `mixed` is not a valid class name in this context, the type paramaters should be removed.

Additionally, the `callable` type paramater can be used on the `callback` validation method.


henriquemoody:
Hey, Tim! I merged it manually and it is available on version 1.1.16!

The information in the pull request is very important, I recommend you do always put such information in your commit message, that's what I did with 020ea1ebb5dc626bb7f1958ff49f69172ff589cc.

Thanks for contributing! 🐼

Issue #921, opened 2018-02-24
Apply contribution guidelines


henriquemoody:
TL;DR
- Make sure you're based on the `master` branch
- Refactor all rules as in #902 
- Write a comment with the rule you're  going to work on before sending a PR

***
It's been a MAJOR or MINOR version of the library has been released. There are enough things to release the version 2.0 but I need your help.

I wrote new [Contribution Guidelines](https://github.com/Respect/Validation/blob/master/CONTRIBUTING.md) and before I creating the tag 2.0.0 I would like to have all rules with the same standard.

Applying the Contribution Guidelines consist of:
* Add docblocks for classes, methods, and properties
* Add your `@author` tag to classes you changed
* Set properties as private
* Set classes (Rule and Exception) as final
* Add PHP 7.1 type hinting
* Make rule test extend RuleTestCase (when possible) 
* Make sure there is 1 (and 1 only) PHPT file for each rule testing only exceptions

I've done some already but there is still a lot to go, write a comment with the rule name before you start working on it, so two people don't work on the same thing!

Hope you can help!

## Rules

- [ ] ~AllOf~ (needs refactoring)
- [x] Alnum
- [x] Alpha
- [x] AlwaysInvalid
- [x] AlwaysValid
- [x] ~AnyOf~ (needs refactoring)
- [x] ArrayType
- [x] ArrayVal
- [x] Attribute
- [x] Base
- [x] Base64
- [x] Between
- [x] BoolType
- [x] BoolVal
- [x] Bsn
- [x] Call
- [x] CallableType
- [x] Callback
- [x] Charset
- [x] Cnh
- [x] Cnpj
- [x] Cntrl
- [x] Consonant
- [x] Contains
- [x] ContainsAny
- [x] Countable
- [x] CountryCode
- [x] Cpf
- [x] CreditCard
- [x] CurrencyCode
- [x] Date
- [x] DateTime
- [x] Digit
- [x] Directory
- [x] ~Domain~ (needs refactoring)
- [x] Each
- [x] Email
- [x] EndsWith
- [x] Equals
- [x] Equivalent
- [x] Even
- [x] Executable
- [x] Exists
- [x] Extension
- [x] Factor
- [x] FalseVal
- [x] Fibonacci
- [x] File
- [x] FilterVar
- [x] Finite
- [x] FloatType
- [x] FloatVal
- [x] Graph
- [x] GreaterThan
- [x] ~HexRgbColor~ (needs refactoring)
- [x] Identical
[...]

PR #978, opened 2018-06-01
Apply contribution guidelines to "Contents" rule

williamespindola:


library/Rules/Contains.php
```diff
@@ -33,7 +55,12 @@ public function validate($input): bool
         return $this->validateEquals($input);
     }
 
-    protected function validateEquals($input)
+    /**
+     * {@inheritdoc}
+     *
+     * @param midex $input Input where $containsValue will be sought
```
batusa: `mixed` for the type.

library/Rules/Contains.php
```diff
@@ -33,7 +55,12 @@ public function validate($input): bool
         return $this->validateEquals($input);
     }
 
-    protected function validateEquals($input)
+    /**
+     * {@inheritdoc}
```
batusa: I don't think there is need for this one here.

library/Rules/Contains.php
```diff
@@ -33,7 +55,12 @@ public function validate($input): bool
...
+     *
+     * @param midex $input Input where $containsValue will be sought
+     */
+    protected function validateEquals($input): bool
```
batusa: could you make this method `private`?

library/Rules/Contains.php
```diff
@@ -44,7 +71,12 @@ protected function validateEquals($input)
         return false !== mb_stripos($inputString, $this->containsValue, 0, mb_detect_encoding($inputString));
     }
 
-    protected function validateIdentical($input)
+    /**
+     * {@inheritdoc}
+     *
+     * @param midex $input Input where $containsValue will be sought
+     */
+    protected function validateIdentical($input): bool
```
batusa: could you make this method `private`?

library/Rules/Contains.php
```diff
@@ -44,7 +71,12 @@ protected function validateEquals($input)
         return false !== mb_stripos($inputString, $this->containsValue, 0, mb_detect_encoding($inputString));
     }
 
-    protected function validateIdentical($input)
+    /**
+     * {@inheritdoc}
+     *
+     * @param midex $input Input where $containsValue will be sought
```
batusa: `mixed` for the type.

library/Rules/Contains.php
```diff
@@ -44,7 +71,12 @@ protected function validateEquals($input)
         return false !== mb_stripos($inputString, $this->containsValue, 0, mb_detect_encoding($inputString));
     }
 
-    protected function validateIdentical($input)
+    /**
+     * {@inheritdoc}
```
batusa: No need for this one here too.

tests/unit/Rules/ContainsTest.php
```diff
@@ -19,8 +19,14 @@
  * @group  rule
```
batusa: could you remove this extra space, please?

tests/unit/Rules/ContainsTest.php
```diff
@@ -19,8 +19,14 @@
...
  * @covers \Respect\Validation\Exceptions\ContainsException
+ *
+ * @author Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
+ * @author Gabriel Caruso <carusogabriel34@gmail.com>
+ * @author Henrique Moody <henriquemoody@gmail.com>
+ * @author Nawarian <nickolas@phpsp.org.br>
+ * @author William Espindola <oi@williamespindola.com.br>
  */
-class ContainsTest extends TestCase
+final class ContainsTest extends TestCase
```
batusa: Here we're doing usage of `RuleTestCase`. Could you do that change? [Example](https://github.com/Respect/Validation/pull/970/files).

batusa: All of the test cases from this class could be written through the providers when inheriting `RuleTestCase`.

library/Rules/Contains.php
williamespindola: for sure.

tests/unit/Rules/ContainsTest.php
williamespindola: Not in this case because the [Contains rule](https://github.com/Respect/Validation/blob/1.1/library/Rules/Contains.php#L19) has params on constructor.

henriquemoody:
There is a type in the commit message it was supposed to be "Contains"

library/Rules/Contains.php
```diff
@@ -13,17 +13,39 @@
 
 namespace Respect\Validation\Rules;
 
-class Contains extends AbstractRule
+/**
+ * @author Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
```
henriquemoody: Can you add an line before telling what this rule does? You will find that in [Contains.md](https://github.com/Respect/Validation/blob/master/docs/Contains.md)

tests/integration/rules/contains.phpt
```diff
@@ -0,0 +1,83 @@
...
+}
+
+try {
+    v::not(v::contains('foo'))->check('fool');
+} catch (ContainsException $exception) {
+    echo $exception->getMessage().PHP_EOL;
+}
+
+try {
+    v::not(v::contains('foo', true))->check(['bar', 'foo']);
```
henriquemoody: The goal of the integration tests is to just test how the exceptions behave. That said we only need to test `check()`, `assert()`, `not(check())`, and `not(assert()` when they throw exception.

Could you remove some extra tests and put them to the unit test instead if they are not there already? If though it's good to have tests it's also code to maintain, and the functionality should be tested in the unit tests.

tests/unit/Rules/ContainsTest.php
henriquemoody: That's no problem, check out the [EqualsTest](https://github.com/Respect/Validation/blob/master/tests/unit/Rules/EqualsTest.php)

tests/unit/Rules/ContainsTest.php
```diff
@@ -16,11 +16,18 @@
 use PHPUnit\Framework\TestCase;
 
 /**
- * @group  rule
+ * @group rule
+ *
  * @covers \Respect\Validation\Rules\Contains
  * @covers \Respect\Validation\Exceptions\ContainsException
```
henriquemoody: 🔪 We're testing the exceptions in the integration tests

tests/unit/Rules/ContainsTest.php
williamespindola: nice!

tests/integration/rules/contains.phpt
williamespindola: for sure!

tests/unit/Rules/ContainsTest.php
williamespindola: I got it!

library/Rules/Contains.php
```diff
@@ -13,17 +13,41 @@
 
 namespace Respect\Validation\Rules;
 
-class Contains extends AbstractRule
+/**
+ * Validates if the input contains some value
```
henriquemoody:
```diff
-- Validates if the input contains some value
++ Validates if the input contains some value.
```

library/Rules/Contains.php
```diff
@@ -13,17 +13,41 @@
...
+    private $containsValue;
+
+    /**
+     * @var bool
+     */
+    private $identical;
 
-    public function __construct($containsValue, $identical = false)
+    /**
+     * Initializes the Contains rule
```
henriquemoody:
```diff
-- Initializes the Contains rule
++ Initializes the Contains rule.
```

henriquemoody:
Can you also remove the `issue #921` I don't think that's useful information for the Git repository, while in GitHub we can make the associations the repository should speak to itself :)

williamespindola:
@henriquemoody take a look at the [commend](https://github.com/Respect/Validation/issues/921#issuecomment-393882069) on 921 issue about possible bug on this rule.

henriquemoody:
I checked https://github.com/Respect/Validation/issues/921#issuecomment-393882069. That's why I don't like rules with boolean flags, they're very confusing.

The way I see there is no point on having contains with identical unless if you're talking about arrays, if you want to talk about a different value you can use `Equals` or `Identical`. I'd rather split it up into `Contains($value)` and `ArrayContains($value, bool $identical = false)`

Perhaps in another PR?

henriquemoody:
I merged it as is in #985 (with just a few changes), let's not close this PR until we figure out what to do with Contains.

PR #979, opened 2018-06-01
Apply contribution guidelines to "Countable" rule

williamespindola:


tests/integration/rules/countable.phpt
```diff
@@ -0,0 +1,70 @@
...
+try {
+    v::countable()->check(new \stdClass());
+} catch (CountableException $exception) {
+    echo $exception->getMessage().PHP_EOL;
+}
+
+try {
+    v::countable()->assert('Not countable!');
+} catch (AllOfException $exception) {
+    echo $exception->getFullMessage().PHP_EOL;
```
henriquemoody: The goal of the integration tests is to just test how the exceptions behave. That said we only need to test `check()`, `assert()`, `not(check())`, and `not(assert()`. Could you remove some extra tests and put them to the unit test instead -- if they are not there already?

tests/unit/Rules/CountableTest.php
```diff
@@ -18,9 +18,16 @@
 /**
  * @group  rule
```
henriquemoody: There is an extra space here

tests/unit/Rules/CountableTest.php
```diff
@@ -18,9 +18,16 @@
...
  * @covers \Respect\Validation\Rules\Countable
```
henriquemoody: It's a good practice to have a space between _groups_ of annotations

library/Rules/Countable.php
```diff
@@ -13,8 +13,16 @@
 
 namespace Respect\Validation\Rules;
 
-class Countable extends AbstractRule
+/**
+ * @author Henrique Moody <henriquemoody@gmail.com>
```
henriquemoody: Can you add an line before telling what this rule does? You will find that in [Countable.md](https://github.com/Respect/Validation/blob/master/docs/Countable.md)

henriquemoody:
Merged by #985!

I imported the functions and classes used in the rule and also removed some extra tests in the PHPT (it should simply test the templates, the inputs are better tested in the unit tests).

Keep it up, panda! 🐼

PR #981, opened 2018-06-01
Apply contribution guidelines to "AlwaysValid" rule

williamespindola:


tests/unit/Rules/AlwaysValidTest.php
```diff
@@ -17,11 +17,20 @@
...
+ *
  * @covers \Respect\Validation\Rules\AlwaysValid
+ *
+ * @author Gabriel Caruso <carusogabriel34@gmail.com>
+ * @author Henrique Moody <henriquemoody@gmail.com>
+ * @author Paulo Eduardo <pauloelr@gmail.com>
+ * @author William Espindola <oi@williamespindola.com.br>
  */
-class AlwaysValidTest extends TestCase
+final class AlwaysValidTest extends TestCase
```
batusa: Could you inherit `RuleTestCase` to keep consistent with the other tests?

williamespindola: @batusa this rule always return valid and is not possible create a teste for invalid providers. To implement RuleTestCase we need to [implement valid and invalid providers](https://github.com/Respect/Validation/blob/master/tests/library/RuleTestCase.php#L41-L53), and in this case is not possible.

tests/integration/rules/always-valid.phpt
```diff
@@ -1,5 +1,3 @@
---TEST--
-alwaysValid()
 --FILE--
 <?php
 require 'vendor/autoload.php';
```
henriquemoody: This test does not cover anything with exceptions.

Could replace its content with `v::not(v::alwaysValid())->check(...)` and `v::not(v::alwaysValid())->assert(...)`?

henriquemoody:
Merged by #985.

Thanks for this pull request, dude! 🐼

PR #984, opened 2018-06-03
Apply contribution guidelines to "Lowercase" rule

DaniloBenevides:


library/Rules/Lowercase.php
```diff
@@ -13,10 +13,29 @@
 
 namespace Respect\Validation\Rules;
 
-class Lowercase extends AbstractRule
+use function is_string;
+use function mb_detect_encoding;
+use function mb_strtolower;
```
williamespindola: @Respect/troubleshooters do we have some rules about declare use of php functions like this? I don't do this on my commits and I don't see this in the code. I just don't know if is good to keep or not.

henriquemoody: I prefer to do that for two reasons, the first and more important one is that you can see what are the dependencies of the class instantly, the second and not that important, IMO, is performance (see https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/)

henriquemoody:
I merged it on #987!

I wrote something about the commit messages, see https://github.com/Respect/Validation/pull/986#issuecomment-394610428

Thanks for using your time for improving this library, it's very much appreciated! 🐼

PR #990, opened 2018-06-06
Apply contribution guidelines to "Negative" rule

IsmaelElias:


tests/unit/Rules/NegativeTest.php
```diff
@@ -13,65 +13,53 @@
...
-            [-0],
-            [null],
-            ['a'],
-            [' '],
-            ['Foo'],
-            [16],
-            ['165'],
-            [123456],
-            [1e10],
+            [$rule, ''],
```
henriquemoody: Because this PR is adding a new behaviour it should have added a few more tests, and updates the changelog of the rule.

PR #991, opened 2018-06-06
Apply contribution guidelines to "Odd" rule

DaniloBenevides:


tests/integration/rules/odd.phpt
```diff
@@ -0,0 +1,41 @@
...
+}
+
+?>
+--EXPECTF--
+2 must be an odd number
+- 2 must be an odd number
+7 must not be an odd number
+- 9 must not be an odd number
+
+
```
henriquemoody: 🔪

tests/integration/rules/odd.phpt
```diff
@@ -0,0 +1,41 @@
...
+    v::not(v::odd())->assert(9);
+} catch (NestedValidationException $exception) {
+    echo $exception->getFullMessage().PHP_EOL;
+}
+
+?>
+--EXPECTF--
+2 must be an odd number
+- 2 must be an odd number
+7 must not be an odd number
```
henriquemoody: I prefer to first test `check()`, and then `assert()`, it makes easy to read.

tests/unit/Rules/OddTest.php
```diff
@@ -13,60 +13,47 @@
...
     {
+        $rule = new Odd();
+
         return [
-            [''],
-            [-2],
-            [-0],
-            [0],
-            [32],
+            [$rule, ''],
```
henriquemoody: Because this PR is adding a new behaviour it should have added a few more tests, and updates the changelog of the rule.

PR #993, opened 2018-06-06
Added .studio tld to TLD domain validator

r--w:
https://icannwiki.org/.studio


henriquemoody:

This patch is now available in the version 1.1.17!

PR #998, opened 2018-06-09
Apply contribution guidelines to "Base" rule

williamespindola:


library/Exceptions/BaseException.php
```diff
@@ -13,6 +13,11 @@
 
 namespace Respect\Validation\Exceptions;
 
-class BaseException extends ValidationException
+/**
+ * @author Carlos André Ferrari <caferrari@gmail.com>
+ * @author Henrique Moody <henriquemoody@gmail.com>
+ * @author William Espindola <not.committed.yet>
```
henriquemoody: You're email is not correct

tests/integration/rules/base.phpt
```diff
@@ -0,0 +1,38 @@
...
+    v::not(v::base(2))->assert('011010001');
+} catch (NestedValidationException $exception) {
+    echo $exception->getFullMessage().PHP_EOL;
+}
+?>
+--EXPECTF--
+"Z01xSsg5675hic20dj" must be valid
+- "" must be valid
+"011010001" must not be valid
+- "011010001" must not be valid
```
henriquemoody: Missing new line at the end of the file

tests/integration/rules/base.phpt
```diff
@@ -0,0 +1,38 @@
...
+
+try {
+    v::not(v::base(2))->assert('011010001');
+} catch (NestedValidationException $exception) {
+    echo $exception->getFullMessage().PHP_EOL;
+}
+?>
+--EXPECTF--
+"Z01xSsg5675hic20dj" must be valid
+- "" must be valid
```
henriquemoody: I'd rather have test the `check()` first and the `assert()` afterwards

library/Exceptions/BaseException.php
```diff
@@ -13,6 +13,11 @@
...
+ */
+final class BaseException extends ValidationException
 {
```
henriquemoody: Could you be so kind to also add some messages to this exception?

```php
    /**
     * {@inheritdoc}
     */
        self::MODE_DEFAULT => [
            self::STANDARD => '{{name}} must be a number in the base {{base}}',
        ],
        self::MODE_NEGATIVE => [
            self::STANDARD => '{{name}} must not be a number in the base {{base}}',
        ],
    ];
```

library/Rules/Base.php
```diff
@@ -14,12 +14,37 @@
...
+     */
+    private $base;
 
+    /**
+     * Initializes the Base rule.
+     *
+     * @param int $base
+     * @param string $chars
+     */
     public function __construct($base = null, $chars = null)
```
henriquemoody:
```diff
-- public function __construct($base = null, $chars = null)
++ public function __construct(int $base, string $chars = null)
```

henriquemoody:
Merged by #1001 with just a few changes.

Thank you very much! 🐼

PR #999, opened 2018-06-09
Apply contribution guidelines to "Base64" rule

williamespindola:


library/Exceptions/Base64Exception.php
```diff
@@ -13,8 +13,16 @@
 
 namespace Respect\Validation\Exceptions;
 
-class Base64Exception extends ValidationException
+/**
+ * @author Henrique Moody <henriquemoody@gmail.com>
+ * @author Jens Segers <segers.jens@gmail.com>
+ * @author William Espindola <not.committed.yet>
```
henriquemoody: You're email is not correct

tests/integration/rules/base64.phpt
```diff
@@ -7,6 +7,18 @@ use Respect\Validation\Exceptions\AllOfException;
...
 
+try {
+    v::base64()->check('=c3VyZS4');
+} catch (Base64Exception $exception) {
+    echo $exception->getMessage().PHP_EOL;
+}
+
+try {
+    v::base64()->assert('=c3VyZS4');
+} catch (AllOfException $exception) {
```
henriquemoody: Use `NestedValidationException` instead

tests/integration/rules/base64.phpt
```diff
@@ -20,5 +32,7 @@ try {
 }
 ?>
 --EXPECTF--
+"=c3VyZS4" must be Base64-encoded
+- "=c3VyZS4" must be Base64-encoded
```
henriquemoody: I'd rather have first the `check()` and then the `assert()`. It makes easier to read, IMO. How about you?

henriquemoody:
Merged by #1001 with just a few changes.

Thank you very much! 🐼

PR #1002, opened 2018-06-10
Apply contribution guidelines to "Bsn" rule

williamespindola:


library/Rules/Bsn.php
```diff
@@ -16,11 +16,13 @@
 /**
```
henriquemoody: Could you also add the use statements for the functions?

tests/unit/Rules/BsnTest.php
```diff
@@ -13,86 +13,58 @@
...
-            ['087880532'],
-            ['386625918'],
-            ['601608021'],
-            ['254650703'],
-            ['478063441'],
-            ['478063441'],
-            ['187368429'],
-            ['541777348'],
-            ['254283883'],
+            [new Bsn(), '612890053'],
```
henriquemoody: Since is has not constructor, you can do something like that:

```php
$rule = new Bsn();

return [
    [$rule, '612890053'],
    ....
```

tests/unit/Rules/BsnTest.php
```diff
@@ -13,86 +13,58 @@
...
-            ['854650703'],
-            ['283958721'],
-            ['231859081'],
-            ['189023323'],
-            ['238150912'],
-            ['382409678'],
-            ['38240.678'],
-            ['38240a678'],
-            ['abcdefghi'],
+            [new Bsn(), '1234567890'],
```
henriquemoody: Since is has not constructor, you can do something like that:

```php
$rule = new Bsn();

return [
    [$rule, '612890053'],
    ....
```

PR #1006, opened 2018-06-11
Apply contribution guidelines to some rules

henriquemoody:


library/Exceptions/UrlException.php
```diff
@@ -17,10 +17,10 @@ class UrlException extends ValidationException
 {
     public static $defaultTemplates = [
         self::MODE_DEFAULT => [
-            self::STANDARD => '{{name}} must be an URL',
+            self::STANDARD => '{{name}} must be a URL',
         ],
         self::MODE_NEGATIVE => [
-            self::STANDARD => '{{name}} must not be an URL',
+            self::STANDARD => '{{name}} must not be a URL',
```
williamespindola: Why change `an` to `a`?

tests/unit/Rules/FilterVarTest.php
```diff
@@ -13,64 +13,57 @@
...
-    }
-
-    /**
+     * @test
+     *
      * @expectedException \Respect\Validation\Exceptions\ComponentException
      * @expectedExceptionMessage Cannot accept the given filter
      */
-    public function testShouldThrowsExceptionWhenFilterIsNotValid(): void
+    public function itShouldThrowsExceptionWhenFilterIsNotValid(): void
```
williamespindola: Why not move this exception test to phpt test?

henriquemoody: Because this exception is a condition of the constructor, not a failure in the validation.

PR #1009, opened 2018-06-12
Apply contribution guidelines to "Cnpj" rule

williamespindola:
This rule can be used with `->digits()` we need to assert this behaviour with this rule in your tests?


library/Rules/Cnpj.php
```diff
@@ -13,8 +13,18 @@
...
 namespace Respect\Validation\Rules;
 
-class Cnpj extends AbstractRule
+use function is_scalar;
+use function mb_strlen;
+use function preg_replace;
+
+/**
+ * Validates the brazilian CNPJ number.
+ */
```
henriquemoody: Missing `@author` tags

library/Rules/Cnpj.php
```diff
@@ -13,8 +13,18 @@
 
 namespace Respect\Validation\Rules;
 
-class Cnpj extends AbstractRule
+use function is_scalar;
+use function mb_strlen;
+use function preg_replace;
+
+/**
+ * Validates the brazilian CNPJ number.
```
henriquemoody:
```diff
-- Validates the brazilian CNPJ number.
++ Validates if the input is a Brazilian National Registry of Legal Entities (CNPJ) number.
```

henriquemoody:
I created #1018 with the rebase version of this commit! 🐼 

Thank you very much for this PR!

PR #1010, opened 2018-06-13
Apply contribution guidelines to "CountryCode" rule

williamespindola:


tests/unit/Rules/CountryCodeTest.php
```diff
@@ -30,59 +36,33 @@ public function testShouldThrowsExceptionWhenInvalidFormat(): void
...
+        return [
+            [new CountryCode(CountryCode::ALPHA2),  'BR'],
+            [new CountryCode(CountryCode::ALPHA3),  'BRA'],
+            [new CountryCode(CountryCode::NUMERIC), '076'],
+            [new CountryCode(CountryCode::ALPHA2),  'DE'],
+            [new CountryCode(CountryCode::ALPHA3),  'DEU'],
+            [new CountryCode(CountryCode::NUMERIC), '276'],
+            [new CountryCode(CountryCode::ALPHA2),  'US'],
+            [new CountryCode(CountryCode::ALPHA3),  'USA'],
+            [new CountryCode(CountryCode::NUMERIC), '840'],
```
williamespindola: I don't know, this `CountryCode(CountryCode::FOO)` getting the param from the same class seems not good 😕

tests/unit/Rules/CountryCodeTest.php
```diff
@@ -30,59 +36,33 @@ public function testShouldThrowsExceptionWhenInvalidFormat(): void
...
-    public function testShouldUseISO3166Alpha2ByDefault(): void
-    {
-        $country = new CountryCode();
-        self::assertEquals(CountryCode::ALPHA2, $country->set);
-    }
-
-    public function testShouldDefineACountryFormatOnConstructor(): void
-    {
-        $country = new CountryCode(CountryCode::NUMERIC);
-        self::assertEquals(CountryCode::NUMERIC, $country->set);
```
williamespindola: I remove this tests because I change the `set` property to private.

library/Rules/CountryCode.php
```diff
@@ -307,7 +333,7 @@ public static function getAvailableSets()
         ];
     }
 
-    private function getCountryCodeList($index)
+    private function getCountryCodeList($index): array
```
batusa: here we can also type hint the `$index` parameter to `int`.

tests/unit/Rules/CountryCodeTest.php
batusa: you mean the constants `NUMERIC`, `ALPHA2`, and `ALPHA3`? I think it's fine.

PR #1013, opened 2018-06-14
fixes #856

### LINKED ISSUE #856: Url validation error message grammatical issue
Url validation error states `an Url` which I believe should be `a URL`.

dmirogin:


henriquemoody:
This is already fixed in the `master` branch. I'm not 100% sure whether we should release that in a patch version or not, because it kinda break backwards compatibility (in case of translations).

dmirogin:
Sorry, tough day. Close #856 please so

PR #1017, opened 2018-06-15
Apply contribution guidelines to "Type" rule

koficodes:
Apply contribution guidlines to "Type" rule as listed in #921 

Also add integration test for Type rule.

### DIFF
--- a/library/Rules/Type.php
+++ b/library/Rules/Type.php
@@ -14,11 +14,34 @@ declare(strict_types=1);
 namespace Respect\Validation\Rules;
 
 use Respect\Validation\Exceptions\ComponentException;
+use function gettype;
+use function is_callable;
+use function mb_strtolower;
+use function print_r;
+use function sprintf;
 
-class Type extends AbstractRule
+/**
+ * Validates the type of input.
+ *
+ * @author Gabriel Caruso <carusogabriel34@gmail.com>
+ * @author Henrique Moody <henriquemoody@gmail.com>
+ * @author Paul Karikari<paulkarikari1@gmail.com>
+ */
+final class Type extends AbstractRule
 {
-    public $type;
-    public $availableTypes = [
+    /**
+     * Type to validate input against.
+     *
+     * @var string
+     */
+    private $type;
+
+    /**
+     * Collection of available types for validation.
+     *
+     * @var array
+     */
+    private $availableTypes = [
         'array' => 'array',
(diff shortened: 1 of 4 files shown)


tests/unit/Rules/TypeTest.php
```diff
@@ -13,94 +13,51 @@
 
 namespace Respect\Validation\Rules;
 
-use PHPUnit\Framework\TestCase;
+use Respect\Validation\Test\RuleTestCase;
 use stdClass;
+use function tmpfile;
 
 /**
  * @group  rule
```
henriquemoody:
```diff
-- * @group  rule
++ * @group  rule
```

tests/unit/Rules/TypeTest.php
```diff
@@ -13,94 +13,51 @@
...
+ *
  * @covers \Respect\Validation\Rules\Type
- * @covers \Respect\Validation\Exceptions\TypeException
+ *
+ * @author Henrique Moody <henriquemoody@gmail.com>
+ * @author Paul Karikari <paulkarikari1@gmail.com>
  */
-class TypeTest extends TestCase
+class TypeTest extends RuleTestCase
```
henriquemoody:
```diff
-- class TypeTest extends RuleTestCase
++ final class TypeTest extends RuleTestCase
``

File: tests/unit/Rules/TypeTest.php [hunk 3]
```diff
@@ -13,94 +13,51 @@
...
-        $rule = new Type('InTeGeR');
-
-        self::assertTrue($rule->validate(42));
-    }
-
-    /**
-     * @expectedException \Respect\Validation\Exceptions\ComponentException
-     * @expectedExceptionMessage "whatever" is not a valid type
-     */
-    public function testShouldThrowExceptionWhenTypeIsNotValid(): void
```
henriquemoody: This test should be kept

library/Rules/Type.php
```diff
@@ -33,7 +56,12 @@ class Type extends AbstractRule
...
 
-    public function __construct($type)
+    /**
+     * Creates new Type rule instance.
+     *
+     * @param string $type
+     */
+    public function __construct(string $type)
     {
         $lowerType = mb_strtolower($type);
```
henriquemoody: Out of scope: I will make it case sensitive instead

PR #1022, opened 2018-06-16
Apply contribution guidelines to "Each" rule

williamespindola:


williamespindola:
Tell me more.

henriquemoody:
Oh, sorry. I thought I had written it down somewhere.

I think this rule can just validate values. For keys we can put in the documentation something like

```php
v::call('array_keys', v::each(v::intVal()))->validate([1 => true, 2 => false]);
```

henriquemoody:
I sent #1034 with the changes I suggested, please check the commit f8715445efd573891cecfbdd8a6cd3dcf519a015 to see the changes.

Because of that I also found a bug which is reported on #1033.

Thanks for this PR! 🐼

PR #1024, opened 2018-06-17
Apply contribution guidelines to "Uploaded" rule

koficodes:
This PR applies some of the guidelines listed in #921.

**Test didn't extend RuleTestCase.**
I keep getting false instead of true when the "shouldValidateValidInput" method of "RuleTestCase" class is called.

**Didn't include .phpt file**
Adding an integration test for files uploaded means the test will be system dependent ie. it will depend on the specific file path given which I think is not ideal.

### DIFF
--- a/library/Exceptions/UploadedException.php
+++ b/library/Exceptions/UploadedException.php
@@ -13,8 +13,18 @@ declare(strict_types=1);
 
 namespace Respect\Validation\Exceptions;
 
-class UploadedException extends ValidationException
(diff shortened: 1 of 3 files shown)


henriquemoody:
I sent #1029 with a few changes! I'm using [uopz](http://php.net/uopz) to overwrite PHP's `is_uploaded_file()` function so we can test everything!

Thanks for contributing! 🐼

PR #1025, opened 2018-06-17
Apply contribution guidelines to "Executable" rule

williamespindola:


williamespindola:
The tests don't pass for php 7.1, I saw that Generator class not found on phpunit on this php version. I will fix this soon

henriquemoody:
I sent #1029 with a few changes! I'm just not using the mock and instead using a file inside "tests/fixtures" directory. Makes it a bit simpler

Thanks for using your time on this! 🐼

PR #1030, opened 2018-06-20
Update composer.json

Mecit:
Suggest friendsofphp/php-cs-fixer instead of fabpot/php-cs-fixer which is no longer maintained.


henriquemoody:
Thanks for this PR, @Mecit! 🐼 

This has been fixed in `master` already, but it's a good idea to also have that in the `1.1`.

Issue #1038, opened 2018-06-27
IntVal validates for `boolean true`


precariouspanther:
https://github.com/Respect/Validation/commit/54d17abcee9c6e734691754c77641a5f51ecc608#r29514787

Recent update to `IntVal::validate()` means passing a boolean `true` passes validation.

precariouspanther:
Candidate fix PR: https://github.com/Respect/Validation/pull/1039

henriquemoody:
Fixed by f5ca4a298a4475a64f0d8526041c1a58167e8c7c

Thanks for reporting and sending the fix, @precariouspanther! 🐼

PR #1039, opened 2018-06-27
#1038 Fix IntVal incorrectly passing on boolean true

precariouspanther:
Fix for https://github.com/Respect/Validation/issues/1038

Recent update to improve `IntVal` to prevent decimal-like values being accepted as integers introduced a regression allowing boolean `true` to be accepted as integer. This change prevents booleans from being considered valid values.

@henriquemoody @pascalzajac

### DIFF
--- a/library/Rules/IntVal.php
+++ b/library/Rules/IntVal.php
@@ -19,6 +19,10 @@ class IntVal extends AbstractRule
             return false;
         }
 
+        if (is_bool($input)) {
+            return false;
+        }
+
         return false !== filter_var($input, FILTER_VALIDATE_INT);
     }
 }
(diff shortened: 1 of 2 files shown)


bc-luke:
Can you use `IntType` instead? I believe the distinction between `IntType` and `IntVal` are strictness of type.

precariouspanther:
@bc-luke This was introduced as part of a change to `IntVal` to make it reject decimal-types that are implicitly converts to integers - but was a regression as previously boolean trues were rejected.

It's not so much that I'm requesting this to be more strict, just that all of the existing code that has been written depending on the current behaviour doesn't break.

henriquemoody:
Merged manually by f5ca4a298a4475a64f0d8526041c1a58167e8c7c

Thank you very much for contributing! I will be in the next path version of 1.1!

🐼

PR #1040, opened 2018-06-28
Fix wrong behavior with nested exceptions

### LINKED ISSUE #1033: Each does not throw a readable message
While executing the code:

```php
try {
    v::each(v::equals(1))->assert([1, 2, 3]);
} catch (Exception $exception) {
}
```

The **expected output** would be:

```
- Each item in `{ "A", "B", "B" }` must be valid
  - "A" must equal 1
  - "B" must equal 1
  - "B" must equal 1
```

Instead the **actual output** is:
```
- All of the required rules must pass for `{ "A", "B", "B" }`
- Each item in `{ "A", "B", "B" }` must be valid
- "A" must equal 1
- "B" must equal 1
- "B" must equal 1
```

henriquemoody:
Fix #1033


library/Exceptions/NestedValidationException.php
```diff
@@ -195,18 +199,21 @@ public function getMessages()
...
         }
 
+        $exceptions = $this->getIterator();
         foreach ($exceptions as $exception) {
-            $depth = $exceptions[$exception]['depth'];
-            $prefix = str_repeat(' ', $depth * 2);
-            $messages[] = sprintf('%s%s %s', $prefix, $marker, $exception->getMessage());
+            $messages[] = sprintf(
+                '%s- %s',
+                str_repeat(' ', ($exceptions[$exception]['depth'] - $leveler) * 2),
```
malukenho: 🎩 🌟 `2`

PR #1056, opened 2018-07-23
Simplify "AbstractComposite" class

henriquemoody:
There is a log of magic in the "AbstractComposite" that allows an user
to add a rule in multiple ways, remove rules, verify if a rule is
inserted or not. Because of that the class has a lot of complexity and
for unnecessary reasons since we do not use these features internally.

The ideal would be to make this class immutable, however due to many
usages is not possible to do it now, but that is the plan for a near
future - hopefully.

### DIFF
--- a/library/Rules/AbstractComposite.php
+++ b/library/Rules/AbstractComposite.php
@@ -15,17 +15,35 @@ namespace Respect\Validation\Rules;
 
 use Respect\Validation\Exceptions\ValidationException;
 use Respect\Validation\Validatable;
-use Respect\Validation\Validator;
+use function array_filter;
+use function array_map;
 
+/**
+ * Abstract class for rules that are composed by other rules.
+ *
+ * @author Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
+ * @author Henrique Moody <henriquemoody@gmail.com>
+ */
 abstract class AbstractComposite extends AbstractRule
 {
-    protected $rules = [];
+    /**
+     * @var Validatable[]
+     */
+    private $rules = [];
 
-    public function __construct(...$rule)
+    /**
+     * Initializes the rule adding other rules to the stack.
+     *
+     * @param Validatable ...$rules
+     */
+    public function __construct(Validatable ...$rules)
     {
-        $this->addRules($rule);
+        $this->rules = $rules;
     }
 
+    /**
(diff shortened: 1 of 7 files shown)


library/Rules/AbstractComposite.php
```diff
@@ -15,17 +15,35 @@
...
-    protected $rules = [];
-
-    public function __construct(...$rule)
+    /**
+     * @var Validatable[]
+     */
+    private $rules = [];
+
+    /**
+     * Initializes the rule.
```
batusa: Doesn't add much.

library/Rules/AbstractComposite.php
```diff
@@ -15,17 +15,35 @@
...
+     *
+     * @param Validatable ...$rules
+     */
+    public function __construct(Validatable ...$rules)
```
batusa: 👍

library/Rules/AbstractComposite.php
```diff
@@ -41,86 +59,61 @@ public function setName(string $name): Validatable
...
 
-    public function hasRule($validator)
+    /**
+     * Validate the whole stack of rule.
+     *
+     * @param mixed $input
+     *
+     * @return ValidationException[]
+     */
+    protected function validateRules($input): array
```
batusa:
```diff
- protected function validateRules($input): array
+ protected function validateWithRules($input): array
```

what do you think?

library/Rules/AbstractComposite.php
```diff
@@ -41,86 +59,61 @@ public function setName(string $name): Validatable
...
-        foreach ($validators as $v) {
-            try {
-                $v->assert($input);
-            } catch (ValidationException $e) {
-                $exceptions[] = $e;
-            }
-        }
-
-        return $exceptions;
+        return null !== $this->getName() && null === $rule->getName();
```
batusa: maybe add a `hasName(): bool` would be better? since name is optional, then we could check for it.

henriquemoody: That would be handy. In other hand I think the `Validatable` already has too many methods

library/Rules/AbstractComposite.php
henriquemoody: I will do that :)

PR #1059, opened 2018-07-25
Fix postalCode for Portugal (PT)

axessweb:


henriquemoody:
Thanks for contributing! 🐼 

This should be already available in version 1.1.20

PR #1060, opened 2018-07-28
Fix casting error in AbstractInterval classes

### LINKED ISSUE #908: AbstractInterval->filterInterval() returns DateTime and that ends with error!
Code: v::intVal()->between(1900, 2040)->setName('year')

everything works fine, until input is something that can parsed as date. For example **2018-01-25**, then in Max/Min classes, input is filtered by filterInterval() but as it returns DateTime object, casting error happens:

"**Object of class DateTime could not be converted to int**"

as there will be comparsion like this: dateTimeObject < 1900

Ideas:
* casting to DateTime should not happen, or do separate classes BetweenDate, BetweenInt
* add class untilFirstFail or smth, which can take rule list like OneOf/AllOf, but rules are checked until first error (so we would not compare() if intVal() already fails)
* before filter return false if input is not number?
* return false if types are not comparable?
* return false if (only?) one is object?

Comparsion problem if converted to DateTime()
'2018-01-10' < '518-01-11' => true (string 2.. is smaller than 5..)
new DateTime('2018-01-10') < new DateTime('518-01-11') => false (year 2018 greater than 518)

henriquemoody:
The classes that are children of "AbstractInterval" convert their values
before comparing them.

Because PHP tries to convert values when making comparisons and an
"DateTime" object cannot be converted to integer or float some
validations would result into PHP triggering an error like:

> Object of class DateTime could not be converted to int
> Object of class DateTime could not be converted to float

This commit prevents that to happen by verifying if both compared values
are scalar or not before comparing them with each other.

Fix #908

### DIFF
--- a/library/Rules/AbstractInterval.php
+++ b/library/Rules/AbstractInterval.php
@@ -25,6 +25,11 @@ abstract class AbstractInterval extends AbstractRule
         $this->inclusive = $inclusive;
     }
 
+    protected function isAbleToCompareValues($left, $right)
+    {
+        return is_scalar($left) === is_scalar($right);
+    }
+
     protected function filterInterval($value)
     {
         if (!is_string($value) || is_numeric($value) || empty($value)) {
--- a/library/Rules/Max.php
+++ b/library/Rules/Max.php
@@ -15,10 +15,17 @@ class Max extends AbstractInterval
 {
     public function validate($input)
     {
-        if ($this->inclusive) {
-            return $this->filterInterval($input) <= $this->filterInterval($this->interval);
+        $filteredInput = $this->filterInterval($input);
+        $filteredInterval = $this->filterInterval($this->interval);
+
+        if (!$this->isAbleToCompareValues($filteredInput, $filteredInterval)) {
+            return false;
         }
 
-        return $this->filterInterval($input) < $this->filterInterval($this->interval);
+        if ($this->inclusive) {
+            return $filteredInput <= $filteredInterval;
+        }
+
+        return $filteredInput < $filteredInterval;
     }
 }
--- a/library/Rules/Min.php
+++ b/library/Rules/Min.php
@@ -15,10 +15,17 @@ class Min extends AbstractInterval
 {
     public function validate($input)
     {
-        if ($this->inclusive) {
-            return $this->filterInterval($input) >= $this->filterInterval($this->interval);
+        $filteredInput = $this->filterInterval($input);
+        $filteredInterval = $this->filterInterval($this->interval);
+
+        if (!$this->isAbleToCompareValues($filteredInput, $filteredInterval)) {
+            return false;
         }
 
-        return $this->filterInterval($input) > $this->filterInterval($this->interval);
+        if ($this->inclusive) {
+            return $filteredInput >= $filteredInterval;
+        }
+
+        return $filteredInput > $filteredInterval;
     }
 }
--- a/tests/unit/Rules/MaxTest.php
+++ b/tests/unit/Rules/MaxTest.php
@@ -62,6 +62,8 @@ class MaxTest extends \PHPUnit_Framework_TestCase
             [200, false, 250],
             [200, false, 1500],
             [200, false, 200],
+            [1900, false, '2018-01-25'],
+            [10.5, false, '2018-01-25'],
         ];
     }
 }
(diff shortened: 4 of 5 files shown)


library/Rules/AbstractInterval.php
```diff
@@ -25,6 +25,11 @@ public function __construct($interval, $inclusive = true)
         $this->inclusive = $inclusive;
     }
 
+    protected function isAbleToCompareValues($left, $right)
+    {
+        return is_scalar($left) === is_scalar($right);
```
batusa: both should be either `true` or `false`?

henriquemoody: Exactly!

Do you think there is a way to make it more clear?

batusa: I can't think of anything at the moment.

I just wanted to double check this method behavior with you ;)

nickl-: I have to agree with @batusa this feels like we are patching the symptoms instead of addressing the problem.

Instead of validating interval we are now validating comparability, is this the desirable conclusion? This also begs the question, is this comparability issue only particular to intervals and why?

If I understood the problem at #908 correctly, `intVal` allows for date/time values shouldn't this be where we apply comparability?

If we do allow comparison with both scalar and date/time values and PHP is unable to successfully do the conversion, then we should apply the appropriate conversion to make this a valid comparison, IMO. Perhaps converting date/time values to UNIX Timestamp makes more sense. 

Another approach might be to split these rules for either int value or date/time comparison because the combination is not compatible.

henriquemoody: What I got from #908 is that the values 1900 and 2040 that are used in `Between` are meant to validate a year, as integer only, and as @urmasrosenberg states _everything works fine, until input is something that can parsed as date_.

That's because what he is expecting is an integer, he even included `IntVal` in the chain to ensure the expected behavior.

The root of the problem is that Between tries to convert any value that can be interpreted as a date/time into a `DateTime` instance. When we compare the integers with the `DateTime` object PHP triggers an error because it cannot convert the `DateTime` into that scalar value.

This simply ensures that PHP will not try to convert a `DateTime` object into a scalar value before we actual do that (when we compare these values).

Converting it into timestamp would be very misleading because you it will allow users to compare _dates_ with scalar values -- this PR ensures that you are comparing dates with dates, and scalar values with scalar values.

Splitting this rule into multiple ones may be a solution -- even thought I think it may make us lose a bit of Validation's swagger --, however we shouldn't do it in a patch version.

nickl-: > This PR ensures that PHP will not try to convert a DateTime object into a scalar value before we actually compare them.

I agree that this is one of the side effects but as implemented we are failing the rule if not comparable which is not correct. We are loosing the actual reason for failing the rule which was not because the value was bigger, smaller or not between but because it was not able to compare both scalars or not scalars.

The php error is then more appropriate IMO, especially if trying to debug why the validation failed.

Perhaps adding an "is same type" rule (do we have such a thing?) to the validation chain is a more appropriate work around to solve this issue. This way we retain the reason for failing the test. Error: "2018-07-29 is not comparable to scalar type 2042."

I just feel the solution should not hide the actual reason for failing the test or change the validation rule specification in any way. This may lead to unnecessary frustrations which could have been avoided.

 > Splitting this rule into multiple ones may be a solution -- even thought I think it may make us lose a bit of Validation's swagger

I agree that type flexibility is a feature and not a bug which needs to be fixed, -1 on splitting the rules.

henriquemoody: > The php error is then more appropriate IMO, especially if trying to debug why the validation failed.

I think Validation should never trigger PHP errors during a validation. We could throw another validation message, but I think it would feel weird if I use `max()` and I get a `IntValException` from it. Maybe add another message to it would be an alternative... (not for a patch version)

> Perhaps adding an "is same type" rule (do we have such a thing?) to the validation chain is a more appropriate work around to solve this issue.

The example in #908 uses `InvVal` in the chain which would do the job if the intention was to validate with `check()`, meaning that after one rule fails Validation would stop the validation chain. Because it uses `assert()` all rules in the chain are validated. That's why rules should be self contained.

The question _is "2018-01-25" bigger than 1900?_ can be "no" because it doesn't make sense. That would be the same of asking _is a Ferrari faster than a MacBook?_ both can have their speed compared but in different ways. So I think it's fine fail the validation if you cannot validate and we do something similar in other rules as well, check out `Regex` and `NoWhitespace`, for example.

nickl-: True, at least it does choke the PHP error and ensures we are not comparing apples with stallions. Would've been nice to have a skip test status or similar instead of failing but yah!

+1 for fail on not comparable patch

PR #1061, opened 2018-08-01
fix IntVal() leading zero behavior

andrdru:
54d17abcee9c6e734691754c77641a5f51ecc608 (1.1.18) changes IntVal() behavior on leading zero validations ("06")


henriquemoody:

The fix should be already available in the version 1.1.22!

PR #1063, opened 2018-08-01
Few improvements on the Validator class

henriquemoody:


library/Validator.php
```diff
@@ -182,51 +184,42 @@ public function check($input): void
     }
 
     /**
+     * Creates a new instance Validator with a rule that was called on the static method.
```
batusa:
```diff
- Creates a new instance Validator with a rule that was called on the static method.
+ Creates a new Validator instance with a rule that was called on the static method.
```

library/Validator.php
```diff
@@ -182,51 +184,42 @@ public function check($input): void
...
      * Create instance validator.
      *
      * @return Validator
      */
-    public static function create()
+    public static function create(): self
     {
-        $ref = new ReflectionClass(__CLASS__);
-
-        return $ref->newInstanceArgs(func_get_args());
```
batusa: I wonder why we had this `func_get_args()`, for nothing? 🤔

henriquemoody: Was to allow `v::create(new Equals('somehing'))`

PR #1066, opened 2018-08-02
Apply contribution guidelines to "NestedValidationException"

henriquemoody:


library/Exceptions/NestedValidationException.php
```diff
@@ -22,76 +22,75 @@
...
+            $this->exceptions = new SplObjectStorage();
+        }
 
-        return $this;
+        return $this->exceptions;
     }
 
     /**
-     * @return RecursiveIteratorIterator
+     * Adds a child to the exception.
```
batusa: > Adds a child

Would we call it a `child` or a `related` exception?

batusa: In the class comment you also talked about `children`.

library/Exceptions/NestedValidationException.php
```diff
@@ -22,76 +22,75 @@
...
+     *
+     * @param ValidationException $exception
+     *
+     * @return self
      */
-    private function getRecursiveIterator()
+    public function addRelated(ValidationException $exception)
```
batusa: Return type hint.

library/Exceptions/NestedValidationException.php
```diff
@@ -22,76 +22,75 @@
...
      *
-     * @return self
+     * @return SplObjectStorage|ValidationException[]
      */
-    public function addRelated(ValidationException $exception)
+    public function getRelated(): SplObjectStorage
     {
-        $this->getRelated()->attach($exception);
+        if (!$this->exceptions instanceof SplObjectStorage) {
+            $this->exceptions = new SplObjectStorage();
```
batusa: what about initialize it in the constructor?

library/Exceptions/NestedValidationException.php
```diff
@@ -22,76 +22,75 @@
...
-        );
+        $this->getRelated()->attach($exception);
 
-        return $recursiveIteratorIterator;
+        return $this;
     }
 
     /**
-     * Returns weather an exception should be omitted or not.
+     * Adds children to the exception.
```
batusa:
```diff
- Adds children to the exception.
+ Adds children to exceptions.
```

library/Exceptions/NestedValidationException.php
```diff
@@ -118,20 +117,25 @@ public function getIterator()
...
+            $childrenExceptions->attach($childException, $currentDepth);
         }
 
         return $childrenExceptions;
     }
 
+    /**
+     * Returns a key->value array with all the messages of the exception.
+     *
+     * In this array they "keys" are the ids of the exceptions (defined name or
```
batusa:
```diff
- In this array they "keys" are the ids of the exceptions (defined name or
+ In this array the "keys" are the ids of the exceptions (defined name or

File: library/Exceptions/NestedValidationException.php [hunk 6]
```diff
@@ -118,20 +117,25 @@ public function getIterator()
...
+     * name of the rule) and the value are the message.
```
batusa:
```diff
- name of the rule) and the value are the message.
+ name of the rule) and the values are the message.
```

library/Exceptions/NestedValidationException.php
```diff
@@ -177,38 +183,43 @@ public function getFullMessage()
...
+        );
     }
 
-    /**
-     * @param array $exceptions
-     *
-     * @return self
-     */
-    public function setRelated(array $exceptions)
+    private function isOmissible(Exception $exception): bool
```
batusa: Moved private method to the bottom, nice 👍

library/Exceptions/NestedValidationException.php
henriquemoody: I'd rather keep it for now. I wanna get a rid of `SplObjectStorage` soon because it allows users to add any objects to the stack

library/Exceptions/NestedValidationException.php
henriquemoody: When... Indeed, even thought they're children we should call them related unless we rename these methods. I will update the dockblocks!

library/Exceptions/NestedValidationException.php
henriquemoody: This method adds children to the `NestedValidationException`

PR #1071, opened 2018-08-14
Implement \LayerShifter\TLDDatabase

gpredl:
Because tldList is always out of date, one might use \LayerShifter\TLDDatabase.
Changed validation to check for TLDDatabase and test TLD against it.

### DIFF
--- a/library/Rules/Tld.php
+++ b/library/Rules/Tld.php
@@ -121,6 +121,12 @@ class Tld extends AbstractRule
 
     public function validate($input)
     {
-        return in_array(strtolower($input), $this->tldList);
+        // prefere LayerShifter\TLDDatabase if available
+        if (class_exists('\LayerShifter\TLDDatabase\Store')) {
+            $tldDb = new \LayerShifter\TLDDatabase\Store();
+            return $tldDb->isExists(strtolower($input));
+        } else {
+            return in_array(strtolower($input), $this->tldList);
+        }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
         "egulias/email-validator": "Strict (RFC compliant) email validation",
         "symfony/validator": "Use Symfony validator through Respect\\Validation",
         "zendframework/zend-validator": "Use Zend Framework validator through Respect\\Validation",
-        "friendsofphp/php-cs-fixer": "Fix PSR2 and other coding style issues"
+        "friendsofphp/php-cs-fixer": "Fix PSR2 and other coding style issues",
+        "layershifter/tld-database": "Database abstraction for Public Suffix List"
     },
     "autoload": {
         "psr-4": {


prisis:
How about a test that will fail if the list is outdated and a update command/bash script to update the list?

gpredl:
This is already done within TLDDatabase. I'm not sure if we should add this whole part into Respect\Validation.

prisis:
hmm then the TLDDatabase packe should go to require-dev and a suggest should be added 

like for the email-validator

gpredl:
I'm beginner to pull requests. Should I add the package to compoeres.json->require-dev and also do a pull request? 
Can I link the 2 request together? 
Not sure what to do now...

prisis:
Do it in the same pr :), just add it to the composer suggest

henriquemoody:
Thanks for contributing, that's indeed a good idea! 🐼

This pull request should be target to the branch `master` because it introduces a new behaviour.

I also checked another library which seems to be more updated than this one: https://github.com/arubacao/tld-checker

Is there a specific reason why you picked `ayershifter/tld-database`?

library/Rules/Tld.php
```diff
@@ -121,6 +121,12 @@ class Tld extends AbstractRule
 
     public function validate($input)
     {
-        return in_array(strtolower($input), $this->tldList);
+        // prefere LayerShifter\TLDDatabase if available
+        if (class_exists('\LayerShifter\TLDDatabase\Store')) {
```
henriquemoody:
```diff
-        if (class_exists('\LayerShifter\TLDDatabase\Store')) {
+        if (class_exists(Store::class)) {
```

library/Rules/Tld.php
```diff
@@ -121,6 +121,12 @@ class Tld extends AbstractRule
...
+            $tldDb = new \LayerShifter\TLDDatabase\Store();
+            return $tldDb->isExists(strtolower($input));
+        } else {
```
henriquemoody:
```diff
-        } else {
-            return in_array(strtolower($input), $this->tldList);
-        }
+        return in_array(strtolower($input), $this->tldList);
```

library/Rules/Tld.php
```diff
@@ -121,6 +121,12 @@ class Tld extends AbstractRule
 
     public function validate($input)
     {
-        return in_array(strtolower($input), $this->tldList);
+        // prefere LayerShifter\TLDDatabase if available
+        if (class_exists('\LayerShifter\TLDDatabase\Store')) {
+            $tldDb = new \LayerShifter\TLDDatabase\Store();
```
henriquemoody: Please import this class

library/Rules/Tld.php
```diff
@@ -121,6 +121,12 @@ class Tld extends AbstractRule
...
+            return $tldDb->isExists(strtolower($input));
```
henriquemoody: If we're get to use `mb_strtoupper` twice may be a good idea to have a variable for that

library/Rules/Tld.php
```diff
@@ -121,6 +121,12 @@ class Tld extends AbstractRule
 
     public function validate($input)
     {
-        return in_array(strtolower($input), $this->tldList);
+        // prefere LayerShifter\TLDDatabase if available
```
henriquemoody: Is that comment really necessary?

gpredl:
@henriquemoody Is just coincidence that I've chosen layershifter/tld-database.

library/Rules/Tld.php
gpredl: Might be easier to understand if one is not sure what's going on here

library/Rules/Tld.php
gpredl: In my opinion strtolower will only be called once, because of the if->return statement. If we use a variable for that, it will consume storage.

library/Rules/Tld.php
gpredl: I've tried to import, but couldn't get it to work if TLDDatabase is not available. Maybe I did something wrong, because I'm beginner to OOP.

henriquemoody:
In the meantime, would you like to check #1073?

henriquemoody:
I think it's better to keep updating the list rather than injecting another dependency to this repository.

I've been updating the list with the command:

```bash
curl -L curl -L https://data.iana.org/TLD/tlds-alpha-by-domain.txt |
  grep -v '^#' |
  tr '[A-Z]' '[a-z]' |
  sed -E "s,^,',; s/$/', /" |
  tr -d '\n' |
  fold -w 72 -s
```

And the last update was about 2 months ago, so I think we should be fine!

However, thank you very much for contributing! 🐼

prisis:
Maybe you can create a test that will fail if a new tld is registerd, im doing something like this in https://github.com/narrowspark/http-status

henriquemoody:
Thanks for the suggestion.

A test like this is very expensive and it doesn't test any behavior. Besides, if you are working on something completely unrelated this test may fail because the list is not updated, therefore you would have to update the list first (doesn't seem like a good workflow to me). Personally, I'd prefer to execute this command daily, perhaps I can do something like that in a cheap server somewhere, and just push that to the repository. What do you think?

PR #1073, opened 2018-08-18
Update "Tld" rule with IANA list

henriquemoody:
This commit will update the list of Top-Level Domains in the "Tld" rule
with the latest version found in the website of IANA (Internet Assigned
Numbers Authority).

The list was created with the commands below:

```bash
curl -L curl -L https://data.iana.org/TLD/tlds-alpha-by-domain.txt |
    grep -v '^#' |
    tr '[A-Z]' '[a-z]' |
    sed -E "s,^,',; s/$/', /" |
    tr -d '\n' |
    fold -w 72 -s
```


prisis:
Maybe you can add a test that will fail if the list ist not up-to-date

henriquemoody:
@prisis I think the updating of this list should be more like a process apart than something like this. 

There also other rules that need to be updated frequently. I'll create an issue for that. Thanks for stopping by! 🐼

PR #1075, opened 2018-08-19
Do not trigger PHP errors in the "Call" rule

henriquemoody:
This commit make sure that when the callable is executed by the "Call"
rule and PHP triggers an error, the user does not have to deal with it,
and instead the rule will "CallException".

Because of the many changes that were made, it didn't make sense to keep
the class "Call" extending the "AbstractRelated" class.

One thing that is a bit problematic with this rule - and with other
rules as well - is that Validation only knows details of a validation
when it fails, because of that we cannot invert the validations that
passed, meaning that the "Not" rule cannot give the proper response to
a validation that passed. This is a know issue that can only be fixed
is we provide a way for Validation do have more granularity control.

### DIFF
--- a/docs/Call.md
+++ b/docs/Call.md
@@ -2,8 +2,9 @@
 
 - `Call(callable $callable, Rule $rule)`
 
-This is a very low level validator. It calls a function, method or closure
-for the input and then validates it. Consider the following variable:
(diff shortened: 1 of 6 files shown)


henriquemoody:
@batusa thanks for your review!

PR #1076, opened 2018-08-19
Remove support "egulias/email-validator" ~1.0

henriquemoody:
The rule "Email" now only supports the version >=2.0 of
"egulias/email-validator". Also a few changes were made to make the rule
a bit simpler.

The method "getEmailValidator()" which was a lazy load of the
"EmailValidator" class was removed in favor of trying to create the
instance of "EmailValidator" in the constructor of "Email". That makes
the rule immutable after its creation.

### DIFF
--- a/library/Rules/Email.php
+++ b/library/Rules/Email.php
@@ -30,28 +30,20 @@ use function is_string;
 final class Email extends AbstractRule
 {
     /**
-     * Assigns email validator.
-     *
-     * @param EmailValidator $emailValidator
+     * @var EmailValidator
      */
-    public function __construct(EmailValidator $emailValidator = null)
-    {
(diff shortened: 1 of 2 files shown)


library/Rules/Email.php
```diff
@@ -30,44 +30,44 @@
...
-        $this->emailValidator = $emailValidator;
-    }
+    private $validator;
 
     /**
-     * Gets email validator.
+     * Initializes the rule assigning the EmailValidator instance.
      *
-     * @return EmailValidator | null
+     * If the EmailValidator instance of not defined, tries to create one.
```
batusa:
```diff
- If the EmailValidator instance of not defined, tries to create one.
+ If the EmailValidator instance is not defined, tries to create one.
```

PR #1077, opened 2018-08-19
Change how "Sf" rule works

henriquemoody:
Instead of creating the Symfony constraints itself "Sf" accepts an
instance of "Symfony\Component\Validator\Constraint".

Creating objects inside a rule, specially from an external library,
makes the rule too complex and also limits the possibilities with the
"Sf" rule since Symfony allows users to create complex validations (even
thought their API is not as simple as ours).

This commit also simplifies the way the messages are passed from Symfony
to the "Sf" when only one constraint has failed; instead of passing
the message of the whole constraint violation list, only the fist
constraint violation message it passed.

The problem that this rule will always have is that when using "Not" to
invert the validation we have a way to get a proper message since
Symfony Validator only return the result of constraints that failed.
That's something the Respect\Validation does in a similar way and to
change it a lot has to be changed.

These changes were checked in "symfony/validator" 4.0 and the version
was added to the "composer.json" file.

### DIFF
--- a/docs/Sf.md
+++ b/docs/Sf.md
@@ -1,19 +1,24 @@
 # Sf
 
-- `Sf(string $validatorName)`
+- `Sf(Constraint $constraint)`
+- `Sf(Constraint $constraint, ValidatorInterface $validator)`
 
-Use Symfony2 validators inside Respect\Validation flow. Messages
-are preserved.
+Validate the input with a Symfony Validator (>=4.0 or >=3.0) Constraint.
 
 ```php
-v::sf('Time')->validate('15:00:00');
+use Symfony\Component\Validator\Constraint\Iban;
+
+v::sf(new Iban())->validate('NL39 RABO 0300 0652 64'); // true
 ```
+
+This rule will keep all the messages returned from Symfony.
 
 ## Changelog
 
 Version | Description
 --------|-------------
-  2.0.0 | Upgraded support to version >=3.0.0 of Symfony Validator
+  2.0.0 | Do not create constraints anymore
+  2.0.0 | Upgraded support to version >=4.0 or >=3.0 of Symfony Validator
   0.3.9 | Created
 
 ***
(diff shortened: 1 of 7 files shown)


docs/Sf.md
```diff
@@ -1,19 +1,23 @@
 # Sf
 
-- `Sf(string $validatorName)`
+- `Sf(Constraint $constraint)`
+- `Sf(Constraint $constraint, ValidatorInterface $validator)`
 
-Use Symfony2 validators inside Respect\Validation flow. Messages
-are preserved.
+Validate the input with a Symfony Validator Constraint.
```
batusa: I think it's better to specify from what Symfony version it works.

tests/unit/Rules/SfTest.php
```diff
@@ -14,84 +14,103 @@
...
-        self::assertFalse(
-            v::sf($constraintName)->validate($invalidConstraintValue),
-            sprintf('"%s" should be invalid under "%s" constraint.', $invalidConstraintValue, $constraintName)
-        );
+        $constraint = new IsNull();
+        $validator = Validation::createValidator();
+
+        $sut = new Sf($constraint, $validator);
+
+        self::assertTrue($sut->validate(null));
```
batusa: how can we assert that the Validator provided is a custom one? here it only validates the result.

tests/unit/Rules/SfTest.php
```diff
@@ -14,84 +14,103 @@
...
     {
-        $constraintName = 'Time';
-        $validConstraintValue = '04:20:00';
-        v::sf($constraintName)->assert($validConstraintValue);
+        $constraint = new IsFalse();
+        $validator = Validation::createValidator();
+
+        $sut = new Sf($constraint, $validator);
+
+        self::assertFalse($sut->validate(true));
```
batusa: same here

tests/unit/Rules/SfTest.php
```diff
@@ -14,84 +14,103 @@
...
+
+        self::assertAttributeInstanceOf(ValidatorInterface::class, 'validator', $sut);
+    }
+
+    /**
+     * @test
+     */
+    public function itShouldPassSymfonyMessageToSfException(): void
+    {
+        $sut = new Sf(new Collection([
```
batusa: I'd pull such test to the integration test suite instead.

PR #1079, opened 2018-08-21
Intercept throwables in the "Call" rule

henriquemoody:
The callable defined to the "Call" rule may also throw an exception and
as we don't want to have errors nor exceptions that are not part of the
Validation during the validation of inputs it just makes sense to
intercept any instance of Throwable.

This change was initially thought of because in Travis the version 7.4
of PHP was throwing "Error" instead of triggering PHP errors which made
the tests fail.

Reference: 2fac861aa102e7274dfbd24fce23668a0dc5d97e

### DIFF
--- a/library/Rules/Call.php
+++ b/library/Rules/Call.php
@@ -15,6 +15,7 @@ namespace Respect\Validation\Rules;
 
 use Respect\Validation\Exceptions\ValidationException;
 use Respect\Validation\Validatable;
+use Throwable;
 use function call_user_func;
 use function restore_error_handler;
 use function set_error_handler;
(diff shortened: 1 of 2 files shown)


library/Rules/Call.php
```diff
@@ -57,7 +58,11 @@ public function assert($input): void
     {
         $this->setErrorHandler($input);
 
-        $this->rule->assert(call_user_func($this->callable, $input));
+        try {
+            $this->rule->assert(call_user_func($this->callable, $input));
+        } catch (Error $error) {
```
batusa: Maybe better expect for `Throwable` instead?

PR #1085, opened 2018-08-29
Fix wrong behavior of "Date" rule

### LINKED ISSUE #1083: Date validation returns different results whether providing a format
Hi!
As you can see in the following code: 

```php
v::date()->validate('2014-99'); //false
v::date('Y-m')->validate('2014-99'); //true
```
The date validator with a format returns `true` with an invalid date. It's due [the use of `date_parse_from_format`](https://github.com/Respect/Validation/blob/1.1/library/Rules/Date.php#L52) that only parse the string according to a format but without check whether the date is valid or not. Why not use [`strtotime`](https://github.com/Respect/Validation/blob/1.1/library/Rules/Date.php#L40) function in that case too?

henriquemoody:
The validation was considering "99" as a valid month and neither
date_parse_from_format() [1] or DateTime::createFromFormat() [2] would
complain about that.

The solution for that was to use checkdate() [3] to verify whether the
date is valid or not. Also, an extra step was added to the validation
that a date that only contains a day would return false since a day
without a month and year is impossible to validate.

Apparently, there is no problem while validation when it comes to time,
therefore nothing needed to be added for this validation.

[1]: http://php.net/date_parse_from_format
[2]: http://php.net/datetime.createFromFormat
[3]: http://php.net/checkdate

***
Fix #1083

### DIFF
--- a/library/Rules/Date.php
+++ b/library/Rules/Date.php
@@ -49,8 +49,39 @@ class Date extends AbstractRule
             $this->format = $exceptionalFormats[$this->format];
         }
 
-        $info = date_parse_from_format($this->format, $inputString);
+        return $this->isValidForFormatProvided($input);
+    }
 
+    private function isValidForFormatProvided($input)
+    {
+        $info = date_parse_from_format($this->format, $input);
+        if (!$this->isParsable($info)) {
+            return false;
+        }
+
+        if ($this->hasDateFormat()) {
+            return $this->hasValidDate($info);
+        }
+
+        return true;
+    }
+
+    private function isParsable(array $info)
+    {
         return ($info['error_count'] === 0 && $info['warning_count'] === 0);
+    }
+
+    private function hasDateFormat()
+    {
+        return preg_match('/[djSFmMnYy]/', $this->format) > 0;
+    }
+
+    private function hasValidDate(array $info)
+    {
+        if ($info['day']) {
+            return checkdate((int) $info['month'], $info['day'], (int) $info['year']);
+        }
+
+        return checkdate($info['month'] ?: 1, $info['day'] ?: 1, $info['year'] ?: 1);
     }
 }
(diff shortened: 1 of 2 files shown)


library/Rules/Date.php
```diff
@@ -49,8 +49,39 @@ public function validate($input)
             $this->format = $exceptionalFormats[$this->format];
         }
 
-        $info = date_parse_from_format($this->format, $inputString);
+        return $this->isValidDateTime($input);
```
batusa: I'd do a few more changes in this code.

- The code from 43 to 50 I'd extract to a private `fixFormatForExceptionalCases()` method.
- I'd rename 'isValidDateTime' to 'isValidForFormatProvided' and I'd use `fixFormatForExceptionalCases` inside of it.

library/Rules/Date.php
```diff
@@ -49,8 +49,39 @@ public function validate($input)
...
+    }
+
+    private function isValidDateTime($input)
+    {
+        $info = date_parse_from_format($this->format, $input);
+        if (!$this->isParsable($info)) {
+            return false;
+        }
+
+        if (!$this->hasDateFormat()) {
```
batusa: If it doesn't have a Date format, is it `true`?

library/Rules/Date.php
henriquemoody: I don't wanna change much because on `master` this rule is completely different

library/Rules/Date.php
henriquemoody: It means that I don't need to validate the date

library/Rules/Date.php
henriquemoody: I renamed `isValidDateTime` to `isValidForFormatProvided`, the `fixFormatForExceptionalCases` shouldn't even be there because we shouldn't change the state of the rule after the validation but we're doing it here, see: https://github.com/Respect/Validation/blob/master/library/Helpers/DateTimeHelper.php#L31-L41

library/Rules/Date.php
henriquemoody: I changed the order of the conditions to make it more clear

Issue #847, opened 2017-08-09
[Idea] Error message handling


Iazel:
Hello!
I really like this library, however it looks like error messages are a serious issue for lots of people; after a quick look at issues, many of them regards error messages and someone don't want to use this awesome validator just for that ^^"

So, I would like to discuss with all the community if the follow solution (for which I can do a PR) would be acceptable or not.

Instead of using Exception, I think would be better to return the actual errors, but given that I don't want to break current contract, I lean more on introducing a new method on AbstractValidator, something like:
```
public function getErrorsFor($input): ?array
```
As the name suggest, it should return an array of errors, possibly scoped by either name or input value.

## Example
```
$errors = v::key('name', v::stringType()->notEmpty()->length(2, 40))
           ->key('age', v::intType()->between(18, 30))
           ->key('address', v::keySet([
              'street' => $this->stringType()->notEmpty(),
              'number' => $this->stringType()->notEmpty(),
           ])
           ->getErrorsFor([
               'name' => '',
               'age' => 10,
           ]);
```
The return value could be something like:
```
[
    'name' => [
        ['rule' => 'notEmpty'],
        ['rule' => 'length', 'extra' => ['min' => 2]],
    ],
    'age' => [
        ['rule' => 'between', 'extra' => ['min' => 18, 'max' => 30]]
    ],
    'address' => [
        'street' => ['rule' => 'notEmpty']
        'number' => ['rule' => 'notEmpty']
    ]
]
```
Given this amount of information, it would be rather easy to make any kind of error message. It is particularly handy when we want to send this info to frontend and render the real message there :)

### Normal Rules
Validating a normal rule should return just one array with keys:
* rule: rule's type, like length, between, etc... Always present.
* ruleName: present only if a name is set through `setName`
* extra: an optional array with extra information useful for constructing the error message, eg: `min`

### Composite Rules
Every rules that extends `AbstractComposite` should aggregate all errors returned by each children rules. The result should be an array indexed by child name.
It should support multiple nesting levels.

thibaudcolas:
Here are specific issues that make this library very hard to use when you want to display useful error messages to end users:

- #628 `keySet` throws an error if a key is missing, but doesn't mention which key it is.
- #769 `allOf` (and `each` as well I think) throw their own errors "All of the required rules must pass" when more than one of their sub-validators throw an error. Those extra messages don't add any value to whoever looks at the message.
- #924 default error messages do not provide much context

[...]

PR #1087, opened 2018-09-03
Rename method "validate()" to "isValid()"

henriquemoody:
Semantically it makes more sense that the method that returns whether
the input is valid or not for a specific rule be called "isValid()"
than "validate()". Also, in many languages, it's very common to prefix
methods that return boolean values in response to some logic with "is",
"can", "has", etc.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,7 @@ create a validator that validates if a string is equal to "Hello World".
 
 The rule itself needs to implement the `Validatable` interface but, it is
 convenient to just extend the `AbstractRule` class.
-Doing that, you'll only need to declare one method: `validate($input)`.
+Doing that, you'll only need to declare one method: `isValid($input)`.
 This method must return `true` or `false`.
 
 If your validator class is `HelloWorld`, it will be available as `v::helloWorld()`
@@ -77,7 +77,7 @@ final class HelloWorld extends AbstractRule
     /**
      * {@inheritdoc}
      */
-    public function validate($input): bool
+    public function isValid($input): bool
     {
         return $input === 'Hello World';
     }
@@ -137,8 +137,8 @@ are able to use any methods of it. By extending `RuleTestCase` you should
 implement two methods that should return a [data provider][] with the rule as
 first item of the arrays:
 
-- `providerForValidInput`: Will test when `validate()` should return `true`
-- `providerForInvalidInput`: Will test when `validate()` should return `false`
+- `providerForValidInput`: Will test when `isValid()` should return `true`
+- `providerForInvalidInput`: Will test when `isValid()` should return `false`
 
 ```php
 <?php
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 [The most awesome validation engine ever created for PHP.](http://bit.ly/1a1oeQv)
 
-- Complex rules made simple: `v::numericVal()->positive()->between(1, 255)->validate($input)`.
+- Complex rules made simple: `v::numericVal()->positive()->between(1, 255)->isValid($input)`.
 - Granularity control for advanced reporting.
 - More than 130 (fully tested) validation rules.
 - A concrete API for non fluent usage.
--- a/docs/concrete-api.md
+++ b/docs/concrete-api.md
@@ -8,7 +8,7 @@ or magic methods. We'll use a traditional dependency injection approach.
 use Respect\Validation\Validator as v;
 
 $usernameValidator = v::alnum()->noWhitespace()->length(1,15);
-$usernameValidator->validate('alganet'); // true
+$usernameValidator->isValid('alganet'); // true
 ```
 
 If you `var_dump($usernameValidator)`, you'll see a composite of objects with
@@ -24,7 +24,7 @@ $usernameValidator = new Rules\AllOf(
     new Rules\NoWhitespace(),
     new Rules\Length(1, 15)
 );
-$usernameValidator->validate('alganet'); // true
+$usernameValidator->isValid('alganet'); // true
 ```
 
 This is still a very lean API. You can use it in any dependency injection
@@ -39,7 +39,7 @@ $usernameValidator = new Rules\AllOf(
     new Rules\Length(1, 15)
 );
 $userValidator = new Rules\Key('name', $usernameValidator);
-$userValidator->validate(['name' => 'alganet']); // true
+$userValidator->isValid(['name' => 'alganet']); // true
 ```
 
 ## How It Works?
--- a/docs/feature-guide.md
+++ b/docs/feature-guide.md
@@ -15,7 +15,7 @@ The Hello World validator is something like this:
 
 ```php
 $number = 123;
-v::numericVal()->validate($number); // true
+v::numericVal()->isValid($number); // true
 ```
 
 ## Chained validation
@@ -25,7 +25,7 @@ containing numbers and letters, no whitespace and length between 1 and 15.
 
 ```php
 $usernameValidator = v::alnum()->noWhitespace()->length(1, 15);
-$usernameValidator->validate('alganet'); // true
+$usernameValidator->isValid('alganet'); // true
 ```
 
 ## Validating object attributes
@@ -44,7 +44,7 @@ Is possible to validate its attributes in a single chain:
 $userValidator = v::attribute('name', v::stringType()->length(1,32))
                   ->attribute('birthdate', v::dateTime()->age(18));
 
-$userValidator->validate($user); // true
+$userValidator->isValid($user); // true
 ```
 
 Validating array keys is also possible using `v::key()`
@@ -78,7 +78,7 @@ v::key(
         ->key('field2', v::stringType())
         ->key('field3', v::boolType())
     )
-    ->assert($data); // You can also use check() or validate()
+    ->assert($data); // You can also use check() or isValid()
 ```
 
 ## Input optional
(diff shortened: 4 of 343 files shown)


henriquemoody:
I agree with @augustohp in a lot of points. I'm a bit afraid of breaking BC because of user adoption. We already broke BC big time on the next version 2.0, changing also this now may not be very wise, however, move to a better direction would be nice.

I've already talked to @batusa about the method names, in fact, my desire was to:
- Rename `validate()` to `isValid()`;
- Rename `assert()` to `assertAll()`;
- Rename `check()` to `assert()`;
- Create a `validate()` method that returns a `Validation` object (see https://github.com/Respect/Validation/issues/847#issuecomment-416336925)

This PR was just the beginning of the changes.

> 1 - Leave check, assert and validate as is

I'm +1 for that

> 2 - Choose something similar to `isValid` for `assert` and `check`

I'm +1 for that as well

> 3 - Expose an adapter class to be used instead of `Validator` with that behavior

I'm -0 for that because I'm not sure this is practical or doable because the `Validator` is also a rule that must implement `Validatable` methods.

> 4 - Silently add `isValid` as an alias to validate without breaking BC

I'm +1 but also need we need something more (see below).

> 5 - Accept current pull request state

I'm also -1 for that one.

After reading @augustohp's input I thought about a possible 6th solution:

**Create alias for `validate()`, `assert()`, and `check()`**
Which we would:

1) Define these new names together;
2) Implement these alias in the codebase
    - Add the `@deprecated` annotation to the old ones;
    - Make the new method leading the implementations (questionable if that's doable or not though)
3) Update the documentation to point to the **new** methods.

That would mean that in the new MAJOR version we would drop the old methods.

I didn't mention anything about https://github.com/Respect/Validation/issues/847#issuecomment-416336925 in this proposal because I think it's a different problem.

Was that something that would make sense for y'all?

augustohp:
# TLDR

My view is that our strongest option, respecting PHP no-bc break policy and our 4k installs today, is [3 - The Adapter class][3]. I exposed my view in this comment by referring to it as the a new `Builder` class.

I don't see any reason, other than preference, to break compatibility with previous versions. As I don't see why adding behavior to a new Builder class doesn't solve the proposed problem.

---

>> 3 - Expose an adapter class to be used instead of Validator with that behavior
> 
> I'm -0 for that because I'm not sure this is practical or doable because the Validator is also a rule that must implement Validatable methods. 
> [^1][1]

[1]: https://github.com/Respect/Validation/pull/1087#issuecomment-418503101

I was thinking something very silly as:

```php
namespace Respect\Validation;

class Builder extends Validator
{
    // ...
}
```

That builder would [decorate][d] all created rules with `isValid`-like methods. I don't like doing that in the current class as I don't see the *practical benefits of that change*.

[d]: https://sourcemaking.com/design_patterns/decorator

---

> After reading @augustohp's input I thought about a possible 6th solution:
> 
> Create aliases for `validate()`, `assert()`, and `check()`
> 
> In which we could:
> 
> 1. Define these new names together;
> [^1][1]

I am a +0 for that.

> 2. Implement these aliases in the codebase
>    * Add the `@deprecated` annotation to the old ones;
>    * Make the new method leading the implementations (questionable if that's doable or not though)
> [^1][1]

I am a -1 for `@deprecated`. I think we must, first, agree upon:

* **Keep current methods** as they are today without any additional functionality or new methods, meaning we just drop this and learn from this discussion
* **Add a new API** by increasing available methods without changing previous functional behavior
* **Change the current API** by using `@deprecated` and then removing existing methods

I am only a -1 for **Changing the current API**.

> 3. Update the documentation to point to the new methods.
> [^1][1]

If I managed to make myself clear in the previous statement, I am a -1 for this because of changing current API.

---

> That would mean that in the new MAJOR version we would drop the old methods.
> 
> I didn't mention anything about #847 (comment) in this proposal because I think it's a different problem.

I agree, but I think they can be solved using the same strategy. Why? Not *breaking backward compatibility* means not harming people who already use the library. By *harming* we can mean many things:

* Making them go through *release notes*
* Requiring them to read documentation they don't intend to
* Forcing people to actually understand *semantic versioning* and not point to `dev-master` - which causes bigger problems like stability
* Creating bugs instead of fixing them just because an API they rely on, changed - and they have some or all of the issues above

I think we can fix both issues by providing an explicit `Builder` with an specific vocabulary and more complex behavior. As [Slim-Validation][1] by @DavidePastore did, as mentioned on #847, by adding behavior without affecting anyone - which I didn't look, just trusting the comments. We already have a Factory which could do that, we just need to offer a different use for it.

[sv]: https://github.com/DavidePastore/Slim-Validation
[3]: https://github.com/Respect/Validation/pull/1087#pullrequestreview-151908266

henriquemoody:
I'm gonna close this PR for now because there are some things more important than that at the moment.

My point on renaming the methods is to make the library more consistent and providing a way to perform a validation without throwing an exception but also giving some context for the users (and I think the `validate()` is perfect for that).

That said if I were to create a Decorator for the current `Validator` class, without a plan to deprecate the old methods and eventually remove them I won't do it because it's more code to maintain and I'd rather avoid it as much as possible.

About the BC breaks, the version that is on master breaks it like no tomorrow and it's on my plans to create a migration document. From the top of my head I can quote:

- All rules are `final`
- All rule properties are `private`
- The `NestedValidationException::findMessages()` was removed
- `ValidationException` had methods renamed, removed, and modified (`__construct` included)
- `Alnum`, `Alpha`, and `Digit` don’t accept whitespaces anymore
- `Date` accepts only dates (not date and time)
- Bank-related rules were removed
- `Each` doesn’t validate keys anymore

For each of these changes, there is a reason (a good reason in my opinion), and I can admit that it will be hard for some users indeed.

Thanks for your comments, especially you @augustohp for pointing out your concerns. I hope we can have a chat about it later - not here. Even thought you're very clear, I'd like to understand your thoughts a little bit better.

Again, thanks everyone! 🐼

augustohp:
> Thanks for your comments, especially you @augustohp for pointing out your concerns. I hope we can have a chat about it later - not here. Even thought you're very clear, I'd like to understand your thoughts a little bit better.

I'd rather have these conversations openly. My only point is to avoid braking BC at all costs, which I understand is no longer an argument.

I rest my case. But I am sad.

henriquemoody:
@augustohp,

I'm sorry that this is making you sad. You still play a huge part in the success of this library and I'd rather have you happy about how the direction of this library.

I didn't say _more code is bad_, simply put more code is more code to maintenance and I'd like to _avoid_ it. In this case, if the intention is to be more consistent, adding another layer and don't even plan to get rid of the inconsistent one and not push the users to use the new one doesn't seem worth it.

In one thing for me is clear, we extremely disagree on BC break policy. I believe that maintainers of any library should have the freedom to change API and break backward compatibility as long as they respect the semantic versioning and I could quote lots of libraries and languages that have done that multiple times. I don't think breaking BC is disrespectful, and it certainly _can_ be depending on how it's done. There are many other ways to disrespect the users, try to improve the library is not one of them.

When I read your message I felt very misjudged because since I've been maintaining this library for years I think I should have the benefit of the doubt that my reasons are good.

And when I said I wanted to talk to you _not here_ was only because I strongly believe in as close the communication is the better that is. Usually threads grown bigger because the loop is too long. That wouldn't mean that the outcome wouldn't be public. Why would you rather have this conversation here?

augustohp:
> I'm sorry that this is making you sad. You still play a huge part in the success of this library and I'd rather have you happy about how the direction of this library.

I am sorry to be blunt here, but I think that ship is long gone. Don't get me wrong here, I am not pointing fingers or blame your for something; I am just being critic and honest: I disagree with most changes being made - I fail to see the value they add to end users.

> I didn't say _more code is bad_, simply put more code is more code to maintenance and I'd like to _avoid_ it. In this case, if the intention is to be more consistent, adding another layer and don't even plan to get rid of the inconsistent one and not push the users to use the new one doesn't seem worth it.

A [previous table of score][1] explains why I can't relate to that rationale: How come *breaking BC* is better than *more code* to have a feature?

I am probably just being completely stupid here, and I am not joking, but I really can't understand how seriously one can take BC breaks while breaking them that easily.

[1]: https://github.com/Respect/Validation/pull/1087#pullrequestreview-151908266

> In one thing for me is clear, we extremely disagree on BC break policy. I believe that maintainers of any library should have the freedom to change API and break backward compatibility as long as they respect the semantic versioning and I could quote lots of libraries and languages that have done that multiple times. I don't think breaking BC is disrespectful, and it certainly _can_ be depending on how it's done. There are many other ways to disrespect the users, try to improve the library is not one of them.

Most people don't understand SemVer. I am sorry to be the messenger here, but there are plenty of examples out there. Breaking BC is **the example** of what not to do when we deal with end users and want to keep adoption increasing (e.g: Python, Perl, etc). In the PHP field, the versioning mess of Guzzle was not made better by SemVer.

I feel people mistake *improvements* for *BC breaks*. An API should have reason to change just for the improvement of its own API, to solve an issue. Which, I may fail to see, is not the reason here. I feel we are changing the API just for the sake of changing it; I can't see the benefits of it other than "I like this API better".

> When I read your message I felt very misjudged because since I've been maintaining this library for years I think I should have the benefit of the doubt that my reasons are good. Whether they are enough to break the BC is open for discussion.

You have the merits of it, you don't have the benefit of doubt as I still fail to understand the *consistency* argument.

> And when I said I wanted to talk to you _not here_ was only because I strongly believe in as close the communication is the better that is. Usually threads grown bigger because the loop is too long. That wouldn't mean that the outcome wouldn't be public. Why would you rather have this conversation here?

I agree with your reasoning but OSS is **open**. Closing discussions doesn't favor collaboration and are more prone to suffer from *Survivorship Bias**, which I think is the case here.

---

## TLDR

I am not trying to mine you efforts in keeping the library, still: I don't understand you *consistency* argument to break an important BC like that.

I don't like the direction of the library for it feels we are treating developers like morons. We look like Symfony now, which is a shame based on Respect's values - and I think you should have expected me to comment those stuff out loud.

But again: don't feel discouraged by it. Keep the good work maintaining the library just don't expect my approval on those changes - my reasons are kept exposed and I feel like they are waiting to be contradicted.

PR #1088, opened 2018-09-03
 Apply contribution guidelines to 'Length' rule

danilosilva87:
I'm only having a doubt in the integration tests.

In the `Length` rule there are 4 files (length_1, length_2, length_3, length_4), but test number 1 was like this:
```php

$validator = v::length(0, 10);

$validator->validate('phpsp');
v::not($validator)->validate('phpsp');
$validator->assert('nickolas');
$validator->check('nawarian');
```
I removed this test, was it necessary?


library/Exceptions/LengthException.php
```diff
@@ -13,13 +13,21 @@
...
     public const LOWER = 'lower';
     public const GREATER = 'greater';
     public const EXACT = 'exact';
 
+    /**
+     * {@inheritdoc}
+     */
     public static $defaultTemplates = [
         self::MODE_DEFAULT => [
             self::BOTH => '{{name}} must have a length between {{minValue}} and {{maxValue}}',
```
henriquemoody: There is also another method in this class (`chooseTemplate()`) that needs the `{@inheritdoc}` annotation.

library/Rules/Length.php
```diff
@@ -46,6 +66,9 @@ public function __construct($min = null, $max = null, $inclusive = true)
         }
     }
 
+    /**
+     * {@inheritdoc}
+     */
     public function validate($input): bool
```
henriquemoody: There are some methods below that are protected, since the class is final now it doesn't make sense for them to be protected anymore.

tests/integration/rules/length.phpt
```diff
@@ -24,8 +25,21 @@ try {
...
+}
 ?>
 --EXPECTF--
 "nawarian" must have a length lower than 5
 "phpsp.org.br" must have a length greater than 13
-"phpsp.org.br" must not have a length between 5 and 20
\ No newline at end of file
+"phpsp.org.br" must not have a length between 5 and 20
+"123456" must have a length of 5
+- "phpsp.org.br" must have a length between 3 and 5
```
henriquemoody: Missing EOF

tests/integration/rules/length.phpt
```diff
@@ -24,8 +25,21 @@ try {
     echo $e->getMessage().PHP_EOL;
 }
 
+try {
+    v::length(5, 5)->check('123456');
+} catch (LengthException $e) {
```
henriquemoody: I'd rather not abbreviate variables, in this case, it would be `$exception` instead.

tests/integration/rules/length.phpt
```diff
@@ -24,8 +25,21 @@ try {
...
+    echo $e->getMessage().PHP_EOL;
+}
+
+try {
+    v::length(3, 5)->assert('phpsp.org.br');
+} catch (AllOfException $e) {
```
henriquemoody: The new tests are using `NestedValidationException` instead

tests/integration/rules/length.phpt
```diff
@@ -24,8 +25,21 @@ try {
...
+    echo $e->getFullMessage();
+}
 ?>
 --EXPECTF--
 "nawarian" must have a length lower than 5
 "phpsp.org.br" must have a length greater than 13
-"phpsp.org.br" must not have a length between 5 and 20
\ No newline at end of file
+"phpsp.org.br" must not have a length between 5 and 20
```
henriquemoody: The exception has 6 possible messages, we should test all of them with `check()` and `assert()`.

tests/unit/Rules/LengthTest.php
```diff
@@ -13,120 +13,57 @@
 
 namespace Respect\Validation\Rules;
 
-use PHPUnit\Framework\TestCase;
+use Respect\Validation\Test\RuleTestCase;
 
 /**
  * @group  rule
```
henriquemoody:
```diff
- * @group  rule
+ * @group rule
```

tests/unit/Rules/LengthTest.php
```diff
@@ -13,120 +13,57 @@
...
+ *
  * @covers \Respect\Validation\Exceptions\LengthException
```
henriquemoody: This test doesn't cover `LengthException`, that's why we have the integration tests :)

library/Rules/Length.php
```diff
@@ -15,13 +15,33 @@
 
 use Respect\Validation\Exceptions\ComponentException;
 
-class Length extends AbstractRule
+/**
+ * @author Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
```
henriquemoody: Would be nice to add a one-line documentation in this docblock (which can be found in documentation of the rule)

library/Rules/Length.php
```diff
@@ -15,13 +15,33 @@
 
 use Respect\Validation\Exceptions\ComponentException;
 
-class Length extends AbstractRule
```
henriquemoody: You would also import the functions that were used in this class.

PR #1092, opened 2018-09-05
Apply contribution guidelines to 'NumericVal' rule

danilosilva87:


library/Rules/NumericVal.php
```diff
@@ -13,8 +13,16 @@
 
 namespace Respect\Validation\Rules;
 
-class NumericVal extends AbstractRule
+/**
+ * @author Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
```
henriquemoody: Would be nice to have a single-line documentation here (which can be found in the documentation file of this rule).

library/Rules/NumericVal.php
```diff
@@ -13,8 +13,16 @@
 
 namespace Respect\Validation\Rules;
 
-class NumericVal extends AbstractRule
```
henriquemoody: Missing `use function is_numeric;`

tests/integration/rules/numeric.phpt
```diff
@@ -0,0 +1,51 @@
+--FILE--
```
henriquemoody: This file should be called `numericVal.phpt`

tests/integration/rules/numeric.phpt
```diff
@@ -0,0 +1,51 @@
...
+<?php
+require 'vendor/autoload.php';
+
+use Respect\Validation\Exceptions\AllOfException;
+use Respect\Validation\Exceptions\NumericValException;
+use Respect\Validation\Validator as v;
+
+try {
+    v::numericVal()->check('a');
+} catch (NumericValException $e) {
```
henriquemoody: I'd rather not abbreviate variables, so I'd call it `$exception`

tests/integration/rules/numeric.phpt
```diff
@@ -0,0 +1,51 @@
+--FILE--
+<?php
+require 'vendor/autoload.php';
+
+use Respect\Validation\Exceptions\AllOfException;
```
henriquemoody: The newer tests are using `NestedValidationException` instead

tests/integration/rules/numeric.phpt
```diff
@@ -0,0 +1,51 @@
...
+    echo $e->getFullMessage().PHP_EOL;
+}
+?>
+--EXPECTF--
+"a" must be numeric
+- "a" must be numeric
+Field must be numeric
+- Field must be numeric
+"1" must not be numeric
+- "1" must not be numeric
```
henriquemoody: Missing EOF

tests/integration/rules/numeric.phpt
```diff
@@ -0,0 +1,51 @@
...
+
+try {
+    v::not(v::numericVal())->assert('1');
+} catch (AllOfException $e) {
+    echo $e->getFullMessage().PHP_EOL;
+}
+?>
+--EXPECTF--
+"a" must be numeric
+- "a" must be numeric
```
henriquemoody: The newer tests are testing:

1. `v::rule()->check($input)`
2. `v::not(v::rule())->check($input)`
3. `v::rule()->assert($input)`
4. `v::not(v::rule())->assert($input)`

These 4 in that order should be enough.

tests/unit/Rules/NumericValTest.php
```diff
@@ -13,66 +13,52 @@
 
 namespace Respect\Validation\Rules;
 
-use PHPUnit\Framework\TestCase;
+use Respect\Validation\Test\RuleTestCase;
 
 /**
  * @group  rule
```
henriquemoody:
```diff
- * @group  rule
+ * @group rule
```

tests/unit/Rules/NumericValTest.php
```diff
@@ -13,66 +13,52 @@
...
+ *
  * @covers \Respect\Validation\Exceptions\NumericValException
```
henriquemoody: This class is not testing `NumericValException` anymore, that's why we have the integration tests

PR #1093, opened 2018-09-06
Apply contribution guidelines to 'StringVal' rule

danilosilva87:


library/Rules/StringVal.php
```diff
@@ -13,8 +13,15 @@
 
 namespace Respect\Validation\Rules;
 
-class StringVal extends AbstractRule
+/**
+ * @author Danilo Correa <danilosilva87@gmail.com>
```
henriquemoody: Would be nice to have a single-line documentation in this docblock (you can find it in the documentation of the rule - the Markdown file)

library/Rules/StringVal.php
```diff
@@ -13,8 +13,15 @@
 
 namespace Respect\Validation\Rules;
 
-class StringVal extends AbstractRule
```
henriquemoody: Missing
```php
use function is_object;
use function is_scalar;
use function method_exists;
```

tests/unit/Rules/StringValTest.php
```diff
@@ -18,10 +18,17 @@
 
 /**
  * @group  rule
```
henriquemoody:
```diff
- * @group  rule
+ * @group rule
```

tests/unit/Rules/StringValTest.php
```diff
@@ -18,10 +18,17 @@
```
henriquemoody: Missing `use function tmpfile;`

PR #1095, opened 2018-09-06
Do not accept whitespace by default in "Alpha" rule

henriquemoody:
The "Alpha" rule is supposed to validate alphabetic values, but instead,
it also validates any whitespace character as valid by default.

The rule also accepts a list of characters on its constructor, so if the
users intentionally want some specific characters to also be allowed,
like any whitespace character, it is better than they defined these
characters on the rule's constructor. That said it would be better to
not allow any whitespaces by default.

While refactoring the rule I could notice that "AbstractCtypeRule" is
just an overhead that does not add much to it, so instead of extending
it "Alpha" now extends "AbstractFilterRule" directly (which is the
parent of "AbstractCtypeRule").

And since we want all rules to follow our contribution guidelines, this
commit also make sure the "Alpha" rule is in accordance with that.


henriquemoody:
Parked until #1096 is finished

PR #1102, opened 2018-09-09
Apply contribution guidelines to "Regex" rule

danilosilva87:


library/Rules/Regex.php
```diff
@@ -13,15 +13,36 @@
...
+ * @author Alexandre Gomes Gaigalas <alexandre@gaigalas.net>
+ * @author Danilo Correa <danilosilva87@gmail.com>
+ * @author Henrique Moody <henriquemoody@gmail.com>
+ */
 class Regex extends AbstractRule
 {
+    /**
+     * @var string
+     */
     public $regex;
```
batusa: I'd rather make this property `private`. I saw it's been used by two tests from other rules, by checking them I didn't find so much value in them. So I'd drop them and make this property `private` anyway. The tests are [NoTest](https://github.com/Respect/Validation/blob/master/tests/unit/Rules/NoTest.php#L28-L55) and [PostalCodeTest](https://github.com/Respect/Validation/blob/master/tests/unit/Rules/PostalCodeTest.php#L28-L51).

What do you think @henriquemoody ?

henriquemoody:
That would be great, @batusa!

PR #1105, opened 2018-09-11
Do not accept whitespace by default in "Alpha" rule

henriquemoody:
The intent of the "Alpha" rule is to validate alphabetic values.
However, it also considers any whitespace character (by default). That
causes some confusion, and unless you check its code or tests, you would
never expect that behavior.

Because of that confusion, I decided to make "Alpha" to not consider
whitespace characters as valid, and since in the constructor of this
rule it's possible to add extra characters to the validation it makes
sense to let the user decide whether they want whitespaces, tabs, new
lines, etc. or not.

This rule, as the same as "Alnum" previously, extends
"AbstractCtypeRule" pretty much to only make it easier to consider any
whitespaces as valid, therefore I saw no reason to keep extending it.
Now "Alpha" extends the "AbstractFilterRule" which is the parent of
"AbstractCtypeRule".

I also took the opportunity to apply our contribution guidelines to
"Alpha" since we want to apply that to all the rules.


library/Rules/Alpha.php
```diff
@@ -13,14 +13,21 @@
 
 namespace Respect\Validation\Rules;
 
-class Alpha extends AbstractCtypeRule
```
batusa: 👍

PR #1106, opened 2018-09-11
Do not accept whitespace by default in "Digit" rule

henriquemoody:
The "Digit" rule is meant to validate digits. However, by default, it
also considers any whitespace character (spaces, new lines, tabs, etc)
as valid.

Since the rule also accepts a list of characters to ignore during the
validation it seemed logical to me to leave the responsibility of
allowing whitespace characters on the hands of the one who uses the
rule.

It's also clear that the "AbstractCtypeRule" is an unnecessary overhead
since it is only a proxy for "AbstractFilterRule". That one can and
should even be removed after this commit is applied especially because
this commit will also remove the method "filterWhiteSpaceOption" which
is the only substantial difference between "AbstractCtypeRule" and
"AbstractFilterRule".

This commit will also apply our guidelines to the "Digit" rule since we
want to do that to all the rules we have.


tests/unit/Rules/DigitTest.php
```diff
@@ -13,105 +13,87 @@
...
-            [' '],
-            [165],
-            [1650],
-            ['01650'],
-            ['165'],
-            ['1650'],
-            ['16 50'],
-            ["\n5\t"],
-            ['16-50', '-'],
+            'positive integer' => [new Digit(), 165],
```
batusa: Nice descriptions for test cases 👍

PR #1107, opened 2018-09-12
Apply contribution guidelines to "NotOptional" rule

danilosilva87:


library/Rules/NotOptional.php
```diff
@@ -15,10 +15,19 @@
 
 use Respect\Validation\Helpers\UndefinedHelper;
 
-class NotOptional extends AbstractRule
+/**
+ * Validates if the given input is not optional. By optional we consider null or an empty string ('').
```
henriquemoody: The `By...` would be better in the next line

PR #1108, opened 2018-09-12
Apply contribution guidelines to "LeapYear" rule

danilosilva87:


library/Exceptions/LeapYearException.php
```diff
@@ -13,6 +13,21 @@
...
+ * @author Henrique Moody <henriquemoody@gmail.com>
+ */
+final class LeapYearException extends ValidationException
 {
+    /**
+     * {@inheritdoc}
+     */
+    public static $defaultTemplates = [
+        self::MODE_DEFAULT => [
+            self::STANDARD => '{{name}} must be leap year',
```
henriquemoody: _... must be a leap year_ sounds better to me

PR #1110, opened 2018-09-13
Test image5.php is duplicate

danilosilva87:
The test that is in the image5.phpt file is already defined in the image.phpt file.


henriquemoody:
I sent #1111 with a different commit message.

I do recommend you to read this post: https://chris.beams.io/posts/git-commit/

Thanks for contributing! 🐼

PR #1114, opened 2018-09-16
Apply contribution guidelines to "Mimetype" rule

danilosilva87:


tests/integration/rules/minetype.phpt
```diff
@@ -0,0 +1,37 @@
+--FILE--
```
henriquemoody: This file should be called "mimetype.phpt" (it's minetype.phpt)

PR #1122, opened 2018-09-19
change rule of kr postalcode regex

ong-ar:
http://www.epost.go.kr/roadAreaCdEng.retrieveRdEngAreaCdList.comm#

Koreans use a five-digit number as their zip code.

Please check the link above.

- [ ] Create regression test


henriquemoody:
I've sent #1148 adding tests that will avoid that we change that back and with a commit message that explains a little bit better why the change is necessary.

The fix is already available in the version 1.1.27.

Issue #1141, opened 2018-10-04
Email validation throws ErrorException "Array to string conversion


KonstantinKolodnitsky:
It happend when an array is passed to the validator.
The class `Egulias\EmailValidator\Validation\RFCValidation` has `(string)$email` but doesn't have check for scalar.

henriquemoody:
Thanks for reporting and for sending the #1142! 🐼

PR #1142, opened 2018-10-04
Check input is scalar for email validation

KonstantinKolodnitsky:
https://github.com/Respect/Validation/issues/1141


henriquemoody:
I created #1146 only validating strings. I also added some tests and made sure there was a more detailed commit message.

Thanks for contributing! 🐼

PR #1144, opened 2018-10-04
Extends the UUID validation to test for a specific version

### LINKED ISSUE #1052: Requesting UUID/GUID v4 validator
Version 4 UUIDs have the form `xxxxxxxx-xxxx-4xxx-Yxxx-xxxxxxxxxxxx` where `x` is any hexadecimal digit and `y` is one of `8`, `9`, `a`, or `b`.

Regex equiv `/^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i`

weeman1337:
The UUID validation now takes a version param. The possible values are defined as constants at the Uuid validation class. Tests are extended and code is cleaned up.

Closes #1052

### DIFF
--- a/docs/rules/Uuid.md
+++ b/docs/rules/Uuid.md
@@ -1,12 +1,15 @@
 # Uuid
 
-- `Uuid()`
+- `Uuid(string $version = Uuid::VERSION_ALL)`
 
-Validates whether the type of an input is a valid UUID. Both version 1 and version 4 or supported.
+Validates whether the type of an input is a valid UUID.  
+It's optional possible to test for a specific version.
 
 ```php
 v::uuid()->validate('Hello World!'); // false
 v::uuid()->validate('eb3115e5-bd16-4939-ab12-2b95745a30f3'); // true
+v::uuid(Uuid::VERSION_1)->validate('eb3115e5-bd16-4939-ab12-2b95745a30f3'); // false
+v::uuid(Uuid::VERSION_4)->validate('eb3115e5-bd16-4939-ab12-2b95745a30f3'); // true
 ```
 
 ## Changelog
--- a/library/Rules/Uuid.php
+++ b/library/Rules/Uuid.php
@@ -13,9 +13,9 @@ declare(strict_types=1);
 
 namespace Respect\Validation\Rules;
 
+use Respect\Validation\Exceptions\ComponentException;
 use function is_scalar;
 use function preg_match;
-use Respect\Validation\Exceptions\ComponentException;
 
 /**
  * This class validates UUIDs.
@@ -56,12 +56,13 @@ class Uuid extends AbstractRule
     /**
      * Uuid constructor.
      *
-     * @param string $version Use one of the Uuid class constants.
+     * @param string $version use one of the Uuid class constants
+     *
      * @throws ComponentException
      */
     public function __construct(string $version = self::VERSION_ALL)
     {
-        if (in_array($version, self::VERSIONS) === false) {
+        if (false === in_array($version, self::VERSIONS)) {
             $message = sprintf('invalid version %s given, possible: %s', $version, join(', ', self::VERSIONS));
             throw new ComponentException($message);
         }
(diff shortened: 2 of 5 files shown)


library/Rules/Uuid.php
```diff
@@ -13,19 +13,78 @@
...
+    private $version;
+
+    /**
+     * Uuid constructor.
+     *
+     * @param string $version use one of the Uuid class constants
+     *
+     * @throws ComponentException
+     */
+    public function __construct(string $version = self::VERSION_ALL)
```
henriquemoody: I think that if you don't wanna do a specific version you just don't pass the version.

library/Rules/Uuid.php
```diff
@@ -13,19 +13,78 @@
...
+    public const VERSION_1 = '1';
+    public const VERSION_2 = '2';
+    public const VERSION_3 = '3';
+    public const VERSION_4 = '4';
+    public const VERSION_5 = '5';
+
+    public const VERSIONS = [
+        self::VERSION_ALL,
+        self::VERSION_1,
+        self::VERSION_2,
```
henriquemoody: For what's seem the version 2 is a bit more complicated than that: http://www.nesubl.eu/data/guidelines/NES%20Guideline%20-%20UUID%20-%20Version%202.pdf, one example that can be found in this documentation is `6E09886B-DC6E-439F-82D1-7CCAC7F4E3B3`, which wouldn't pass this validation.

I think would be better to simply drop the version 2 for now.

tests/integration/rules/uuid.phpt
```diff
@@ -0,0 +1,27 @@
...
+try {
+    v::uuid(Uuid::VERSION_2)->check('3111e7ae-3ce2-4ea8-b9e5-371345f8552a');
+} catch (UuidException $exception) {
+    echo $exception->getMessage().PHP_EOL;
+}
+
+?>
+--EXPECTF--
+"4785effc-383b-6507-b27a-3db6369060b9" must be a valid UUID
+"3111e7ae-3ce2-4ea8-b9e5-371345f8552a" must be a valid UUID
```
henriquemoody: Since we're adding support for versions, would be nice to be a little bit more explicit about it, because the message says that this one is not a valid UUID while it's just not a valid UUID version 2.

library/Rules/Uuid.php
```diff
@@ -13,19 +13,78 @@
...
 
+use Respect\Validation\Exceptions\ComponentException;
 use function is_scalar;
 use function preg_match;
 
+/**
+ * This class validates UUIDs.
+ *
+ * Henrique Moody <henriquemoody@gmail.com>
+ * Michael Weimann <mail@michael-weimann.eu>
```
henriquemoody: You're missing the `@author` annotation here, also `Dick van der Heiden <d.vanderheiden@inthere.nl>` is also an author of this rule.

tests/unit/Rules/UuidTest.php
```diff
@@ -13,45 +13,182 @@
...
+     *
+     * @test
+     * @dataProvider provideValidUuidConstructorParams
+     *
+     * @param mixed $version The version passed to the constructor
+     */
+    public function instantiate($version): void
+    {
+        self::expectNotToPerformAssertions();
+        new Uuid($version);
```
henriquemoody: I don't think that needs to be tested once you're using that in the data providers.

docs/rules/Uuid.md
```diff
@@ -1,12 +1,15 @@
 # Uuid
 
-- `Uuid()`
+- `Uuid(string $version = Uuid::VERSION_ALL)`
 
-Validates whether the type of an input is a valid UUID. Both version 1 and version 4 or supported.
+Validates whether the type of an input is a valid UUID.  
+It's optional possible to test for a specific version.
```
henriquemoody: Nice that you also thought about the documentation!

weeman1337:
>  Thank you very much for contributing, @weeman1337!
I've created #1150 with some changes, could you review that for me?

Well - we did twice the work then :(

weeman1337:
Closed in favor of  #1150

PR #1145, opened 2018-10-06
Apply contribution guidelines to "LanguageCode" rule

DaniloBenevides:


library/Exceptions/LanguageCodeException.php
```diff
@@ -13,14 +13,22 @@
...
+ */
+final class LanguageCodeException extends ValidationException
 {
+    /**
+     * {@inheritdoc}
+     */
     public static $defaultTemplates = [
         self::MODE_DEFAULT => [
-            self::STANDARD => '{{name}} must be a valid language',
+            self::STANDARD => '{{name}} must be a valid language code set for ISO 639',
```
batusa: Perhaps `{{name}} must be a valid ISO 639 language code`?

library/Rules/LanguageCode.php
```diff
@@ -14,21 +14,29 @@
...
+ * @author Danilo Benevides <danilobenevides01@gmail.com>
+ * @author Emmerson <emmersonsiqueira@gmail.com>
+ * @author Henrique Moody <henriquemoody@gmail.com>
  */
-class LanguageCode extends AbstractRule
+final class LanguageCode extends AbstractRule
 {
-    public const ALPHA2 = 'alpha-2';
-    public const ALPHA3 = 'alpha-3';
+    private const ALPHA2 = 'alpha-2';
```
batusa: In this case the constants being public is fine because the user can use it instead of a literal string as an option to this rule.

library/Rules/LanguageCode.php
```diff
@@ -14,21 +14,29 @@
...
+    private const ALPHA3 = 'alpha-3';
 
     /**
      * @see http://www.loc.gov/standards/iso639-2/ISO-639-2_utf-8.txt
      *
      * @var array
      */
-    protected $languageCodeList = [
+    private $languageCodeList = [
```
batusa: 👍

library/Rules/LanguageCode.php
```diff
@@ -517,9 +525,19 @@ class LanguageCode extends AbstractRule
         ['', 'ZZA'], // ZAZA; DIMILI; DIMLI; KIRDKI; KIRMANJKI; ZAZAKI
     ];
 
-    public $set;
-    public $index;
+    /**
+     * @var string
+     */
+    private $set;
```
batusa: This attribute is not being used internally, could you remove it?

library/Exceptions/LanguageCodeException.php
DaniloBenevides: Saw it on LanguageCode.md, but I can remove if needed.

DaniloBenevides:
Thanks for the kindly review , soon I will push the changes requested.

henriquemoody:
I sent #1165 applying some changes over yours, @DaniloBenevides! 

Thanks for contributing! 🐼

Issue #855, opened 2017-09-19
Curious why Video URL does not support more providers


coder4life:
So was looking at the Video URL providers and noticed there is only two.

Since there are a variety of video providers
Twitter
Twitch
Youtube
ect.
I find it hard for a library like this to support all of them, which means it makes sense to allow thrid parties to add to the service providers. However since services is private variable and no get and set methods we do not have access to the variables. Maybe services should be made a protected variable so extend classes could override it?

Now it was also interesting

henriquemoody:
The problem with public properties, specially in PHP, is that you have no idea of which value it will have. If I have a private property, I can be sure that it will be an integer, a string, or a float. But if it's public I have no idea of which type it will have, let alone it's value. 

This library has a lot of public properties and I don't think it's a good thing. We _hope_ that the users will use them according to our implementation, or else we're triggering errors all over. IMO, everything is private until somebody proves me it's public (or even protected).

About the extensibility and completeness of the VideoUrl rule, I can say that that is one of the many rules who need to always be updated (CountryCode, CurrencyCode, PostalCode, LanguageCode, SubdivisionCode...), and we also _hope_ that somebody would contribute to keep those updated. So, if you know there are more providers, and you're willing to extend our class with those, what would be the reason to keep it for you instead of sharing by sending a pull request?

sprak3000:
I'm willing to give a PR to add Twitch videos to the validator a go. After reading the contributing guidelines, I think this would be considered a new "feature", so I'll branch off of `master`. Let me know otherwise.

henriquemoody:
With so many open issues, it's become hard to really know what's really important and what I have fixed already.

I'm closing issues that are too old, and I'm not really looking at them all. If this is still relevant, please comment here and I will reopen this issue.

Thanks for reporting! 🐼

PR #1152, opened 2018-10-12
Issue #855 - Support Twitch Video URL

sprak3000:
* Adds a `twitch:video` and `twitch:clip` URL patterns to the `VideoUrl` rule. Videos are created by the Twitch channel owner. Clips are created by people watching a channel. Kept them separate for ease of regexp creation and because they are different types of content on the Twitch platform.
* Updated unit tests to include the new URL patterns.
* Updated documentation to include the new URL patterns.


henriquemoody:
Thanks for contributing, @sprak3000. 

Would it be better to only have `twitch`, with the regex `@^https?://((www|clips)\.)?twitch\.tv/(videos/[0-9]+|[a-zA-Z0-9]+)$@i`?

sprak3000:
@henriquemoody 

I considered that. Videos and clips are two different "things" in Twitch, but I'm not sure if the distinction would matter that much. The regex would have to be adjusted slightly though. The `clips` URL is only valid for the non-numeric URL portion. Wouldn't be too difficult to sort out.

henriquemoody:
Wouldn't that REGEX I wrote satisfy these conditions?

sprak3000:
@henriquemoody 

I've updated the PR to just have a `twitch` video URL service instead of the two separate.

batusa:
@sprak3000 I've gathered your changes in this Pull Request https://github.com/Respect/Validation/pull/1154

Thanks a lot for contributing! 🐼

Issue #878, opened 2017-11-01
Idea new function for case contains list character :)


tieutantan:
Hi, can you create function require must contain one or more element in list character be like this example.

`v::contains_list(['word', '@', '/'])->validate('lorem ipsum'); ` 
// return false

`v::contains_list(['word', '@', '/'])->validate('lorem ipsum word'); ` 
// return true

`v::contains_list(['word', '@', '/'])->validate('lorem ipsum @ word'); ` 
// return true

Now you have `v::in(mixed $haystack)` and `v::contains(mixed $value)`, but there only check 1 element like word or character, we need function can check multiple value.

Thanks for reading. Love you 💯%

henriquemoody:
With so many open issues, it's become hard to really know what's really important and what I have fixed already.

I'm closing issues that are too old, and I'm not really looking at them all. If this is still relevant, please comment here and I will reopen this issue.

Thanks for reporting! 🐼

PR #1155, opened 2018-10-23
New rule: ContainsAny (requested in issue #878)

dlussky:
Hi there!
I made a new rule "ContainsAny" as requested in #878

I tried to keep the behaviour consistent with contains(), but the inconsistency between checking strings and arrays is bothering me (strings are compared case insensitively contrary to array elements). I understand, that changing "contains" rule would be wrong from the backward compatibility standpoint, but maybe I should add "caseSensitive" argument to containsAny rule and make it's behaviour consistent between array elements and strings?


henriquemoody:
I just sent #1175 with a few changes.

About the behavior of the `Contains` rule, I'm not really sure whether that would be a good idea to change, perhaps we could create another rule and remove the boolean flag that verifies that. If you would like to work on it I'll be more than happy to review it!

Issue #1161, opened 2018-11-02
PostalCode Validation broken for Canada


TheOtherBrianJ:
It appears that commit [5a067fa](https://github.com/Respect/Validation/commit/5a067faec79d0cd74aa189fc535a644fa29e0c70#r31152463) may have broken PostalCode validation for Canada by inserting an errant space into the regex.

TheOtherBrianJ:
This issue is addressed by PR #1162

henriquemoody:
Thanks for reporting and for cooking the PR! 🐼

PR #1162, opened 2018-11-03
Fix Canada PostalCode; unit test. Issue #1161

TheOtherBrianJ:
This PR removes the errant space in the CA PostalCode regex. This space was introduced by 5a067fa.

The PR also adds unit tests for the CA PostalCode to validate the fix, and protect against future breaks.


batusa:
I've enhanced your commit message by adding some info to it. This fix was merged at at https://github.com/Respect/Validation/pull/1163 with tag 1.1.29.

Thanks for your contribution! 🐼

PR #1170, opened 2018-12-02
Contribution Guidelines - SymbolicLink

GusAntoniassi:
Hi, this adapts the SymbolicLink rules to match the contribution guidelines noted on #921.

I'm not sure about the monkey patching done in the `SymbolicLinkTest` (overriding `is_link`), but since it was already there I did not want to change it, and I also implemented the same idea on the integration test (`symbolicLink.phpt`).

### DIFF
--- a/library/Exceptions/SymbolicLinkException.php
+++ b/library/Exceptions/SymbolicLinkException.php
@@ -16,8 +16,11 @@ namespace Respect\Validation\Exceptions;
 /**
  * @author Henrique Moody <henriquemoody@gmail.com>
  */
-class SymbolicLinkException extends ValidationException
+final class SymbolicLinkException extends ValidationException
 {
+    /**
+     * {@inheritdoc}
+     */
     public static $defaultTemplates = [
         self::MODE_DEFAULT => [
             self::STANDARD => '{{name}} must be a symbolic link',
--- a/library/Rules/SymbolicLink.php
+++ b/library/Rules/SymbolicLink.php
@@ -16,8 +16,11 @@ namespace Respect\Validation\Rules;
 /**
  * @author Henrique Moody <henriquemoody@gmail.com>
  */
-class SymbolicLink extends AbstractRule
+final class SymbolicLink extends AbstractRule
 {
+    /**
+     * {@inheritdoc}
+     */
     public function validate($input): bool
     {
         if ($input instanceof \SplFileInfo) {
--- a/tests/integration/rules/symbolicLink.phpt
+++ b/tests/integration/rules/symbolicLink.phpt
@@ -1,0 +1,61 @@
+--FILE--
+<?php
+namespace Respect\Validation\Rules;
+
+require 'vendor/autoload.php';
+
+use Respect\Validation\Exceptions\SymbolicLinkException;
+use Respect\Validation\Exceptions\NestedValidationException;
+use Respect\Validation\Validator as v;
+
+$GLOBALS['is_link'] = null;
+
+// This override is only needed for the test, since 
+// symbolic links cannot be mocked properly in PHP.
+// See this issue for more info: https://github.com/mikey179/vfsStream/issues/89
+function is_link($link)
+{
+    $return = \is_link($link);
+    if (null !== $GLOBALS['is_link']) {
+        $return = $GLOBALS['is_link'];
+        $GLOBALS['is_link'] = null;
+    }
+
+    return $return;
+}
+
+try {
+    $GLOBALS['is_link'] = false;
+    v::symbolicLink()->check('/path/of/invalid/symbolic/link');
+} catch (SymbolicLinkException $exception) {
+    echo $exception->getMessage().PHP_EOL;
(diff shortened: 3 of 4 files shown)


GusAntoniassi:
This was the output of phpstan:
```
 ------ ----------------------------------------------------------------------- 
  Line   tests/integration/rules/symbolicLink.phpt                              
 ------ ----------------------------------------------------------------------- 
  3      Namespace declaration statement has to be the very first statement in  
         the script on line 3                                                   
 ------ ----------------------------------------------------------------------- 
```
How should this integration test be handled? I was able to do it by symlinking to /tmp/symlink.lnk but I believe that wouldn't be too cross-platform.

library/Exceptions/SymbolicLinkException.php
```diff
@@ -16,8 +16,11 @@
 /**
  * @author Henrique Moody <henriquemoody@gmail.com>
```
henriquemoody: It's just a keyword, but defining a class as final is a major change, I'd add `Gus Antoniassi <gus.antoniassi@gmail.com>` as an author here.

library/Rules/SymbolicLink.php
```diff
@@ -16,8 +16,11 @@
 /**
  * @author Henrique Moody <henriquemoody@gmail.com>
```
henriquemoody: Same here

library/Rules/SymbolicLink.php
```diff
@@ -16,8 +16,11 @@
...
  */
-class SymbolicLink extends AbstractRule
+final class SymbolicLink extends AbstractRule
 {
+    /**
+     * {@inheritdoc}
+     */
     public function validate($input): bool
     {
         if ($input instanceof \SplFileInfo) {
```
henriquemoody: We are importing everything from the global namespace, which means that the use statements of this class should be something like:

```php
use SplFileInfo;
use function is_link;
use function is_string;
```

tests/integration/rules/symbolicLink.phpt
```diff
@@ -0,0 +1,61 @@
+--FILE--
+<?php
+namespace Respect\Validation\Rules;
```
henriquemoody: You could've removed that together with the function workaround. Best way is to really have a symbolic link in the repository. It was recently added, btw: `tests/fixtures/symbolic-link`

tests/integration/rules/symbolicLink.phpt
```diff
@@ -0,0 +1,61 @@
...
+    echo $exception->getFullMessage().PHP_EOL;
+}
+
+?>
+--EXPECTF--
+"/path/of/invalid/symbolic/link" must be a symbolic link
+"/path/of/valid/symbolic/link" must not be a symbolic link
+- "/path/of/invalid/symbolic/link" must be a symbolic link
+- "/path/of/valid/symbolic/link" must not be a symbolic link
+
```
henriquemoody: You could've remove this extra line here.

tests/unit/Rules/SymbolicLinkTest.php
```diff
@@ -14,9 +14,13 @@
...
 
 use PHPUnit\Framework\TestCase;
+use Respect\Validation\Validatable;
 
 $GLOBALS['is_link'] = null;
 
+// This override is only needed for the test, since 
+// symbolic links cannot be mocked properly in PHP.
+// See this issue for more info: https://github.com/mikey179/vfsStream/issues/89
 function is_link($link)
```
henriquemoody: As a said in a different comment, you could've removed this function workaround and use a real symbolic link instead.

tests/unit/Rules/SymbolicLinkTest.php
```diff
@@ -29,56 +33,87 @@ function is_link($link)
 }
 
 /**
- * @group  rule
+ * @group rule
  * @covers \Respect\Validation\Exceptions\SymbolicLinkException
```
henriquemoody: Because we're testing the exceptions in the PHPT file, there is no need for testing this here.

tests/unit/Rules/SymbolicLinkTest.php
```diff
@@ -29,56 +33,87 @@ function is_link($link)
...
  * @covers \Respect\Validation\Rules\SymbolicLink
  *
  * @author Gabriel Caruso <carusogabriel34@gmail.com>
  * @author Henrique Moody <henriquemoody@gmail.com>
+ * @author Gus Antoniassi <gus.antoniassi@gmail.com>
```
henriquemoody: Would be better in alphabetical order.

tests/unit/Rules/SymbolicLinkTest.php
```diff
@@ -29,56 +33,87 @@ function is_link($link)
...
  */
 class SymbolicLinkTest extends TestCase
```
henriquemoody:
```suggestion
final class SymbolicLinkTest extends TestCase
```

Issue #1173, opened 2018-12-11
Create "NotEmojii" rule


henriquemoody:
There were quite some issues in the past about validating "names".

Even thought validating names is something ~almost~ impossible having a rule that validates whether the input is an emojii or not may help with that.

henriquemoody:
Closed by #1182

PR #1181, opened 2019-01-27
Add NotEmoji Rule

mazentouati:
Hey there,
This pull request bring a new validator as mentioned in this issue #1173 .

Checking the emoji is very tricky as their Unicode ranges are spread, so I used the official reference for the Unicode v11 to gather the emoji data. I created a script to scrap, organize and handle these information to generate an optimized Regex pattern that's able to catch the emoji.

### DIFF
--- a/docs/list-of-rules.md
+++ b/docs/list-of-rules.md
@@ -94,6 +94,7 @@
 - [Length](rules/Length.md)
 - [Lowercase](rules/Lowercase.md)
 - [NotEmpty](rules/NotEmpty.md)
+- [NotEmoji](rules/NotEmoji.md)
 - [NoWhitespace](rules/NoWhitespace.md)
 - [PhpLabel](rules/PhpLabel.md)
 - [Printable](rules/Printable.md)
--- a/docs/rules/NotEmoji.md
+++ b/docs/rules/NotEmoji.md
@@ -1,0 +1,20 @@
+# NotEmoji
+
+- `v::notEmoji()`
+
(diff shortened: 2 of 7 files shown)


mazentouati:
Any thoughts why the style fixing command cause Travis to fail when building for PHP7.2 ?. Thought, it works locally with PHP7.2.2.
```
vendor/bin/php-cs-fixer --diff --dry-run --verbose fix
```

It throws the following error on travis
```
PHP Fatal error:  Class 'null' not found in /home/travis/build/Respect/Validation/vendor/symfony/options-resolver/OptionsResolver.php on line 1004
```

mazentouati:
I found the issue source : it's this line in Travis Config : 
` - pecl install uopz || echo "Cound not install uopz" 1>&2`
because the extension `uopz` will cause `symfony/options-resolver` that's used by PHP-cs-fixer to fail. Though when ignoring the extension it will cause the integration test for the Upload.phpt to fail. I don't know what you like to do either drop the use of the extension and the Upload test or ignore the build fail.

henriquemoody:
Hey, @mazentouati! 

You did some amazing work in this PR! 🎉 

I'm sorry by my delay, I switched jobs recently and things are still a bit busy, but I should get back now!

I sent #1182 with a few changes to your code, but nothing compared to the work you've done!

mazentouati:
Oh Great Thanks ^^, thought I'll PR a more accurate ranges later on this night. I'll provide more details...

PR #1185, opened 2019-02-03
Enhance the emoji ranges & Fix documentation link

mazentouati:
The new ranges are now 100% accurate by not including any Unicode point that is not defined into the reference file. 

For example in the old ranges there's this interval:
```php
'[\x{3297}-\x{3299}]'
```
That is not accurate because there's no `3298` Unicode point in the specs. it may be used by another character and it will cause the check to fail if the user included that particular character.
The new ranges for that previous example are now
```php
'\x{3297}',
'\x{3299}',
```
and so on...

Also I renamed the property name from `GROUPS` to `RANGES` because these are ranges not groups.


henriquemoody:
Once again, great work @mazentouati! 🐼

Do you think you could also add some tests that would cover the changes?

mazentouati:
@henriquemoody These changes have no conflict with the previous ones, I don't think that we need to define a new tests. The new ones as I previously mentioned are 100% accurate as I looped through each Unicode point and made sure that they are successive in order to be in the same interval. Also, I made a test that check if the generated pattern able to catch all the emojis and it succeeded.

Please refer to this repository to learn more about the generator I created for this particular validator rule : https://github.com/mazentouati/no-emoji

henriquemoody:
I cherry-picked your commit as a8cb87217fee65ee642a9acf3efdf6a4c48b003b

The only change I made was to add the same information you wrote in the pull request to the commit message, so the commit has some context.

Issue #843, opened 2017-07-31
new validation rule: IBAN 


Chriskamphuis:
I would propose one new validation rule that deterimines if an input is a proper IBAN as described on [this](https://en.wikipedia.org/wiki/International_Bank_Account_Number) wikipedia page.

Chriskamphuis:
I would like to submit a pull request to implement this.

henriquemoody:
Hey, @Chriskamphuis! I will be more than happy with your PR!

DaniloBenevides:
Hello , anyone working on this issue ?

mazentouati:
@henriquemoody is this implemented ? If not I'll work on it;

henriquemoody:
It's all yours @mazentouati!

henriquemoody:
Closed by #1194

PR #1188, opened 2019-02-04
Create "IBAN" rule

mazentouati:
As requested #843 . 

The validation procedure is adhering to what's described in this [section](https://en.wikipedia.org/wiki/International_Bank_Account_Number#Validating_the_IBAN)


mazentouati:
@henriquemoody  Hey there, I really like the changes you made. I'll adhere to that rules further on. I though that writing the acronyms with upper cases would make more sense.

PR #1189, opened 2019-02-05
Apply contribution guidelines to "Zend" rule

danilosilva87:


henriquemoody:
I've worked a bit one it, I think the way this rule is right now it's pretty useless, I am almost 100% sure people just don't use it.

I've done a bit of work on that, though, but there is still some things to do:

https://github.com/Respect/Validation/compare/master...henriquemoody:rule/zend?expand=1

danilosilva87:
I agree with you @henriquemoody, here are some things we can consider.

- Create a method that instantiates the zend validator rule.
- We check if that rule really exists.

henriquemoody:
I merged #1210! 

Thank you very much for cooking this PR and for commenting on it!

PR #1191, opened 2019-02-07
Enhance SubDivisionCodes code style

mazentouati:
I think it's better this way, Also these changes will increase the code quality score.


henriquemoody:
Thank you very much for taking the time to do that! I really appreciate the effort.

However, I'm not quite sure whether that's the right direction. This is pretty much raw data, and I think it should be somewhere else (maybe a XML, JSON, or PHP array), not in the code like that. What do you think?

There are quite some other rules that are pretty much the same, like Tld.

mazentouati:
Yeah, that's what I was thinking about. I was about to write an issue suggesting that we should change these file to PHP arrays and include them in the designated classes. But I noticed that this will break a lot of the base code and tests because many dependents assuming that these are an implementation of `Validatable` interface.

Also, I struggled with learning the base code as there's no comments or diagrams that describe what's going on.

And don't worry about time, for cases like this I usually write a script that do the repetitive and time consuming tasks ^^

henriquemoody:
Hey! I started to work on that, see #1211.

https://github.com/Respect/Validation/blob/e9c502f5fb0242e971e17fb41eed6f3d5fd5d027/library/Rules/SubdivisionCode.php

I'm not very sure about the implementation, I'd like to not handle the file in the rule itself. However, I really think that using JSON for these cases make more sense.

What do you think about it?

mazentouati:
@henriquemoody yeah I think puting the data in a data format makes more sense. Though as you said handling the files inside the rule is trouble maker as it violates the Single Responsibility Principle. The appropriate way IMO is to create an utility class for Example 'Storage' with concise API that will handle the files logic.

henriquemoody:
Would you like to take over my PR's content and change it from there? I'll be more than glad if you do!

mazentouati:
@henriquemoody Yeah absolutely, I'll start working on it as soon as I have some spare time.

henriquemoody:
Closed in favor of #1212

Issue #1200, opened 2019-02-19
PostalCode Validator not accepting Irish (IE) Postal codes in Eircode format


DanielAlt:
Had reports from my Irish users that postal codes were not being accepted.

Ireland moved over to Eircode in 2015, [i read the wikipedia page](https://en.wikipedia.org/wiki/Postal_addresses_in_the_Republic_of_Ireland#Pre-Eircode_postal_districts), and found some sample postal codes on google maps. 

They aren't being matched by the regexp [used at line 84 of PostalCode.php](https://github.com/Respect/Validation/blob/1.1/library/Rules/PostalCode.php#L84)

Check this Rubular. 
https://rubular.com/r/T90vyL1hXOkWQu

So i'm writing a PR. Just have to read your contributing document first.

henriquemoody:
Fixed! Thank you very much for reporting and also for cooking the pull request!

As I mentioned in the pull request, the fix is available on version [1.1.30](https://github.com/Respect/Validation/releases/tag/1.1.30)

PR #1201, opened 2019-02-19
bugfix for Irish postal codes

DanielAlt:
Addresses Issue #1200 

Replaces the Regexp to only work for [Eircode postal codes](https://www.eircode.ie/what-is-eircode) in Ireland (IE). This is the official format since 2015. 

Prior to 2015 Dublin, and Cork, had independent formats, and the rest of the country had none. 

Struggled with the idea of combining previous regular expression with this one using an or (|), but settled on replacing it. 

Let me know.

### DIFF
--- a/library/Rules/PostalCode.php
+++ b/library/Rules/PostalCode.php
@@ -81,7 +81,7 @@ class PostalCode extends Regex
         'HT' => '/^(?:HT)*(\d{4})$/',
         'HU' => '/^(\d{4})$/',
         'ID' => '/^(\d{5})$/',
-        'IE' => '/^[A-Z]\d{2}\s?([A-Z]|\d){4}$/',
+        'IE' => '/^([A-Z]\d{2}|D6W)\s?([A-Z]|\d){4}$/',
         'IL' => '/^(\d{5})$/',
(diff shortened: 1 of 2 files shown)


library/Rules/PostalCode.php
```diff
@@ -81,7 +81,7 @@ class PostalCode extends Regex
         'HT' => '/^(?:HT)*(\d{4})$/',
         'HU' => '/^(\d{4})$/',
         'ID' => '/^(\d{5})$/',
-        'IE' => '/^[A-Z]\d{2}$|^[A-Z]{3}[A-Z]{4}$/',
+        'IE' => '/^[A-Z]\d{2}\s?([A-Z]|\d){4}$/',
```
henriquemoody: The current validation will not pass with `D6W RY72`.

I was taking a look at this page: https://www.citizensinformation.ie/en/consumer_affairs/telecommunications_and_postal_services/eircode.html#lba536

> The routing key is the first 3 characters of an Eircode. The first character is always a letter, followed by 2 numbers (except for D6W). The letters are not linked to a county or city name, except for postal districts in Dublin, which have had their existing postal codes transferred into a routing key format such as D03, D12 and D22. The same routing key can be shared by several towns and townlands.

DanielAlt: My mistake. 

 [```^([A-Z]\d{2}|D6W)\s?([A-Z]|\d){4}$```](https://rubular.com/r/WfiGQCquHFsdT1)

This should do instead.

henriquemoody:
I merged it manually with a couple of changes in the commit 9bafee191d9c560827105a849e2b29d8446f0f6f

This fix is already available on version [1.1.30](https://github.com/Respect/Validation/releases/tag/1.1.30).

PR #1211, opened 2019-03-10
Refactor "SubdivisionCode" rule

henriquemoody:

PR #1212, opened 2019-03-20
Refactor "SubdivisionCode" rule

mazentouati:
As discussed in #1191 
The provided Data helper is easy to extend through loaders to adapt new data types in future ( YAML, PHP Arrays...).
___

Enhance #1211


henriquemoody:
I'm still quite puzzled with this rule, actually not only that one but others as well. 

I'm not sure whether I'd like to keep having this data into our repository, that has some maintenance cost that I'm not sure whether I'd like to have.

What do you think about using an external library to handle this data? I've been checking some and the nicest one so far was that one: https://github.com/sokil/php-isocodes

mazentouati:
Yeah, letting another library handling this logic makes sense for me. But, the implementation of `sokil/php-isocodes`  makes me a little anxious. I made a little peek on the code source. They put all the subdivisions ( not grouped by country ) in one file and eventually somewhere they have to loop all fields to group them by country when you use the `getAllByCountryCode($countryCode)`. That's a wasted memory and execution time in my opinion.
___
https://github.com/sokil/php-isocodes/blob/master/databases/iso_3166-2.json

henriquemoody:
I've seen that and indeed it's a bit upsetting because it loads 459K json in memory, plus parsing it and iterating to find the subdivision codes from it.

I've created an issue in their repository, maybe we get some answer. I think I'd rather contribute to their repository than keeping another one which is gonna do pretty much the same.

Any thoughts?

mazentouati:
yeah, I'm up to contribute in their repository too, let's wait for their response. Also, I'm open to the idea of creating a compatible and optimized ISO Codes repository under Respect organization to handle the logic there.

henriquemoody:
Hey, @mazentouati! I created a temporary solution until we don't decide to use an external library, see #1226.

I changed your implementation just to make it simpler since it's temporary ~I hope~ and we don't want to maintain much code related to loading this type of data.

* Here the commit that is on master 718bacad046b5f636c33ab0f9d37bca4dbacb149
* Here the new SubdivisionCode: https://github.com/Respect/Validation/blob/master/library/Rules/SubdivisionCode.php
* Here the class that loads the Subdivisions: https://github.com/Respect/Validation/blob/master/library/Helpers/Subdivisions.php

Anyways, thank you very much for contributing! 🐼 

***

Please, if there is any concerns about those changes let me know!

mazentouati:
Thanks @henriquemoody, I'll pull request my suggestions.

Issue #1213, opened 2019-03-26
Length exception does not account for inclusive parameter


nekosaur:
The exception messages for the Length rule do not account for the `inclusive` parameter. 

As an example, the current wording for `MODE_DEFAULT` and `LOWER` is 
```
{{name}} must have a length greater than {{minValue}}
```

when it should be 
```
{{name}} must have a length greater than or equal to {{minValue}}
```

if `inclusive` is true.

mazentouati:
make sense, I'll upload the fix tomorrow.

henriquemoody:
Fixed, but because it breaks backwards compatibility that will be released only in the next version.

Thanks for reporting! 🐼

PR #1214, opened 2019-03-27
Add inclusive messages to LengthException

mazentouati:
as requested #1213

### DIFF
--- a/tests/integration/rules/length.phpt
+++ b/tests/integration/rules/length.phpt
@@ -1,5 +1,6 @@
 --CREDITS--
 Danilo Correa <danilosilva87@gmail.com>
+Mazen Touati <mazen_touati@hotmail.com>
 --FILE--
 <?php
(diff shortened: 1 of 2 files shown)


tests/integration/rules/length.phpt
```diff
@@ -85,14 +86,14 @@ try {
 ?>
 --EXPECT--
 "phpsp.org.br" must have a length lower than 5
-"phpsp.org.br" must have a length greater than 20
+"phpsp.org.br" must have a length greater than or equal to 20
```
henriquemoody: I think it would be better to keep both messages

mazentouati: Indeed, I updated the pull request to include all messages.

nickl-:
@alganet is back!!! Whoop whoop!!! \o/

Long time no see...

PR #1215, opened 2019-03-29
Fix bug to validate "01234567890" input in CPF rule

p4ulodi4s:
I think "01234567890" input is not valid. I added a condition to fix it


henriquemoody:

henriquemoody:
This fix is already available on version [1.1.30](https://github.com/Respect/Validation/releases/tag/1.1.30).

PR #1228, opened 2019-04-07
Optimize subdivisions helper for subsequent calls

mazentouati:
I think that the current implementation will unnecessarily load and parse the requested country's data file when the rule is called many times for the same country.

---
Optimize #1226

### DIFF
--- a/library/Helpers/Subdivisions.php
+++ b/library/Helpers/Subdivisions.php
@@ -19,26 +19,42 @@ use function file_get_contents;
 use function json_decode;
 use function sprintf;
 
+/**
+ * Helper to fetch the designated subdivision codes.
+ *
+ * @author Henrique Moody <henriquemoody@gmail.com>
+ * @author Mazen Touati <mazen_touati@hotmail.com>
+ */
 final class Subdivisions
 {
     /**
      * @var mixed[]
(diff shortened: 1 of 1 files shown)


mazentouati:
hey @henriquemoody  sorry for late reply, I was a little bit busy. Well, this suggestion is not intended to optimize validation for **multiple countries** but it's intended to optimize the validation of **multiple subdivisions** in the **same country**. Maybe for example, some web app that validate **from** somewhere **to** somewhere in the same country...

henriquemoody:
I'm sorry for my delay, @mazentouati! The last weeks have been quite busy for me, I hope things slow down a bit from now on.

I think your changes make sense, what I'm not sure is whether it is worth to keep the state and add more conditions to that class in order to get a little bit of performance for a case that doesn't seem to be very frequent.

I would rather keep it as is for the moment and hopefully we will get rid of this class anyways.

I will keep it open in case you have some other argument that justifies the change, but if either you or me can close this pull request.

Anyways, I thank you very much for taking the time to make the changes, I really appreciate! 🐼

mazentouati:
Well, I have nothing else to articulate. It's fine either way 🦄.

PR #1230, opened 2019-04-11
Apply contribution guidelines to "KeyValue" rule

danilosilva87:
Hi @henriquemoody 

As we talked a few days ago, I'm doing the "keyValeu" rule, however in my phpt test this with some behaving odd about that rule.

Since you said that I could send PR, I'll try to explain the strange behavior you're experiencing.

My KeyValeu.phpt file looks like this:

```php
declare(strict_types=1);

use Respect\Validation\Exceptions\EqualsException;
use Respect\Validation\Exceptions\KeyValueException;

// test 1
try {
    v::keyValue('password', 'equals', 'password_confirmation')
        ->check([
            'password' => 'shuberry',
            'password_confirmation' => '_shuberry_',
        ]);
} catch (EqualsException $exception) {
    echo $exception->getMessage().PHP_EOL;
}

// test 2
try {
    v::keyValue('foo', 'equals', 'bar')->check(['bar' => 42]);
} catch (KeyValueException $exception) {
    echo $exception->getMessage().PHP_EOL;
}

// test 3
try {
    v::keyValue('foo', 'equals', 'bar')->check(['foo' => 42]);
} catch (KeyValueException $exception) {
    echo $exception->getMessage().PHP_EOL;
}

// test 4
try {
    v::keyValue('foo', 'json', 'bar')->check(['foo' => 42, 'bar' => 43]);
} catch (KeyValueException $exception) {
    echo $exception->getMessage().PHP_EOL;
}

// test 5
try {
    v::not(v::keyValue('password', 'equals', 'password_confirmation'))
        ->check([
            'password' => 'shuberry',
            'password_confirmation' => 'shuberry',
        ]);
} catch (EqualsException $exception) {
    echo $exception->getMessage().PHP_EOL;
}

// test 6
try {
    v::keyValue('password', 'equals', 'password_confirmation')
        ->assert([
            'password' => 'shuberry',
            'password_confirmation' => '_shuberry_',
        ]);
} catch (NestedValidationException $exception) {
}

--EXPECT--
test 1 -> password must equal "password_confirmation"
test 2 -> Key "foo" must be present
test 3 -> Key "bar" must be present
test 4 -> "bar" must be valid to validate "foo"
test 5 -> password must not equal "password_confirmation"
test 6 -> password must equal "password_confirmation"
```
The output of my code above is the output that I think should happen, but it is not what happens with test 5 and test 6

As you can see, my test 1 and my test 5 are practically the same, but test 5 is denial.
In my test 1 it throws an "EqualsException" exception and in my test 5 it throws a "KeyValueException", is it?

In my test 6 I do the same validation of test 1, however using the `assert` method. Note that in my test 1 the output message references the names of the two fields correctly and in my test 6 it does not do this correctly, it expects the following output: `- password must equal "_shuberry_"`.

I tried to explain the maximum, any doubt I am here.

Note: The tests are with the results that I imagine are correct, but they will fail, if this is an API problem, I can write the tests myself and send them back.


henriquemoody:
Hey, @danilosilva87! I've very sorry by my delay, quite busy weeks lately.

I've checked the `KeyValue` rule and it was quite a puzzle to make it work as we intend. This rule suffers from one of the biggest problems this library has, which is the lack of granularity control.

I've sent #1232 with your work and some changes that will fix it.

Thank you very much for sending the pull request, and also for the very comprehensive description of the problem! 🐼

PR #1231, opened 2019-04-14
Apply contribution guidelines to "Vatin" rule

danilosilva87:
As the "Vatin" rule is a "Plvatin" wrapper I considered using the same
unit tests from "PlVatin" to "Vatin".

I separated the exception test that was in the Vatin.php file into the Vatin.phpt file.

### DIFF
--- a/tests/unit/Rules/VatinTest.php
+++ b/tests/unit/Rules/VatinTest.php
@@ -13,19 +13,56 @@ declare(strict_types=1);
 
 namespace Respect\Validation\Rules;
 
-use Respect\Validation\Test\TestCase;
+use Respect\Validation\Test\RuleTestCase;
 use Respect\Validation\Validatable;
+use stdClass;
 
 /**
- * @group  rule
+ * @group rule
+ *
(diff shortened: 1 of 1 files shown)


tests/integration/rules/vatin.phpt
```diff
@@ -8,11 +9,18 @@ declare(strict_types=1);
...
 require_once 'vendor/autoload.php';
 
+use Respect\Validation\Exceptions\ComponentException;
 use Respect\Validation\Exceptions\Locale\PlVatinException;
 use Respect\Validation\Exceptions\NestedValidationException;
 use Respect\Validation\Exceptions\VatinException;
 use Respect\Validation\Validator as v;
 
+try {
+    v::vatin('BR')->check('1645865778');
```
henriquemoody: This exception is not part of an integration test. Could you restore the unit test?

danilosilva87: ok @henriquemoody , I thought this test was part of the integration test

PR #1234, opened 2019-05-11
Create "ParameterStringifier"

### LINKED ISSUE #1227: Allow custom "Stringifier"
# Problem
When we create the exception message, we use the `stringify` function (from [Respect\Stringifier]) to convert the input [1] all the properties that come from the rule [2] to a string.

There are some cases, described in #1048 and #1157 in which a user would like to have a different way to convert those values to string.

# Proposed solution

Pass [Stringifier] to the [Factory] and then to the [ValidationException], so that users can customize how to stringify values.

[1]: https://github.com/Respect/Validation/blob/718bacad046b5f636c33ab0f9d37bca4dbacb149/library/Exceptions/ValidationException.php#L155
[2]: https://github.com/Respect/Validation/blob/718bacad046b5f636c33ab0f9d37bca4dbacb149/library/Exceptions/ValidationException.php#L187
[Respect\Stringifier]: https://github.com/Respect/Stringifier
[Stringifier]: https://github.com/Respect/Stringifier/blob/master/src/Stringifier.php
[Factory]: https://github.com/Respect/Validation/blob/718bacad046b5f636c33ab0f9d37bca4dbacb149/library/Factory.php
[ValidationException]: https://github.com/Respect/Validation/blob/718bacad046b5f636c33ab0f9d37bca4dbacb149/library/Exceptions/ValidationException.php

henriquemoody:
I think the `Factory` and `ValidationException` are dealing with too many things. Perhaps I should re-think how to deal with it.

Both the Translator and the Stringifier are just dealing with the message. Perhaps if I create a single object that handles the message based on the default templates, mode, and chosen templates it would be better.

***
Close #1227

### DIFF
--- a/library/Exceptions/ValidationException.php
+++ b/library/Exceptions/ValidationException.php
@@ -14,11 +14,8 @@ declare(strict_types=1);
 namespace Respect\Validation\Exceptions;
 
 use InvalidArgumentException;
-use function call_user_func;
-use function is_string;
+use Respect\Validation\Message\Formatter;
 use function key;
-use function preg_replace_callback;
-use function Respect\Stringifier\stringify;
 
 /**
  * Default exception class for rule validations.
@@ -67,9 +64,9 @@ class ValidationException extends InvalidArgumentException implements Exception
     private $params = [];
 
     /**
-     * @var callable
+     * @var Formatter
      */
-    private $translator;
+    private $formatter;
 
     /**
      * @var string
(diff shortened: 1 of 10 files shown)


library/Exceptions/ValidationException.php
```diff
@@ -76,17 +75,28 @@ class ValidationException extends InvalidArgumentException implements Exception
...
      * @param mixed $input
      * @param mixed[] $params
      */
-    public function __construct($input, string $id, array $params, callable $translator)
-    {
+    public function __construct(
+        $input,
+        string $id,
+        array $params,
+        callable $translator,
```
henriquemoody: I'm not satisfied with this being a callable anymore

library/StandardParameterStringifier.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Respect\Validation;
+
+use function is_string;
+use function Respect\Stringifier\stringify;
+
+final class StandardParameterStringifier implements ParameterStringifier
```
henriquemoody: I don't like this name

PR #1243, opened 2019-06-06
Added PSR-7 UploadedFileInterface support for the size rule 

v0idpwn:
The PSR-7 UploadedFileInterface is used by many people (including frameworks), and the validation requires just an "or", because its Size function behaves the same as SplFileInfo. I think it is valid addition.


henriquemoody:
Merged it manually by ff253c78b3cad7c084d728375c5082d09862d4e2 with a couple of additions

Thanks for contributing! 🐼

PR #1249, opened 2019-07-03
change variable name in doc

GPierre-Antoine:
variable name in dynamic method when does not match variable name in When object.
https://github.com/Respect/Validation/blob/master/library/Rules/When.php


henriquemoody:
Merged by 4d706b0845a22b094767373dd7c410515c12aa8b

Thanks for contributing! 🐼

PR #1250, opened 2019-07-05
Removed Key Prefix

noobshow:
Most Validation errors are sent to Users/Visitors or Clients and as such, they might not need to know it was a Key their inputs are being validated upon.

The PR can be rejected if it seems not a valid point or there's a way to change Rules\Key() errors.

Thanks

### DIFF
--- a/library/Exceptions/KeyException.php
+++ b/library/Exceptions/KeyException.php
@@ -30,12 +30,12 @@ final class KeyException extends NestedValidationException implements NonOmissib
      */
     protected $defaultTemplates = [
         self::MODE_DEFAULT => [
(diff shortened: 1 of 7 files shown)


CameronHall:
@noobshow you'll need to update the tests as well :)

noobshow:
> @noobshow you'll need to update the tests as well :)

Will do that now...

noobshow:
> @noobshow you'll need to update the tests as well :)

Updated the tests and seems valid, but some composer requirements couldn't pass for HP 8.0.0

henriquemoody:
Merged by 029fa7fe73521b2485076bcfaae4c4e9e859b1fa

Thanks for contributing! 🐼

PR #1252, opened 2019-07-16

bmorg:
This corrects the fix introduced in #1061. The current behavior allows for leading zeros, but only if the integer can be interpreted as an octal, so that e.g. `09` and `00` fail to validate.


bmorg:
I had chosen the wrong base branch. -> #1253

PR #1253, opened 2019-07-16

bmorg:
This corrects the fix introduced in #1061. The current behavior allows for leading zeros, but only if the integer can be interpreted as an octal, so that e.g. `09` and `00` fail to validate.

### DIFF
--- a/docs/rules/IntVal.md
+++ b/docs/rules/IntVal.md
@@ -2,7 +2,7 @@
 
 - `v::intVal()`
 
-Validates if the input is an integer.
+Validates if the input is an integer (allowing leading zeros).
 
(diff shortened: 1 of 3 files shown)


bmorg:
Will this bugfix be merged? Is version 1.x end of life?

henriquemoody:
Thet version 1.x is indeed end of life.

I suggest you to keep your fork and customize your composer.json to use your package instead of ours so you can keep this fix.

Thanks for contributing, anyways 🐼

bmorg:
This bug is still present in the current 2.x release with the exact same changes fixing it. I'll open a new pull request.

PR #1255, opened 2019-08-01
Small typo fix on "me" which should be "be".

mark-bromell:


henriquemoody:
This fix is available on version 2.0, see https://respect-validation.readthedocs.io/en/2.0/rules/When/

Thanks for contributing! 🐼

PR #1257, opened 2019-08-22
Feature/filter domain support

chrisramakers:
Added support for `FILTER_VALIDATE_DOMAIN` in `FilterVar`

- Tests added
- Docs updated


henriquemoody:
I've changed you commit slightly, as you can see on #1309.

You will find this fix on version 2.0.15.

Thanks for contributing! 🐼

PR #1259, opened 2019-08-29
Changed regex for IL to allow 5 and 7 digits

gpredl:
Changed regex for IL to allow 5 and 7 digits.
Reason: In 2013, after the introduction of the 7 digit codes, 5 digit codes were still being used widely.
https://en.wikipedia.org/wiki/List_of_postal_codes


henriquemoody:
Merged manually by b7b67da31ef38b721d01998b951863aa62db4bf2

This fix _will_ be available in version 2.0.6, which will be released together with a couple more of postal code fixes.

Thanks for contributing! 🐼

PR #1265, opened 2019-10-29
update snippet to PSR12

geekcom:
Refer: [PSR12](https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements)


henriquemoody:
Thanks for contributing! 🐼 

We're now following PSR-12 on it's completion.

PR #1275, opened 2020-03-04
test against php 7.4

sergiy-petrov:


henriquemoody:
Not needed anymore. 

I won't change the support for version 1.x, following on maintaining version 2.x from now on, and it's running with PHP 7.4 on Travis already.

Anyways, thanks for contributing, @sergiy-petrov! And thanks for refreshing this PR, @alystair!

PR #1277, opened 2020-04-08
Fix Email validator when egulias/email-validator was not installed

mpesari:
`egulias/email-validator` is an optional dependency. Validation 2.0 broke the `Email` rule by trying to use the dependency when it was not available.


henriquemoody:
Thank you for contributing! 🐼 

This fix is available on version 2.0.1.

Issue #1278, opened 2020-04-18
Automatic namespace resolution for custom exceptions


caseyamcl:
I noticed that in version 2, you need to manually add the exception namespace to the factory class as follows:

```php
$factory = (new \Respect\Validation\Factory())->withExceptionNamespace("TestRespect\\Exceptions");
\Respect\Validation\Factory::setDefaultInstance($factory);
```

Otherwise, you'll get the default `ValidationException` every time, regardless of whether or not you create your own custom exceptions?

This involves a lot of boiler-plate code for my use-case, and it would be easily remediated by using the `ReflectionClass`, which appears to already be initialized in the Factory class.

Am I doing something wrong, or is this the expected behavior?  If not, then I'd be happy to submit a PR to correct this.

henriquemoody:
Hi, @caseyamcl! 

Yes, indeed you need to manually add the exception namespace to the factory, or else you indeed get `ValidationException` every time. I made that choice for two reasons:

1) **Not all rules need a specific exception**: that gives users more flexibility when using this library. Before you needed to create an exception, even if it was just a placeholder.
2) **It makes explicit what's the exception namespace**: that gives users more flexibility when choosing when they want to put their exceptions. If this library didn't have the many rules they have, I would particularly prefer to keep the exceptions next to the rules.

Can you elaborate a bit better how this involves a lot more of boiler-plate in your use-case? I find it a bit hard to understand because the only effort I see is to add another line when defining the default factory, taking your code as an example, it would be:

```php
use Respect\Validation\Factory;
    
Factory::setDefaultInstance(
    (new Factory())
        ->withRuleNamespace('TestRespect\\Rules')
        ->withExceptionNamespace('TestRespect\\Exceptions')
);
```

caseyamcl:
Thanks for the clarification, @henriquemoody.

I read [the documentation](https://respect-validation.readthedocs.io/en/2.0/custom-rules/) several times, and it wasn't clear to me that this behavior had changed, so I didn't know if it was intentional or not.

The overhead that I believe is added versus v1 is the dependency on a service class, so you can't just run `Validator::[rule]` anymore, unless you set the default factory, but that comes with its own complication.

Whereas before you could just run `Validator::[rule]` and expect custom exceptions to "just work", now you've got to inject the `Factory` service class.

[...]

PR #1279, opened 2020-04-18
Fixed auto-resolution for custom exceptions to match v1 behavior.

caseyamcl:
This addresses the issue that I brought up (#1278).


henriquemoody:
That took me a while, but I'm finally accepting this change.

You can find this change on 8b2819e9f8d43c67bd6bac676eb260ed59b01de9, and it will be available on version 2.1.

Thanks for contributing! 🐼

Issue #1282, opened 2020-04-24
restore_error_handler() in Call validator is not invoked upon validation failure


pathumhdes:
In assert and check functions in Call validator, it sets a custom error handler and supposed to restore at the end of the function call. But the restore_error_handler is not invoked if the validation fails. 

**check function code**
```lang-php
$this->setErrorHandler($input);

try {
    $this->rule->check(call_user_func($this->callable, $input));
} catch (ValidationException $exception) {
    throw $exception;
} catch (Throwable $throwable) {
    throw $this->reportError($input);
}

restore_error_handler();
```
**Check the following example.**

```lang-php
<?php
require('autoload.php');

use Respect\Validation\Exceptions\NestedValidationException;
use Respect\Validation\Validator as v;

validate();
try {
    triggerError();
} catch (Exception $e) {
    echo get_class($e) . "\n";
}

function validate() {
    try {
        $number = '1,-2,3';
        v::stringType()->noWhitespace()
            ->call(function ($input) {
                return explode(',', $input);
            }, v::arrayVal()->each(v::intVal()->positive()))->check($number);
        echo "Valid\n";
    } catch (NestedValidationException $ne){
        echo 'Not Valid : ' . $ne->getFullMessage() . "\n";
    } catch (Exception $e) {
        echo 'Not Valid - ' . $e->getMessage() . "\n";
    }
}

function triggerError() {
    trigger_error('test error');
}
?>

```
**Output**
```
Not Valid : - "-2" must be positive
Respect\Validation\Exceptions\CallException
```
Here, after the `triggerError()` function is called it still throws a `Respect\Validation\Exceptions\CallException`.

pathumhdes:
**The fix would be the following** (both check and  assert functions):

```
$this->setErrorHandler($input);

try {
    $this->rule->check(call_user_func($this->callable, $input));
} catch (ValidationException $exception) {
    throw $exception;
} catch (Throwable $throwable) {
    throw $this->reportError($input);
} finally {
    restore_error_handler();
}
```

pathumhdes:
Hi @henriquemoody 
Can you please have a look at this?

henriquemoody:
This fix is already available on version 2.0.4

Thanks for reporting and sending a pull request with the fix! 🐼

PR #1283, opened 2020-04-29
Fixing issue #1282

pathumhdes:
Fix for: restore_error_handler() in Call validator is not invoked upon validation failure #1282


dktapps:
I was just about to PR the exact same fix :+1: . I just came across some bizarre fatal errors in my project that were caused by @silenced errors triggered while the validation error handler was in effect.

henriquemoody:
Merged manually by 7154e90522417ada235c4360223b7d12ef0241a0

This fix is already available on version 2.0.4

Thanks for contributing! 🐼

PR #1289, opened 2020-05-16
Each::assert(): fix only the first error being reported when given a …

dktapps:
…composite rule

```yml
  lol:
    default: 2
    description: []
    children:
      - nope
```

```php
$validator = new Validator(
			(new Callback(function(array $input) : bool{
				$keyValidator = new Validator(new ArrayType(), new Each(new StringType()));
				$keyValidator->assert(array_keys($input));
				return true;
			}))->setName("permission keys"),
			new Each(new Validator(
				new Key("default", new OneOf(new StringType(), new BoolType()), false), //TODO: enumerate possible values
				new Key("description", new StringVal(), false),
				new Key("children", new ArrayType(), false)
			))
		);
		$validator->setName("permissions");
		try{
			$validator->assert($data);
		}catch(NestedValidationException $e){
			var_dump($e->getFullMessage());
		}
```

produced:
```
- default must be of type string
```
while it should produce something like:
```
- These rules must pass for `{ "default": 2, "description": { }, "children": { "nope" } }`
  - Only one of these rules must pass for default
    - default must be of type string
    - default must be of type boolean
  - description must be a string
```

### DIFF
--- a/library/Rules/Each.php
+++ b/library/Rules/Each.php
@@ -55,7 +55,7 @@ final class Each extends AbstractRule
         $exceptions = [];
         foreach ($input as $value) {
             try {
-                $this->rule->check($value);
+                $this->rule->assert($value);
             } catch (ValidationException $exception) {
                 $exceptions[] = $exception;
             }


henriquemoody:
Are you sure that's really the behavior you expect? In your example, there will be only one element that fails, if you try with the input above:

```php
$data = [
    [
        'default' => 2,
        'description' => [],
        'children' => ['nope'],
    ],
    [
        'default' => 2,
        'description' => [],
        'children' => ['nope'],
    ],
];
```

Then you will get the message:

```
- Each item in `{ { "default": 2, "description": { }, "children": { "nope" } }, { "default": 2, "description": { }, "children": { "nope" } } }` must be valid\n
  - These rules must pass for `{ "default": 2, "description": { }, "children": { "nope" } }`\n
    - Only one of these rules must pass for default\n
      - default must be of type string\n
      - default must be of type boolean\n
    - description must be a string\n
  - These rules must pass for `{ "default": 2, "description": { }, "children": { "nope" } }`\n
    - Only one of these rules must pass for default\n
      - default must be of type string\n
      - default must be of type boolean\n
    - description must be a string'
```

Rather than:

```
- Each item in `{ { "default": 2, "description": { }, "children": { "nope" } }, { "default": 2, "description": { }, "children": { "nope" } } }` must be valid
  - default must be of type string
  - default must be of type string
```

Particularly, I don't find neither very useful, but I was curious about what you were expecting.

dktapps:
@henriquemoody the problem is that only the first error in each list element is being reported. In this case, there's only one list element, but that single element contains multiple errors that should be reported.

The current behaviour (which I believe is just a plain bug) suggests to the user that the only problem is with the `default` key and that it has to be a string only, which defeats the point of using `assert()`. There are actually 2 problems: `default` is neither a string nor bool, _and_ `description` is not a string.

I pasted actual vs expected output in the OP at the bottom.

henriquemoody:
Merged manually by 350119229370247ad89f101486288f5947b1b1b1

This fix is already available on version 2.0.5.

Thanks for contributing and for you reply! 🐼

dktapps:
Great, thanks!

PR #1292, opened 2020-05-19

bmorg:
This is the pull request of #1253 for the 2.0 branch.

It corrects the fix introduced in #1061. The current behavior allows for leading zeros, but only if the integer can be interpreted as an octal, so that e.g. `06` and `07` validate but `09` and `00` do not.

### DIFF
--- a/docs/rules/IntVal.md
+++ b/docs/rules/IntVal.md
@@ -2,7 +2,7 @@
 
 - `IntVal()`
 
-Validates if the input is an integer.
+Validates if the input is an integer (allowing leading zeros).
 
 ```php
 v::intVal()->validate('10'); // true
--- a/library/Rules/IntVal.php
+++ b/library/Rules/IntVal.php
@@ -16,7 +16,9 @@ namespace Respect\Validation\Rules;
 use function filter_var;
 use function is_bool;
 use function is_float;
-use const FILTER_FLAG_ALLOW_OCTAL;
+use function is_string;
+use function ltrim;
+use function strlen;
 use const FILTER_VALIDATE_INT;
 
 /**
@@ -39,6 +41,16 @@ final class IntVal extends AbstractRule
             return false;
         }
 
-        return filter_var($input, FILTER_VALIDATE_INT, FILTER_FLAG_ALLOW_OCTAL) !== false;
+        if (is_string($input) && strlen($input) > 0) {
+            // allow leading zeros
+            $input = ltrim($input, '0');
+
+            if ($input === '') {
+                // input contained only zeros
+                return true;
+            }
+        }
+
+        return filter_var($input, FILTER_VALIDATE_INT) !== false;
     }
 }
--- a/tests/unit/Rules/IntValTest.php
+++ b/tests/unit/Rules/IntValTest.php
@@ -41,7 +41,9 @@ final class IntValTest extends RuleTestCase
             [$rule, 123456],
             [$rule, PHP_INT_MAX],
             [$rule, '06'],
+            [$rule, '09'],
             [$rule, '0'],
+            [$rule, '00'],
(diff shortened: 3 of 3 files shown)


bmorg:
Great CI integration! I fixed the QA errors in the initial commit but I'm not sure if I can squash the three commits or if you'd have to do that on merge.

danilocorrea87:
Hi @bmorg , you can do a `squash` in your local environment and do `push - force` of your branch.

You can run the command `composer qa` in your local environment, it will run all tasks in the CI

bmorg:
Squashed the commits and ran QA locally. Thanks for the advice @danilocorrea87 !

henriquemoody:
Hi, @bmorg! 

I'm not sure whether this fix is correct, though, because 09, is not an _integer_. 

```
php > var_dump(09);
Parse error: Invalid numeric literal in php shell code on line 1
```

But 07, in on the other hand, it's an integer, because it uses the octal numeral system:

```
php > var_dump(07);
int(7)
```

See also an example of 01234567:

```
php > var_dump(01234567);
int(342391)
```

The octal numeral system is the base-8 number system, and uses the digits 0 to 7 (see [Wikipedia](https://en.wikipedia.org/wiki/Octal)).

Aren't you maybe confusing this rule with the [Digit](https://respect-validation.readthedocs.io/en/2.0/rules/Digit/) rule?

bmorg:
Hi @henriquemoody!

I think there is some confusion about integer literals (as they are e.g. typed into source code) and integer values (the actual value they represent) here.

Yes, `014` is a valid octal representation of the decimal number `12`, but it is also a valid way to write the decimal number `14` ([see](https://en.wikipedia.org/wiki/Integer)).

I might be looking at this problem from a limited point of view, since we are using the library for validating user input in HTML forms exclusively, but let me elaborate.

The docs don't actually specify what exactly *integer* is supposed to mean in the context of the library, and without that definition it's impossible to understand what the validator should be doing.

There are at least these three definitions in this context:

---

**1. "Can be expected from a user entering an integer into a HTML form"**

From a web user standpoint IMO the only interesting definition is entering a [mathematical integer](https://en.wikipedia.org/wiki/Integer) into a web form. The server side implementation details are of no concern.

**2. "Can be cast to type 'int' by PHP"**

PHP happily accepts all kind of strings and converts them to the internal integer type, e.g. `(int) '089ab'` results in `int(89)` and the addition `'012ab' + 5` (without explicit casting) results in `int(17)`.

**3. "Can be interpreted as an integer literal in PHP source code"**

If I understand you correctly, this is the definition you are arguing for in the comment [above](https://github.com/Respect/Validation/pull/1292#issuecomment-631388740_).

---

I would argue for definition (1) to be the most useful. Then [leading zeros](https://en.wikipedia.org/wiki/Leading_zero) would not influence the values of the integers and `08` and `09` would indeed be valid integers. As you suggested, that is possibly what the `Digit()` rule now checks for in v2.x?

If the validator is to be interpreted in the sense of definition (3) then it should probably also accept binary and hexadecimal literals, e.g. `0b101010` and `0x2a`.

NB: Both `Number` and `NumericVal` validate `08`.

Sorry for the long comment!

henriquemoody:
Thanks for your patience, as I took a long time to reply.

I see your point now, and I agree that `IntVal` should allow leading zeros. I have updated your commit, adding some more information about the change and also updating the documentation. 

You can find my changes in the pull request #1306, and before I merge that, I would like to hear some feedback on the documentation, commit message, and on the changes as well.

Thank you for going that far on your contribution, I highly appreciate that! 🐼

henriquemoody:
You can find this fix on version 2.0.14.

Thanks everyone for contributing! 🐼

PR #1300, opened 2020-06-23
I think you meant to import the built-in `Error` class...

caseyamcl:
...rather than the `PHPUnit\Framework\Error\Error` class.  The tests succeed with the former and fail with the latter. (php v7.4.5)


henriquemoody:
I actually meant the PHPUnit exception, but for some reason it works on Travis but it doesn't work on your machine, also not on mine after I've updated things. 

Anyways, I've fixed the issue, and you shall find the changes on commit 6f3979cb5192f53d0c8990b33155078dad463253.

Thanks for contributing! 🐼

PR #1301, opened 2020-06-23
Update dateTime.phpt

caseyamcl:
Set default timezone explicitly to UTC so that tests do not fail in other time-zones.  This doesn't seem to affect other tests when I run `composer phpunit`


henriquemoody:
I've manually merged your commit by f3faf50340ab220b6eeab92d705cfb72efcf2973, only formatting the commit message according to our standards and setting the timezone to UTC to other date- and time-related rules.

Thanks for contributing! 🐼

PR #1306, opened 2020-07-14
Allow leading zeros when using "IntVal" rule

henriquemoody:
There is some confusion about integer literals (as we type them into
source code) and integer values (the actual value they represent).

When casting the integer 08 (without quotes), PHP triggers an error as
integers starting with 0 should have base 8. However, when casting the
string '08' as an integer PHP returns the integer 8.

This commit will change the behavior of the "IntVal" rule, allowing it
to accept any integer type and any representation of an integer as a
string.

### DIFF
--- a/docs/rules/IntVal.md
+++ b/docs/rules/IntVal.md
@@ -2,12 +2,27 @@
 
 - `IntVal()`
 
-Validates if the input is an integer.
+Validates if the input is an integer, allowing leading zeros and other number bases.
 
 ```php
 v::intVal()->validate('10'); // true
+v::intVal()->validate('089'); // true
 v::intVal()->validate(10); // true
+v::intVal()->validate(0b101010); // true
+v::intVal()->validate(0x2a); // true
 ```
+
+This rule will consider as valid any input that PHP can convert to an integer,
+but that does not contain non-integer values. That way, one can safely use the
+value this rule validates, without having surprises.
+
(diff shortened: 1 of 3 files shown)


docs/rules/IntVal.md
```diff
@@ -2,24 +2,40 @@
 
 - `IntVal()`
 
-Validates if the input is an integer.
+Validates if the input is an integer, allowing leading zeros.
```
batusa, line 5:
```suggestion
Validates if the input is an integer, allowing leading zeros and other number bases.
```

bmorg:
Hi @henriquemoody! I'm providing some feedback you asked for in [#1292](https://github.com/Respect/Validation/pull/1292#issuecomment-662070222). Sorry I've only come around to providing it now that the change is already merged.

Overall I think the `IntVal` validator is in a good state now! Your version is much more straight forward than the one I provided.

The commit message (and the docs) might be slightly misleading since it says the validator will "... accept ... any representation of an integer as a string" while the string `"0b101010"` will actually not validate (which is ok IMO).

henriquemoody:
Thanks for you feedback, @bmorg!

Unfortunately, we can't change the commit message anymore, but about the documentation, could you submit a pull request with how you think that could be better explained? I would be very pleased to merge it!

Once again, thank you very much for your feedback! 🐼

Issue #1312, opened 2020-08-16
Formatted postal code validation for Greece (GR)


omega3000:
Hi, I have an issue with greek postcodes.
You validate for 5-digits, which is fine but an empty space after the third digit is quite commonly used and should not fail the validation.

https://www.geonames.org/postalcode-search.html?q=&country=GR
https://www.grcdi.nl/gsb/greece.html

PS, great job with this validation package!

alganet:
Hey @omega3000, I did a quick change that allows for the optional space, let's wait for PR reviews! If everyone agrees, it will be on the next version.

Thanks for reporting! :D

henriquemoody:
The fix will be available on version 2.1. 

I've made some changes in the continuous integration on master, so now it's easier to wait for the next version than creating a patch.

The most important thing I'm considering for version 2.1 is to make some changes to #1279, but I haven't done anything yet.

I will keep you posted, @omega3000. 

And thanks for the pull request, @alganet! 🐼

PR #1315, opened 2020-09-01
Allow optional space for greek postal codes

alganet:
See #1312 for more details


henriquemoody:
Good seeing you here, @alganet! 🐼

PR #1330, opened 2020-11-16
Add new country: Kosovo in CountryCode.php

bastienrouge:
This country was missing inside the country codes array.


henriquemoody:
I've updated the whole list of countries, integrating with [php-isocodes](https://github.com/sokil/php-isocodes).

You should find Kosovo when using version 2.2.0.

Thanks for contributing! 🐼

PR #1339, opened 2021-02-09
Update PostalCode.php

gpredl:
Whitespace in CZ zip is mandatory


gpredl:
Source to review:
https://en.wikipedia.org/wiki/List_of_postal_codes

henriquemoody:
This fix is available on version 2.3!

Issue #1326, opened 2020-10-14
IntVal fails with negative integer strings


franksl:
Hi,
I noticed a behaviour that I don't understand: passing a negative integer string like "-1" makes IntVal fail.
The culprit is the internal call to ctype_digit that evaluates to false as documented, but to my understanding, a rule called IntVal should evaluate to true even for negative integers.
Thanks,
Frank

PR #1355, opened 2021-05-05
IntVal: Validate negative numbers properly

l-x:
This is a quck fix for #1326


l-x:
If the failing phpstan is my fault please let me known :)

PR #1384, opened 2022-06-08
Rule/portuguese nif rule

goncalo-andrade:
Hi!

I work in a project where we use Respect/Validation. Something that we regularly have to check for is the validity of a Portuguese fiscal identification number (NIF). Since Validation does not, at this time, have such a validation rule, I decided to contribute one. I hope you find it useful.
I named the rule `PortugueseNif` because there is already a rule called `Nif`. I'm not sure this is the best name for it, so I ask for your input on this.

I tried to create this pull request in a way that follows Validation's contribution guidelines. However, if anything is not as you would expect, let me know and I'll try to fix it :)

### DIFF
--- a/docs/rules/PortugueseNif.md
+++ b/docs/rules/PortugueseNif.md
@@ -1,0 +1,19 @@
+# Portuguese NIF
+
+- `PortugueseNif()`
+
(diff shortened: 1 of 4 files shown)


TCB13:
Please merge this. It follows the official NIF spec.

TCB13 (approved):
Looks it properly implements the official gov spec.

PR #1394, opened 2022-11-15
Use `json_validate` in `Rules\Json` if available

Ayesh:
[`json_validate` function](https://wiki.php.net/rfc/json_validate) [added in PHP 8.3](https://php.watch/versions/8.3/json_validate) validates a given string input to contain valid JSON without decoding it in memory.

This adds a function availability check to `Rules\Json`, and uses the new function instead of decoding the given input, followed by a last-error check.

### DIFF
--- a/library/Rules/Json.php
+++ b/library/Rules/Json.php
@@ -36,6 +36,10 @@ final class Json extends AbstractRule
             return false;
         }
 
+        if (function_exists('json_validate')) {
+            return json_validate($input);
+        }
+
         json_decode($input);
 
         return json_last_error() === JSON_ERROR_NONE;


nickl-:
This is a valid patch...keeping open for reference if for nothing else

Ayesh:
Thank you. I accidentally deleted my fork, but I'm available to make any further changes if necessary.

nickl-:
> Thank you. I accidentally deleted my fork, but I'm available to make any further changes if necessary.

I figured as much =) 

The diff is still working and it is a simple enough addition. I don't want to have you make more effort because I don't know when someone will be available to merge this. Thank you for your patience.

PR #1405, opened 2023-02-17
Check uninitialized properties without throwing exception

gtrbunny:
PHP 7.4.0+ throws an uninitialized exception when using object validation if the property hasn't been initialized. This additional check should be safe given 7.3 support is dropped.


henriquemoody:
There's an issue with your fix. If a property exists, the Attribute rule should pass regardless.

I've made a few changes and merged it on 8d7d783698cb1c41dc9a35c919413c586f14b745. This fix will be present in version 2.3.

Thank you for contributing, @gtrbunny! 🐼

PR #1412, opened 2023-03-11
Increase test coverage for some rules

dcorrea777:
- Add test for null value in the Cnpj rule.
- Add UploadedFileInterface object test for Size rule.
- Add test for invalid values in Sorted rule.


dcorrea777:
I don't know if the message pattern is correct, but I can change it if necessary.

PR #1418, opened 2023-03-22
Rename "Attribute" rule to "Property"

henriquemoody:
Because now we have the concept of attributes in PHP, the rule with the name "Attribute" makes no sense because it doesn't validate attributes but properties.

In the future, it might be possible that Validation will have a rule called "Attribute" to validate PHP attributes.


henriquemoody:
I'm not in a rush to get any of my changes merged. I was writing some code just because, as a manager, I sometimes miss coding :D

PR #1420, opened 2023-04-02
Improve tests

henriquemoody:


henriquemoody:
@alganet, There's still some work to do here, but I figured that upgrading to PHPUnit 10 would mean that we would need to drop support for PHP 8.0. I would say that we should go for it, at the moment there are only security fixes for version 8.0. Any thoughts?

henriquemoody:
That makes sense, @alganet!

I don't think there's any changes that required PHP 8.1 besides upgrading `respect/coding-standards` and `phpunit/phpunit`.

I will update a few commit messages, remove the commits with the package upgrades, and will let you know when the changes are good for review!

PR #1424, opened 2023-04-22
Bump "respect/stringifier" to version 2.0.0

henriquemoody:


henriquemoody:
@alganet , there are a few more things that I'd like to contribute to (upgrade PHPUnit for example), but without dropping support for PHP 8.0 I can't. However, contributing on master is very tricky, because 2.3 is growing on a different branch.

What should I do?

henriquemoody:
Hey, @alganet!

I assumed you took a break from Validation, so I went ahead and made changes to it. A lot of changes... I mean, a lot... I hope that's okay with you, but if it's not, we can always revert stuff.

PR #1430, opened 2023-07-10
Documentation: Fix parameter type of keySet

Csardelacal:
The documentation states that we should use Validator::keySet() in combination with Validator::key() but the return type of key() does not match the expected parameter type of keyset().

If I am not mistaken, the StaticValidator interface is only used as a mixin for IDEs and Static Code Analysis to determine whether the library is being used correctly, so this change should have no side effects.

### DIFF
--- a/library/StaticValidator.php
+++ b/library/StaticValidator.php
@@ -10,7 +10,6 @@ declare(strict_types=1);
 namespace Respect\Validation;
 
 use finfo;
-use Respect\Validation\Rules\Key;
 
 interface StaticValidator
 {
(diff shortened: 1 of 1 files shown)


Csardelacal:
Some tests were failing on my machine. It seems like the library is generating debug output, which looks like this:

```
--- Expected
+++ Actual
@@ @@
-'bool(false)\n
+'Standard input code:19:\n
+bool(false)\n
+Standard input code:20:\n
 bool(true)\n
```

These were failing before and after the change, and manually reviewing them seems that they're just broken by the debugging information.

henriquemoody:
I've made a few changes and I merged it via 4c91e3397ac600e3c53562a977002f3edc17ff6a

This change is available on version 2.3.0.

Thank you for contributing! 🐼

PR #1453, opened 2024-03-20
Allow building rules using prefixes

henriquemoody:
This PR introduce the allowing to create rules with the following prefixes.

I'm not so sure about that, and before I merge that to master, I thought about gathering some feedback first.

Here's how it will look like:

```php
v::keyEquals('foo', 12)->assert(['foo' => 10]);
// foo must equal 12

v::propertyBetween('foo', 1, 3)->assert((object) ['foo' => 5]);
// foo must be between 1 and 3


v::lengthGreaterThan(3)->assert('foo');
// The length of "foo" must be greater than 3

v::maxOdd()->assert([1, 2, 3, 4]);
// As the maximum of `[1, 2, 3, 4]`, 4 must be an odd number

v::minEven()->assert([1, 2, 3]);
// As the minimum from `[1, 2, 3]`, 1 must be an even number

v::notBetween(1, 3)->assert(2);
// 2 must not be between 1 and 3


v::nullOrBoolType()->assert('string');
// "string" must be of type boolean

v::nullOrBoolType()->assert(null);
// 👆 doesn't throw any exception


v::undefOrUrl()->assert('string');
// "string" must be a URL

v::undefOrUrl()->assert(null);
// 👆 doesn't throw any exception

v::undefOrUrl()->assert('');
// 👆 doesn't throw any exception
```

Any thoughts?


tests/integration/transformers/prefix.phpt
```diff
@@ -0,0 +1,84 @@
...
+[
+    'foo' => 'foo must equal 12',
+]
+
+length
+⎺⎺⎺⎺⎺⎺
+The length of "foo" must be greater than 3
+- The length of "foo" must be greater than 3
+[
+    'length' => 'The length of "foo" must be greater than 3',
```
henriquemoody, line 35: This is an example of something that could be tricky. The actual rule is "length" but what the user called was "lengthGreaterThan"

PR #1457, opened 2024-04-21
Increase test coverage of the following Rules

dcorrea777:
- Base64
- Base
- Bsn
- Contains
- Executable
- FilterVar
- In
- Ip
- Isbn
- NoWhiteSpace
- Pesel
- Phone
- PolishIdCard
- PortugueseNif
- PublicDomainSuffix
- StartsWith
- Subset
- Version
- VideoUrl


tests/unit/Rules/BaseTest.php
```diff
@@ -11,12 +11,23 @@
...
 #[CoversClass(Base::class)]
 final class BaseTest extends RuleTestCase
 {
+    #[Test]
+    public function itShouldThrowsExceptionWhenBaseIsNotValid(): void
+    {
+        $this->expectException(InvalidRuleConstructorException::class);
+        $this->expectExceptionMessage('a base between 1 and 62 is required');
+
+        (new Base(63))->evaluate('011010001');
```
henriquemoody, line 28:
```suggestion
        new Base(63);
```

tests/unit/Rules/PhoneTest.php
```diff
@@ -47,6 +49,15 @@ public function shouldValidateInvalidInputWithCountryCode(string $countryCode, m
         self::assertInvalidInput(new Phone($countryCode), $input);
     }
 
+    #[Test]
+    public function itShouldThrowsExceptionWhenCountryCodeIsNotValid(): void
+    {
+        $this->expectException(InvalidRuleConstructorException::class);
+        $this->expectExceptionMessage('Invalid country code BRR');
+
+        (new Phone('BRR'))->evaluate('+1 11 91111 1111');
```
henriquemoody, line 58:
```suggestion
        new Phone('BRR');
```

tests/unit/Rules/VideoUrlTest.php
```diff
@@ -11,12 +11,23 @@
...
 #[CoversClass(VideoUrl::class)]
 final class VideoUrlTest extends RuleTestCase
 {
+    #[Test]
+    public function itShouldThrowsExceptionWhenVideoUrlIsNotValid(): void
+    {
+        $this->expectException(InvalidRuleConstructorException::class);
+        $this->expectExceptionMessage('"tiktok" is not a recognized video service.');
+
+        (new VideoUrl('tiktok'))->evaluate('https://tiktok.com/video/71787467');
```
henriquemoody, line 28:
```suggestion
        new VideoUrl('tiktok');
```

henriquemoody (changes requested):
Happy to see you here, @dcorrea777!

I've made a few suggestions removing the `->evaluate(...)`. That's because when we keep it, it's not clear that the exception is throw on a constructor level, in fact, it gives the impression that the exception is throw in the `evaluate()` method.

dcorrea777:
I was actually a bit unsure whether it was really necessary to call the evaluate method, but now it's clear to me. Thank for you reply @henriquemoody

henriquemoody:
The only thing I would change in your merge request is the commit message. The short message says, "Increase test coverage of the following Rules," but when you see that in the commit list, that doesn't make much sense because the following rules are not in the short message. In this case, I would just say, "Increase code coverage for some rules"

henriquemoody:
Merged by 2d90bc3484dc0739e5dab91e37b4067eecb22e60

Thank you for contributing, @dcorrea777!! 🐼

Issue #1462, opened 2024-07-03
DateDiff rule


gvieiragoulart:
Hey guys, I'm new in this project and I want to contribute by doing the DateDiff rule. But I noticed that we don't have any of differentiation rule.

What do you guys think about creating a new Core rule to help these differentiation rules?

henriquemoody:
Hi there!

I've started working on that rule, but due to a succession of major life events, I couldn't complete it for version 3.0. 

https://github.com/henriquemoody/Validation/commit/9e631a8ba18329ecae419fe4f165b656bfb5d406

Is that what you were looking for, @gvieiragoulart? I'd be happy if anyone picks this up.

gvieiragoulart:
Hey @henriquemoody @alganet !

If I could pick this task I would be really happy to try complete but it seems to me that you have done almost the entire work. What do you see that we have to improve?

henriquemoody:
@gvieiragoulart, I don't really know what needs to improve. I know the tests are not passing, though. I won't be working on it any time soon, so if you want to pick it up, I'd be very happy with it!

gvieiragoulart:
@alganet 

I didn't find any Age validator on the code to reuse :( i think i missed something.

gvieiragoulart:
Hey guys, I believe that I completed the feature. 

@alganet  I didn't use those validators because the _type_ is specific to the DateInterval types, so I created another validation.

PR here https://github.com/Respect/Validation/pull/1463

henriquemoody:
Closed by cf03f8ad5f25e6f887ad45ae726d1ba623987e63

Thank y'all for the comments!

PR #1463, opened 2024-07-14
Rule/date time diff

gvieiragoulart:
Resolves https://github.com/Respect/Validation/issues/1462

----

### What?

This PR introduces a new validation rule, DateTimeDiff wich validates the difference of date/time against a specific rule.

The `$type` argument on our construct should follow PHP's DateInterval properties. So we can use types such as _year_, _months_,_days_,_full days_,_hours_,_minutes_,_seconds_,_microseconds_.

----

**Note:** This is my first PR in a relatively large open-source project, so if I've done anything incorrectly, please let me know. :+1:

### DIFF
--- a/library/Helpers/CanValidateDateTime.php
+++ b/library/Helpers/CanValidateDateTime.php
@@ -21,10 +21,7 @@ trait CanValidateDateTime
 {
     private function isDateTime(string $format, string $value): bool
     {
-        $exceptionalFormats = [
-            'c' => 'Y-m-d\TH:i:sP',
-            'r' => 'D, d M Y H:i:s O',
-        ];
+        $exceptionalFormats = $this->getExceptionalFormats();
 
         $format = $exceptionalFormats[$format] ?? $format;
 
@@ -74,5 +71,14 @@ trait CanValidateDateTime
         }
 
         return checkdate($info['month'] ?: 1, 1, $info['year'] ?: 1);
+    }
+
+    /** @return array<string, string> */
+    private function getExceptionalFormats(): array
+    {
+        return [
+            'c' => 'Y-m-d\TH:i:sP',
+            'r' => 'D, d M Y H:i:s O',
+        ];
     }
 }
--- a/library/Helpers/CanValidateDateTime.php
+++ b/library/Helpers/CanValidateDateTime.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 namespace Respect\Validation\Helpers;
 
+use DateInterval;
 use DateTime;
 use DateTimeZone;
 
@@ -16,6 +17,10 @@ use function checkdate;
 use function date_default_timezone_get;
 use function date_parse_from_format;
 use function preg_match;
+use function array_keys;
+use function in_array;
+use function get_object_vars;
+
 
 trait CanValidateDateTime
 {
@@ -54,6 +59,17 @@ trait CanValidateDateTime
     private function isDateTimeParsable(array $info): bool
     {
         return $info['error_count'] === 0 && $info['warning_count'] === 0;
+    }
+
+    /**
+     * Validates if the given string is a valid DateInterval type.
+     * 
+     * @param string $age
+     * @return bool
+     */
+    private function isDateIntervalType(string $age): bool
+    {
+        return in_array($age, array_keys(get_object_vars((new DateInterval('P1Y')))));
     }
 
     private function isDateFormat(string $format): bool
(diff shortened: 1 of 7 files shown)


library/Rules/DateTimeDiff.php
```diff
@@ -0,0 +1,152 @@
...
+            'm' => 'months',
+            'd' => 'days',
+            'days' => 'full days',
+            'h' => 'hours',
+            'i' => 'minutes',
+            's' => 'seconds',
+            'f' => 'microseconds',
+        };
+    }
+}
```
gvieiragoulart, line 152: If we are going to use this, I believe that we can improve to a Enum file. I didn't create because the project don't have any

library/Helpers/CanValidateDateTime.php
```diff
@@ -59,6 +61,17 @@ private function isDateTimeParsable(array $info): bool
...
+    /**
+     * Validates if the given string is a valid DateInterval type.
+     * 
+     * @param string $age
+     * @return bool
+     */
+    private function isDateIntervalType(string $age): bool
+    {
+        return in_array($age, array_keys(get_object_vars((new DateInterval('P1Y')))));
+    }
```
gvieiragoulart, line 73: I know this seems weird, but i wanted to use those DateInterval vars. what do you guys think about it? How do we can improve this validation?

library/Rules/DateTimeDiff.php
```diff
@@ -35,7 +35,7 @@ final class DateTimeDiff extends Standard
 
     private readonly Validatable $rule;
 
-    /** @param "years"|"months"|"days"|"hours"|"minutes"|"seconds"|"microseconds" $type */
+    /** @param string $type "years"|"months"|"days"|"hours"|"minutes"|"seconds"|"microseconds" */
```
henriquemoody: This was like that on purpose, it's a literal constant. See https://phpstan.org/writing-php-code/phpdoc-types#literals-and-constants

library/Rules/DateTimeDiff.php
```diff
@@ -35,7 +35,11 @@ final class DateTimeDiff extends Standard
 
     private readonly Validatable $rule;
 
-    /** @param string $type "years"|"months"|"days"|"hours"|"minutes"|"seconds"|"microseconds" */
+    /** 
+     * @param string $type "years"|"months"|"days"|"hours"|"minutes"|"seconds"|"microseconds" 
+     * @param string|null $format Example: "Y-m-d H:i:s.u"
+     * @param DateTimeImmutable|null $now The value that will be compared to the input
```
henriquemoody, line 47: I'm a against adding documentation to the code like that, I think the name of the parameter should be pretty obvious for what it does, otherwise we should do that for everything. I understand that it can be useful to add some documentation here and there, but for this specific method, I would rather not have it.

library/Rules/DateTimeDiff.php
```diff
@@ -36,22 +35,20 @@ final class DateTimeDiff extends Standard
     private readonly Validatable $rule;
 
     /** 
-     * @param string $type "years"|"months"|"days"|"hours"|"minutes"|"seconds"|"microseconds" 
-     * @param string|null $format Example: "Y-m-d H:i:s.u"
+     * @param string $type DateInterval format examples: (y, m, d, days, h, i, s, f)
```
henriquemoody: I do understand the appeal of using DateInterval formats directly. However, I think having more descriptive values is better as it's more understandable. With the literal constant, there's not reason to translate anything.

library/Rules/DateTimeDiff.php
```diff
@@ -135,4 +127,18 @@ private function createDateTimeObject(mixed $input): ?DateTimeInterface
...
         return $dateTime;
     }
+
+    private function getTranslatedType(string $type): string
+    {
+        return match ($type) {
+            'y' => 'years',
+            'm' => 'months',
+            'd' => 'days',
+            'days' => 'full days',
```
henriquemoody, line 145: Very nice catch!

library/Rules/DateTimeDiff.php
```diff
@@ -95,7 +95,7 @@ public function evaluate(mixed $input): Result
         return (new Result($nextSibling->isValid, $input, $this, $parameters))->withNextSibling($nextSibling);
     }
 
-    private function comparisonValue(DateTimeInterface $now, DateTimeInterface $compareTo)
+    private function comparisonValue(DateTimeInterface $now, DateTimeInterface $compareTo): int|float
```
henriquemoody, line 98: Not sure if that's correct, because `days` can return false

henriquemoody:
I mixed some of my changes and yours @gvieiragoulart! Here's the merge request: #1475

I'm not sure whether we should add `full days` or not yet, it can be useful, but it doesn't seem necessary. What are your thoughts?

PR #1473, opened 2024-11-21
Improve docs adding namespace import

jomisacu:
Like an example it must be functional. So I've added the import part to be more explicit for newbies.

### DIFF
--- a/docs/02-feature-guide.md
+++ b/docs/02-feature-guide.md
@@ -72,6 +72,8 @@ $data = [
 Using the next combination of rules, we can validate child keys.
 
 ```php
+use Respect\Validation\Validator as v;
+
 v::key(
     'parentKey',
     v::key('field1', v::stringType())
@@ -93,6 +95,8 @@ For that reason all rules are mandatory now but if you want to treat a value as
 optional you can use `v::optional()` rule:
 
 ```php
+use Respect\Validation\Validator as v;
+
 v::alpha()->validate(''); // false input required
 v::alpha()->validate(null); // false input required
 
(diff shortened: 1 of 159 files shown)


alganet:
Hi @jomisacu!

We already have the instructions for namespace import right at the top of the docs:

---

image

---

I believe this introduction should be enough, and save a lot of repetition from the examples. 

Let me know if I'm missing something here! Perhaps there's another reason for adding it and I'm just not getting the idea right.

jomisacu:
> Hi @jomisacu!
> 
> We already have the instructions for namespace import right at the top of the docs:
> 
> image
> 
> I believe this introduction should be enough, and save a lot of repetition from the examples.
> 
> Let me know if I'm missing something here! Perhaps there's another reason for adding it and I'm just not getting the idea right.

Hi @alganet, thanks for the rapid response.

I've noticed the namespace import. Sincerely, I had to check the docs because the example "dosn't work". It isn't a big deal but I think it could be great to have a fully functional example for the reasons below:

- v is not a real element on the respect/validation context, using it as static class lead to confusion
- Newbies can start using the library just following the examples due them are functional

Again, it isn't a big deal, feel free to close this PR I'll be happy to contribute in other way. However I think it could be worthy.

Thanks!

henriquemoody:
It can indeed be a bit of an issue to have those examples that don't work. However, I think that adding those `use` to every docblock is not ideal ether.

You gave me an idea, though, @jomisacu! What if we ship this peace of code with the library?

```php
if (!class_exists('v') ) {
    class_alias(Respect\Validation\Validator::class, 'v');
}
```

With this, we'll always have `v` as an alias to the original class, it will make using the library much simpler.

Any thoughts?

henriquemoody:
Hi @jomisacu!

I appreciate you taking the time to make those changes, but I would like to keep the documentation a bit simpler. Besides, at this point there would be too many files for you to change.

Regardless, thank you very much for contributing this made me think about how to have better documentation and for that I'm grateful already! 🐼

jomisacu:
Hi @henriquemoody!

Thanks for your time.

PR #1485, opened 2024-12-11
Only create Length with subsequents when it's possible

henriquemoody:
Since I updated the validation engine[1], it became possible to create results with siblings. This commit changes the "Length", allowing it to create a result with a sibling when possible. That will improve the clarity of the error messages.

[1]: 238f2d506a6bd717f5bded5fb95c395d2b8f4199

### DIFF
--- a/docs/rules/Length.md
+++ b/docs/rules/Length.md
@@ -24,10 +24,31 @@ v::length(v::equals(0))->isValid(new SplPriorityQueue()); // true
 
 ### `Length::TEMPLATE_STANDARD`
 
+Used when it's possible to get the length of the input.
+
 | Mode       | Template      |
 |------------|---------------|
 | `default`  | The length of |
 | `inverted` | The length of |
+
+This template serve as message prefixes.:
+
+```php
+v::length(v::equals(3))->assert('tulip');
+// Message: The length of "tulip" must be equal to 3
+
+v::not(v::length(v::equals(4)))->assert('rose');
+// Message: The length of "rose" must not be equal to 4
+```
+
+### `Length::TEMPLATE_WRONG_TYPE`
+
+Used when it's impossible to get the length of the input.
+
+| Mode       | Template                                           |
+|------------|----------------------------------------------------|
+| `default`  | {{name}} must be a countable value or a string     |
+| `inverted` | {{name}} must not be a countable value or a string |
 
 ## Template placeholders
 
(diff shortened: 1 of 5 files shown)


tests/feature/TranslatorTest.php
```diff
@@ -16,20 +16,20 @@ function (): void {
...
-        Validator::stringType()->lengthBetween(2, 15)->phone('US')->assert(0);
+        Validator::stringType()->lengthBetween(2, 15)->phone('US')->assert([]);
     },
     <<<'FULL_MESSAGE'
-    - Todas as regras requeridas devem passar para 0
-      - 0 must be a string
-      - O comprimento de 0 deve possuir de 2 a 15 caracteres
-      - 0 deve ser um número de telefone válido para o país Estados Unidos
+    - Todas as regras requeridas devem passar para `[]`
+      - `[]` deve ser uma string
```
henriquemoody, line 30: I changed this just because it was wrong, but it doesn't have anything to do with the changes.

tests/feature/Rules/LengthTest.php
```diff
@@ -41,3 +41,18 @@
     '- The length of Cactus must be equal to 3',
     ['lengthEquals' => 'The length of Cactus must be equal to 3']
 ));
+
+test('Chained wrapped rule', expectAll(
+    fn() => v::length(v::between(5, 7)->odd())->assert([]),
```
henriquemoody, line 46: This is pretty much the essence of the changes. 

If you see the other tests, `length()` receives only 1 rule, this test receives two rules, and the two of them fail. That means that we can't simply add the prefix ("The length of") to the result of that rule because the result has two children.

When `Length` receives a rule that produces a result with children, we need to add the prefix ("The length of") to all the children of that result.

henriquemoody: @dmjohnsson23 I wrote some comments to this merge request. If you have any questions, you can just write here, I'll be happy to answer them!

docs/rules/Length.md
```diff
@@ -24,11 +24,32 @@ v::length(v::equals(0))->isValid(new SplPriorityQueue()); // true
...
+
+```php
+v::length(v::equals(3))->assert('tulip');
+// Message: The length of "tulip" must be equal to 3
+
+v::not(v::length(v::equals(4)))->assert('rose');
+// Message: The length of "rose" must not be equal to 4
+```
+
+### `Length::TEMPLATE_WRONG_TYPE`
```
henriquemoody, line 44: Because this is a new template, I had to add it to the documentation.

PR #1490, opened 2024-12-11
Create deprecated classes to keep backward compatibility

henriquemoody:
I'm not sure if I really want to go down that path. It will inflate the codebase quite a lot, and it will also be difficult to implement those.


henriquemoody:
I decided not to go down that path, it's already a lot of work the way it is 😅

PR #1498, opened 2024-12-18
Improve messages in an array format

henriquemoody:
- Name: That's what is in the placeholder, and could be used for path and id as well
- Id: Tha could be the name of the rule, just a means to identify it and be able to build templates for it on that level.
- Path: That's the real path of the rule, which will be useful when dealing with arrays or objects.


library/Result.php
```diff
@@ -39,8 +39,8 @@ public function __construct(
         public readonly Mode $mode = Mode::DEFAULT,
         ?string $name = null,
         ?string $id = null,
+        public readonly string|int|null $path = null,
```
henriquemoody, line 42: I could set the value of this property with the value of id, but then I wouldn't be able to identify those results that easily -- although I could check if the path is not the same as the id... Food for thought

library/Message/StandardFormatter.php
```diff
@@ -91,32 +93,49 @@ public function full(
...
+            $childKey = $child->path ?? $child->id;
+
+            $messages[$childKey] = $this->array(
                 $child,
                 $this->selectTemplates($child, $selectedTemplates),
                 $translator
             );
-            if (count($messages[$child->id]) !== 1) {
+
+            if ($childKey == 'each' && is_array($messages['each'])) {
```
henriquemoody, line 123: Gotta find a better way to do this. A way to do it would be to check if all the children of a rule has a path and the parent doesn't.

tests/feature/Issues/Issue1427Test.php
```diff
@@ -0,0 +1,69 @@
...
+            'arrayVal' => '`false` must be an array value',
+            'groups' => 'groups must be present',
+            'permissions' => 'permissions must be present',
+        ],
+        24 => [
+            '__root' => 'These rules must pass for `["permissions": false]`',
+            'groups' => 'groups must be present',
+            'permissions' => 'permissions must be iterable',
+        ],
+    ]
```
henriquemoody, line 68: Would it have been better if it was like what I'm describing below?

```php
[
    '_self' => [
        'each' => 'Each item in `[16: ["groups": [1, "A", 3, 4, 5], "permissions": ["perm1": true, "perm2": false, "perm3": "boom!"]], 18: false, ... ]` must be valid',
    ],
    '_children' => [
        16 => [
            '_self' => [
                'allOf' => 'These rules must pass for `["groups": [1, "A", 3, 4, 5], "permissions": ["perm1": true, "perm2": false, "perm3": "boom!"]]`',
            ],
            '_children' => [
                'groups' => [
                    '_self' => [
                        'each' => 'Each item in groups must be valid',
                    ],
                    '_children' => [
                        1 => [
                            '_self' => [
                                'intVal' => 'groups must be an integer value',
                            ],
                        ],
                    ],
                ],
                'permissions' => [
                    '_self' => [
                        'each' => 'Each item in permissions must be valid',
                    ],
                    '_children' => [
                        'perm3' => [
                            '_self' => [
                                'boolVal' => 'permissions must be a boolean value',
                            ],
                        ],
                    ],
                ],
            ],
        ],
        18 => [
            '_self' => [
                'allOf' => 'All the required rules must pass for `false`',
                'arrayVal' => '`false` must be an array value',
            ],
            '_children' => [
                'groups' => [
                    '_self' => [
                        'keyExists' => 'groups must be present',
                    ],
                ],
                'permissions' => [
                    '_self' => [
                        'keyExists' => 'permissions must be present',
                    ],
                ],
            ],
        ],
        24 => [
            '_self' => [
                'allOf' => 'These rules must pass for `["permissions": false]`',
            ],
            '_children' => [
                'groups' => [
                    '_self' => [
                        'keyExists' => 'groups must be present',
                    ],
                ],
                'permissions' => [
                    '_self' => [
                        'permissions' => 'permissions must be iterable',
                    ],
                ],
            ],
        ],
    ],
]
```

tests/feature/Issues/Issue1427Test.php
```diff
@@ -0,0 +1,93 @@
...
+                            'perm3' => [
+                                'messages' => ['boolVal' => '"boom!" must be a boolean value'],
+                            ],
+                        ],
+                    ],
+                ],
+            ],
+            18 => [
+                'messages' => ['allOf' => '`false` must pass all the rules'],
+                'details' => ['arrayVal' => '`false` must be an array value'],
```
henriquemoody, line 70: I added this key because some rules have children but they're respective to the parent validation, not a child. I thought I would add an extra key there to identify those. I think that one would always use `details` instead of `messages` when it's available because `messages` would usually be only headers of nesting validations.

tests/feature/Issues/Issue1427Test.php
```diff
@@ -0,0 +1,93 @@
...
+                            1 => [
+                                'messages' => ['intVal' => '"A" must be an integer value'],
+                            ],
+                        ],
+                    ],
+                    'permissions' => [
+                        'messages' => ['each' => 'Each item in `["perm1": true, "perm2": false, "perm3": "boom!"]` must be valid'],
+                        'children' => [
+                            'perm3' => [
+                                'messages' => ['boolVal' => '"boom!" must be a boolean value'],
```
henriquemoody, line 62: I'm not replacing the input of each rule with the key, because I think that, since the structure is already in the array, it's best to keep the original message instead, so you can actually see the value that failed.

tests/feature/Issues/Issue1427Test.php
```diff
@@ -0,0 +1,93 @@
...
+                    'groups' => [
+                        'messages' => ['keyExists' => '`["permissions": false]` must be present'],
+                    ],
+                    'permissions' => [
+                        'messages' => ['each' => '`false` must be iterable'],
+                    ],
+                ],
+            ],
+        ],
+    ]
```
henriquemoody, line 92: I gotta say, we can added a whole bunch of other things here, I'm just not sure how they can be useful. Each element could contain the keys:

* `messages`: all the messages that fails
*`details`: more details about the messages that failed
* `children`: all the children (which in turn could have all the keys I'm mentioning here
* `input`: the input that was proceeded by the rule
* `path`: the path of the rule that failed

But when I start to really think about this, I think this is not even an exception anymore, it should be an object that contains all this information and handles all of them properly. It could very well be the case.

Maybe I'm just thinking of providing this array because of how the library worked before, but I could just drop the whole array thing, and return a proper object in `Validator::validate()` that would contain all the failed messages and the extra information that is needed -- maybe I'm just conditioned to how the library worked in the past.

alganet: I can offer some perspective on what I was thinking when I made the original library.

The array thing and a lot of things (like setting properties in runtime objects) were implementation details, mostly private stuff, that could be done in whatever way fits best (I decided to go for the solution that had lass code back then).

To me, the `reportError` (or was it `reportException`?) method was part of the validatable interface that could be extended and the thing I wanted to look nice. Together with the base lib, it offered two standard ways of creating exceptions:

  - ValidationException, which is a simple single message object.
  - AbstractNestedException, which is a more elaborate object, mostly used by aggregate validators, but actually an object of its own.

AbstractNestedException had things like `findMessages` with paths back then, although the workings were slightly different. Since each validator had to had a dedicated exception, I went for an abstract instead of a concrete named object like `ValidationResults`.

The decision to use exceptions also played a role in the design of `Validatable::check`, which I think doesn't exist anymore. It was designed to catch exceptions by type early instead of running the whole thing (the decision whether to fail fast or not). This complicated the internal workings of the project, and it was a compromise I'm not sure we still need to make. It seems our users don't care about fail fast that much. 

So, you're still dealign with a lot of interlocked design decisions I made back then :D. You don't need to follow any of them if they don't make sense anymore. I would just go for whatever fullfills your vision best. That was what I did: I just moved things around until it was cohese.

henriquemoody: What you mentioned about the trees is so true... It got a bit better with the `Result` since we don't loose some information that we would have lost if we were only dealing with exceptions (since we would only have exceptions when the validation failed, not if it succeeded), but still, there's quite a bit of overhead, `Result` objects that serve no purpose but to hold the validation of its children. I'm reducing the overhead with the `Reducer` rule, but still.

As for this specific change connected to your last comment, just as you, I also thought of `getMessages()` as a useful and simple to handle HTML forms (or API requests). That's why I'm a bit reluctant to create this super structure here, it seems like when people get to the point of processing this structure, an array wouldn't be as useful anyways.

Good to hear your thoughts. It was also good to be reminded of the `findMessages()` method. I've never used it myself, but I see how useful that can be, and I would like to keep it.

I'm thinking of repurposing the `validate()` method, to return an object with failures. It could be something like this.

```php
$failures = v::key('foo', v::intVal()->positive()->greaterThan())->validate($input);
if ($failures === null) {
    echo 'No failures' . PHP_EOL;
    return;
}

// Getting all messages
echo $failures->getMessage() . PHP_EOL;
echo $failures->getFullMessage() . PHP_EOL;
echo print_r($failures->getMessages(), true) . PHP_EOL;

// Finding specific messages
echo $failures->findMessage('foo') . PHP_EOL;
echo print_r($failures->findMessages(['foo' => 'Something went wrong']), true) . PHP_EOL;
```

Thoughts?

alganet: I think it looks cool! 

I would use `$result` instead of `$failures`, and maybe do a `$result->hasPassed()` method instead of the null check. I have grown wary of optional nulls in interfaces recently :)

henriquemoody: I thought about that too, and I suggested returning the failures because I think it's no use for the user to deal with anything but the failures. Apart from that, the `evaluate()` method already returns a `Result`.

I do get what you're saying about interfaces that return nullable objects. Maybe something in-between could work:

```php

$failures = v::key('foo', v::intVal()->positive()->greaterThan())->validate($input);
if (count($failures) === 0) {
    echo 'No failures' . PHP_EOL;
    return;
}
```

It looks very similar to Symfony Validator 🙃

alganet: I like it! 🐼

PR #1501, opened 2024-12-19
Allow `Min`, `Max`, and `Each` to wrap chained rules.

dmjohnsson23:
Following the pattern set in #1485, allow, the `Min`, `Max`, and `Each` rules to elegantly support chained rules with well-formatted validation messages.

### DIFF
--- a/docs/rules/Max.md
+++ b/docs/rules/Max.md
@@ -24,12 +24,6 @@ empty, the validation will fail.
 
 ### `Max::TEMPLATE_STANDARD`
 
-| Mode       | Template                    |
-|------------|-----------------------------|
-| `default`  | As the maximum of {{name}}, |
-| `inverted` | As the maximum of {{name}}, |
-
-### `Max::TEMPLATE_NAMED`
 
 | Mode       | Template       |
 |------------|----------------|
--- a/docs/rules/Min.md
+++ b/docs/rules/Min.md
@@ -24,13 +24,6 @@ empty, the validation will fail.
 
 ### `Min::TEMPLATE_STANDARD`
 
-| Mode       | Template                      |
-|------------|-------------------------------|
-| `default`  | As the minimum from {{name}}, |
-| `inverted` | As the minimum from {{name}}, |
-
-### `Min::TEMPLATE_NAMED`
-
 | Mode       | Template         |
 |------------|------------------|
 | `default`  | The minimum from |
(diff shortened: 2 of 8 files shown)


dmjohnsson23:
I've done `Min` and `Max`. I've also removed the separate named and standard templates from these rules. I didn't see a good way to implement these in the new pattern, and also felt the language of the old standard template was somewhat difficult to read anyway.

Still looking at `Each`. I think it may require a slightly different pattern than the others. I'm looking through the codebase to get a better idea of how the formatter actually works before I tackle that one.

This also contains a small update to a few of the test helper functions to prevent issues when running tests on systems with different line endings.

dmjohnsson23:
I'm not actually sure what needs to be done for `Each`. I added a few tests as a way to evaluate the current behavior and see what needed changed, and I actually was able to get the tests to pass without making any changes to the `Each` rule itself. So, there may be something I'm still not understanding.

@henriquemoody, do the tests I added in the most recent commit reflect intended behavior, or is there something that needs changed? The only things that seemed like they might be incorrect to me are:

* The shortened error message uses the first leaf in the tree of messages, rather than the root. However, I'm assuming this is intended behavior, as the test you added for `Length` enforces the same behavior.
* In the array structure, "allOf" branches collapse down to a single string when there is only one error, rather than remaining an array. The full message does not exhibit this same behavior, so this seems like it may not be the intended behavior to me.

henriquemoody:
Hey, @dmjohnsson23 !

Thank you so much for your changes!

What I would expect from when `v::each()` is called with a single rule would be something like this:

```php
v::each(v::intType())->assert([1, 2, 3, 'B']);
// Message: All items in `[1, 2, 3, 'B']` must be an integer type
```

But that's quite complicated, and I'm not sure if I want to replicate the same behaviour from other rules, if I'm honest. I will think a bit better about that, and I will let you know. 

I created #1502 with your changes, but with atomic commits and changing the commits slightly to look like the other commits in the repo. I've noticed that your email is not configured properly, so I changed your author to `Dominick Johnson <dmjohnsson23@users.noreply.github.com>`, I hope you don't mind?

dmjohnsson23:
That's fine, thank you

henriquemoody:
Closed by #1502

Thank you so much for contributing, @dmjohnsson23!

henriquemoody:
Thank _you_! Much appreciated 🐼

PR #1507, opened 2024-12-22
Use paths to identify when a rule fails

henriquemoody:
When nested-structural validation fails, it's challenging to identify which rule failed from the main exception message. A great example is the `Issue796Test.php` file. The exception message says:

host must be a string

But you're left unsure whether it's the `host` key from the `mysql` key or the `postgresql` key.

This commit changes that behaviour by introducing the concept of "Path." The `path` represents the path that a rule has taken, and we can use it in structural rules to identify the path of an array or object.

Here's what it looks like before and after:

```diff
-host must be a string
+`.mysql.host` must be a string
```

Because paths are a specific concept, I added a dot (`.`) at the beginning of all paths when displaying them. I was inspired by the `jq` syntax. I also added backticks around paths to distinguish them from any other value.

I didn't manage to fix a test, and I skipped it instead of fixing it because I want to make changes in how we display error messages as arrays, and it will be easier to fix it then.


library/Message/StandardRenderer.php
```diff
@@ -29,15 +31,17 @@ final class StandardRenderer implements Renderer
 
     private readonly Stringifier $stringifier;
 
-    public function __construct(?Stringifier $stringifier = null)
-    {
+    public function __construct(
+        ?Stringifier $stringifier = null,
```
henriquemoody, line 35: Gotta create our own stringifier to remove the Callable one, and also create one to parse a `Path` and `Name` value objects, that will create more flexibility. Maybe even `Trans` and `Raw` value objects could be a thing.

library/Rules/Lazy.php
```diff
@@ -21,22 +21,26 @@
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 final class Lazy extends Standard
 {
-    /** @var callable(mixed): Rule */
+    /** @var callable(mixed): Rule|callable(mixed): Result */
     private $ruleCreator;
 
-    /** @param callable(mixed): Rule $ruleCreator */
+    /** @param callable(mixed): Rule|callable(mixed): Result $ruleCreator */
```
henriquemoody, line 27: Gotta move that to a different commit, it's not part of those changes, but a much needed change nonetheless

library/Transformers/DeprecatedKeyValue.php
```diff
@@ -50,15 +49,13 @@ static function ($input) use ($comparedKey, $ruleName, $baseKey) {
...
-                            $comparedKey,
-                        ));
-
-                        return $rule;
+                        return (new AlwaysInvalid())
+                            ->evaluate($baseKey)
+                            ->withPath($baseKey)
+                            ->withTemplate('{{name}} must be valid to validate {{comparedKey}}')
+                            ->withExtraParameters([
+                                'comparedKey' => $comparedKey,
```
henriquemoody, line 57:
```suggestion
                                'comparedKey' => new Path($comparedKey),
```

Once we have our own stringifier, that would be doable

tests/feature/Issues/Issue1289Test.php
```diff
@@ -45,23 +45,23 @@
...
     <<<'FULL_MESSAGE'
-    - These rules must pass for `["default": 2, "description": [], "children": ["nope"]]`
-      - Only one of these rules must pass for default
-        - default must be a string
-        - default must be a boolean
-      - description must be a string value
+    - These rules must pass for `.0`
+      - Only one of these rules must pass for `.default`
+        - `.default` must be a string
+        - `.default` must be a boolean
```
henriquemoody, line 53: Tests are failing here (and in other places as well). The problem is that it gets only the value, but not the path (`.default`). I think I need to change the path right before rendering the message, updating the path of the child with the parent, but I wonder if there are unintended consequences to that.

tests/feature/Issues/Issue1289Test.php
```diff
@@ -45,23 +45,23 @@
...
     FULL_MESSAGE,
     [
         0 => [
             '__root__' => 'These rules must pass for `["default": 2, "description": [], "children": ["nope"]]`',
             'default' => [
                 '__root__' => 'Only one of these rules must pass for default',
-                'stringType' => 'default must be a string',
-                'boolType' => 'default must be a boolean',
+                'stringType' => '`.default` must be a string',
+                'boolType' => '`.default` must be a boolean',
```
henriquemoody, line 62: Same as https://github.com/Respect/Validation/pull/1507/files#r1894786273

tests/feature/Issues/Issue1334Test.php
```diff
@@ -13,37 +13,37 @@ function (): void {
...
             [
                 ['region' => 'Oregon', 'country' => 'USA', 'other' => 123],
                 ['street' => '', 'region' => 'Oregon', 'country' => 'USA'],
                 ['street' => 123, 'region' => 'Oregon', 'country' => 'USA'],
-            ]
+            ],
         );
     },
-    'street must be present',
+    '`.0.street` must be present',
```
henriquemoody, line 25: I think that's very cool!! It will make so much easier to identify which message is failing when dealing with nested arrays/objects.

tests/feature/Issues/Issue796Test.php
```diff
@@ -40,17 +40,17 @@
                 'schema' => 'schema',
             ],
         ]),
-    'host must be a string',
+    '`.mysql.host` must be a string',
```
henriquemoody, line 43: 😍

tests/feature/Rules/EachTest.php
```diff
@@ -225,60 +225,60 @@
...
+    'First item should have been an integer',
     <<<'FULL_MESSAGE'
-    - Each item in Wrapped must be valid
-      - Wrapped must be an integer
-      - Wrapped must be an integer
-      - Wrapped must be an integer
+    - Here a sequence of items that did not pass the validation
+      - First item should have been an integer
+      - Second item should have been an integer
+      - Third item should have been an integer
```
henriquemoody, line 233: I didn't realise that those tests were wrong

tests/feature/Rules/KeyOptionalTest.php
```diff
@@ -9,70 +9,70 @@
...
 ));
 
 test('With "Not" name, inverted', expectAll(
     fn() => v::not(v::keyOptional('foo', v::intType()))->setName('Not')->assert(['foo' => 12]),
-    'foo must not be an integer',
-    '- foo must not be an integer',
-    ['foo' => 'foo must not be an integer']
+    'Not must not be an integer',
+    '- Not must not be an integer',
+    ['foo' => 'Not must not be an integer'],
```
henriquemoody, line 63: The above 3 changes are a massive bonus, but they could also be a liability. 

The problem here is if you have a series of children, they will all have the same name. This is not desirable.

henriquemoody: Even though they're not related, we should not overwrite the name of rules that have paths. Rules that have paths should have a name deliberately set. A way to go about it is probably not to update the name of children that have paths, but update the name of the parent is all cases.

tests/feature/Issues/Issue1289Test.php
henriquemoody: We should reproduce the same behavior we have in the "StandardFormatter::main" into "full" and "array".

I first thought to update the path of children in "Result::withPath", but that might be weird because it looks more like a formatting concern

tests/feature/Issues/Issue1348Test.php
```diff
@@ -24,57 +24,57 @@
             v::key('manufacturer', v::equals('Ford'))->key('model', Validator::in(['F150', 'Bronco'])),
         ),
     )->assert($cars),
-    'manufacturer must be equal to "Honda"',
+    '`.2.manufacturer` must be equal to "Honda"',
```
henriquemoody, line 27: Come to thing that it would be misleading. That's the trouble with `OneOf`

tests/feature/Rules/EachTest.php
```diff
@@ -246,11 +246,11 @@
     <<<'FULL_MESSAGE'
     - Each item in `[2, 4]` must be valid
       - All the required rules must pass for `.0`
-        - 2 must be between 5 and 7
-        - 2 must be an odd number
+        - `.0` must be between 5 and 7
+        - `.0` must be an odd number
```
henriquemoody, line 250:
```
It must be an odd number
```

Something to try out

library/Result.php
```diff
@@ -114,14 +110,25 @@ public function withIdFrom(Rule $rule): self
         return $this->clone(id: lcfirst(substr((string) strrchr($rule::class, '\\'), 1)));
     }
 
-    public function withUnchangeableId(string $id): self
+    public function withPath(string|int $path): self
     {
-        return $this->clone(id: $id, unchangeableId: true);
+        if ($this->path === $path) {
```
henriquemoody, line 115: Better convert the path to array and formar it as you want later

tests/feature/Transformers/DeprecatedKeyValueTest.php
```diff
@@ -9,60 +9,60 @@
...
     fn() => v::keyValue('foo', 'equals', 'bar')->assert(['foo' => 42]),
-    'bar must be present',
+    '`.bar` must be present',
     'The keyValue() rule has been deprecated and will be removed in the next major version. Use nested lazy() instead.',
 ));
 
 test('Scenario #8', expectMessageAndError(
     fn() => v::keyValue('foo', 'json', 'bar')->assert(['foo' => 42, 'bar' => 43]),
-    'bar must be valid to validate foo',
+    '`.bar` must be valid to validate `.foo`',
```
henriquemoody, line 54: What does that even mean?

library/Message/StandardFormatter.php
```diff
@@ -227,6 +264,11 @@ private function extractDeduplicatedChildren(Result $result): array
         $deduplicatedResults = [];
         $duplicateCounters = [];
         foreach ($result->children as $child) {
+            if ($child->path !== null) {
+                $deduplicatedResults[$child->path] = $child->isValid ? null : $child;
```
henriquemoody, line 268: This is gonna go wrong

library/Result.php
henriquemoody: I will do that another time, as long as it's working, this is fine now. I will make more changes to it later.

tests/feature/Transformers/DeprecatedKeyValueTest.php
henriquemoody: Not worrying about a deprecated rule, so I'm dropping this

PR #1511, opened 2024-12-26
Handle names via the `Named` rule

henriquemoody:
Because of how the validation engine works now [1], there's no reason to keep adding names to each rule. Instead, create a single rule that handles naming rules with a few other accessories. This change is not necessarily simple, but it shrinks the `Rule` interface, and it's more aligned with how the library works right now. 

Personally, I think this API is much more straightforward than the `setName()` method, as it's way more explicit about which rule we're naming. Because of this change, the behaviour changed slightly, but it's for the best.

Because of this change, I managed to remove a lot of code, but unfortunately, it's quite a big-bang commit. It would be too complicated to make it atomic since names are an intrinsic part of the library.

[1]: https://github.com/Respect/Validation/commit/238f2d506a6bd717f5bded5fb95c395d2b8f4199


tests/feature/Rules/PropertyTest.php
```diff
@@ -36,22 +36,22 @@
 ));
 
 test('With wrapped name, missing property', expectAll(
-    fn() => v::property('foo', v::intType()->setName('Wrapped'))->setName('Wrapper')->assert(new stdClass()),
-    'Wrapped must be present',
-    '- Wrapped must be present',
-    ['foo' => 'Wrapped must be present'],
+    fn() => v::property('foo', v::intType()->setName('Wrapped'))->assert(new stdClass()),
+    'foo must be present',
+    '- foo must be present',
```
henriquemoody, line 41: This shouldn't be the case, it's weird that I missed that

PR #1518, opened 2025-01-06
Prepare library for the next major version

henriquemoody:
Version 3.0 will include a few crucial deprecations. This commit adds some soft deprecations to warn users about these changes.

Some of the biggest changes are:

* The method `validate()` will be renamed to `isValid()`.

* The method `validate()` will be repurposed to return an object with failures.

* It won't be possible to handle rules directly; users will need to use the `Validator` class to validate with any rule.

* It won't be possible to catch specific exceptions; users will need to catch `ValidationException`.

There will some more changes, but those are some of the most important ones, and are the ones that are easy to deprecate right now.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -22,6 +22,7 @@ jobs:
           - "8.1"
           - "8.2"
           - "8.3"
+          - "8.4"
 
     steps:
       - name: Checkout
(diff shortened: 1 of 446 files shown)


henriquemoody:
This merge request got merged as branch 2.4

henriquemoody:
See https://github.com/Respect/Validation/releases/tag/2.4.0

PR #1524, opened 2025-01-17
Use PHP Domain Parser to validate domains

henriquemoody:
It isn't really working well right now, I just put it together, gotta do more work on it.


henriquemoody:
Closing this one for the moment. I don't think this library is a good fit for us, maybe in the future we can find something better, but for now let's stick to our own Rule.

PR #1532, opened 2025-03-24
Trace paths instead of IDs

henriquemoody:


tests/feature/Issues/Issue1376Test.php
```diff
@@ -28,11 +28,7 @@
         '__root__' => '`stdClass { +$author="foo" }` must pass all the rules',
         'title' => '`.title` must be present',
         'description' => '`.description` must be present',
-        'author' => [
-            '__root__' => '`.author` must pass all the rules',
-            'intType' => '`.author` must be an integer',
-            'lengthBetween' => 'The length of `.author` must be between 1 and 2',
-        ],
```
alganet: Is the intent here to pick just one of the submessages, or are we working to flatten the message tree?

Flattening would leave something like `author.intType`, `author.lengthBetween` and so on (recursively, so multiple.nested.validators.all.get.their.path), all in the root array.

This flattening behavior is similar to what `findMessages` did in the past, although `findMessages` was using a trick (the user provided the path, we just walked it).

I *also* had trouble with making unambiguous paths (without relying on the lib user to provide search paths for the errors he/she is interested in).

It is an interesting problem though.

alganet: Okay, so I see that this is just temporary dev stuff and the intention is not to flatten it out (at least not in the StandardFormatter).

@henriquemoody we should have a discussion about design here.

I'm worried about each:

```
        0 => '`.my_int` must be present',
        1 => '`.my_int` must be an odd number',
        2 => [
            '__root__' => '`.2` must pass all the rules',
            'arrayType' => '`.2` must be an array',
            'my_int' => '`.my_int` must be present',
        ],
```

It would be ideal if all the items for `Each` had the same topology. Some items report a single string (0 and 1) and others report an array (2). I'm assuming you also struggled with choices regarding being concise or easy to navigate the structure (not requiring users to do `is_array` and stuff).

From what I understood now, you don't want to flatten the entire message results, but you want that uniform topology. Maybe a single string for each item (thus limiting a little bit the detail it can deliver).

I think that if we don't collapse (0 and 1 would still be arrays with some verbosity), we can avoid having to do full path tracing. The trick would be to do it just for `Each` (the design choice is: first level of `Each` is always verbose, to simplify consuming it). In simple terms, each item in `Each` should have a root.

I could be missing something regarding templates and stuff. As I said, I need some time to get really close to 3.0 architecture. I'm learning it. If you have any insights you think could help me, please share.

henriquemoody: It's gonna be difficult to talk about this until I am back from vacation on the 15th of December.

All I can tell you from the top or my head is that after years working with findMessages() and getMessages(), and getting multiple bug reports expecting different structures, I came to realise that the best is to keep the structure as similar as the input -- and even that is kinda impossible. The idea of getMessages() is to return a structure that people can use to, let's say, attach to a `<form>`, but doing that only through arrays is impossible. The findMessages provided a much more elegant solution for that case, but it wouldn't play so well making a distinction between a single failed validation or a composite.

I can also say that `Each` is the biggest challenge with any version of Validation, IMO. So, for 3.0 my goal was to just get getMessages to return a useful structure, and figure out in future versions a mechanism for people to iterate over the failed results.

This PR is not so much about the getMessages(), though, but about getMessage() returning a path that is recognisable, right now if a nested key fails, we only see the key that fails without the context of its parents and this is what I am trying to solve here.

henriquemoody: Also, I think I might have several changes in my local branch...

alganet: I think we are aligned in what we want for `Each`: A predictable response structure. I agree that is one of the hardest validators to deal with (and we dropped key validation! This monster used to validate array keys as well, not only values).

Would you be OK with having some `if ($result->rule instanceof Each` in the StandardFormatter? Maybe an interface instead of the specific rule just to be cleaner.

If we can do that, we can flatten the `Each` result. Example with the problematic skipped test:

```
        '__root__' => 'Each item in `[["not_int": "wrong"], ["my_int": 2], "not an array"]` must be valid',
        '0.my_int' => '`.my_int` must be present',
        '1.my_int' => '`.my_int` must be an odd number',
        '2' => '`.2` must pass all the rules',
        '2.arrayType' => '`.2` must be an array',
        '2.my_int' => '`.my_int` must be present',
```

You can see that I don't have a clear "path" (mixes keys with validator names in some results). There's no one-size-fits all solution here. Some "forms" (or API error decorators) will want some things, others will want other things.

If we can find the expected array we want, we find the solution.

I'll keep thinking about it, and stop botching your vacation 🐼 We can discuss it better after you come back!

library/Message/Formatter/NestedListStringFormatter.php
```diff
@@ -58,7 +55,7 @@ private function formatRecursively(
                 $indentation,
                 $this->renderer->render(
                     $this->templateResolver->resolve(
-                        $depth > 0 ? $result->withDeepestPath() : $result,
+                        $result->asOrphan(),
```
henriquemoody: I think we need to check the `$depth` here, and perhaps create a distinction between "orphan" or "path orphan". But I'm not completely sure. We just don't want to show the name twice, I think, but the easiest would be not to care and always pass the name. The consequence is that we'd have something like:

```
- Beetlejuice must pass all the rules
  - `.mysql` (<- Beetlejuice) must pass the rules
    - `.host` (<- Beetlejuice) must be a string
  - `.postgresql` (<- Beetlejuice) must pass the rules
    - `.user` (<- Beetlejuice) must be a string
```

I'm not very happy with it, but I'm not to upset either. Ideally it would be something like this:

```
- Beetlejuice must pass all the rules
  - `.mysql` must pass the rules
    - `.host` must be a string
  - `.postgresql` must pass the rules
    - `.user` must be a string
```

What do you think, @alganet?

library/Rules/Each.php
```diff
@@ -23,12 +25,17 @@
 )]
 final class Each extends FilteredNonEmptyArray
 {
+    public function evaluate(mixed $input): Result
+    {
+        throw new SkippedWithMessageException('Remove this once other tests have been fixed.');
```
henriquemoody: I want to first figure out how to handle other failed results before do anything with `Each`, this is its own beast.

library/Rules/Named.php
```diff
@@ -10,22 +10,27 @@
...
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
 final class Named extends Wrapper implements Nameable
 {
-    public function __construct(
-        Rule $rule,
-        private readonly string $name,
-    ) {
+    private readonly Name $name;
+
+    public function __construct(Rule $rule, string|Name $name)
```
henriquemoody, line 26:
```suggestion
    public function __construct(Rule $rule, Name $name)
```

tests/feature/Issues/Issue1376Test.php
```diff
@@ -29,11 +29,7 @@
             '__root__' => '`stdClass { +$author="foo" }` must pass all the rules',
             'title' => '`.title` must be present',
             'description' => '`.description` must be present',
-            'author' => [
-                '__root__' => '`.author` must pass all the rules',
-                'intType' => '`.author` must be an integer',
-                'lengthBetween' => 'The length of `.author` must be between 1 and 2',
-            ],
+            'author' => 'The length of `.author` must be between 1 and 2',
```
henriquemoody, line 32: This is happening because we're not checking if the keys exists, we just overwrite it.

library/Message/Formatter/NestedListStringFormatter.php
alganet: The best would be to not repeat the name, but I'm not upset by it either. We can work on not repeating it in later PR perhaps!

library/Rules/Each.php
alganet: I agree!

library/Rules/Named.php
henriquemoody: Actually, it's better to keep it as is, because it makes much easier for users to use that.

library/Message/Formatter/NestedListStringFormatter.php
henriquemoody: Actually, I made it work!

tests/feature/Issues/Issue179Test.php
```diff
@@ -35,4 +35,4 @@ function (): void {
             'host' => '`.host` must be a string',
             'user' => '`.user` must be present',
         ]),
-));
+))->skip('This still needs to be fixed in order to pass.');
```
henriquemoody, line 38: @alganet I'm gonna skip these for now, I will work on that in another branch, but now we can merge this and you can work on `Each`.

Issue #1471, opened 2024-11-11
Support for UUID 6/7/8


jberns88:
Would it please be possible to add support for UUID6/7/8?

Longer term it would be cool if the lib detected Ramsey\Uuid for full validation. If I find some free time I'll try and give it ago.

Thanks

henriquemoody:
That sounds like great idea! I think it might take some time for me to implement that myself, but would you like to give it a go?

henriquemoody:
Closing it now, since we'll soon roll out version 3.0

PR #1542, opened 2025-05-22
Add support for UUID validation using ramsey/uuid to support all UUID…

srjlewis:
… versions

As suggested in https://github.com/Respect/Validation/issues/1471

Defaults to the original behavior if ramsey/uuid is not installed or explicitly disabled.

### DIFF
--- a/library/Rules/Uuid.php
+++ b/library/Rules/Uuid.php
@@ -9,8 +9,10 @@ declare(strict_types=1);
 
 namespace Respect\Validation\Rules;
 
+use Ramsey\Uuid\Uuid as RamseyUuid;
 use Respect\Validation\Exceptions\ComponentException;
 
+use function class_exists;
 use function is_string;
 use function preg_match;
 use function sprintf;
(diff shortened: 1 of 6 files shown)


library/Rules/Uuid.php
```diff
@@ -34,18 +37,39 @@ final class Uuid extends AbstractRule
...
 
     /**
      * Initializes the rule with the desired version.
      *
      * @throws ComponentException when the version is not valid
      */
-    public function __construct(?int $version = null)
+    public function __construct(?int $version = null, ?bool $useRamseyUuid = null)
     {
+        if ($useRamseyUuid && !$this->ramseyUuidIsLoaded()) {
```
henriquemoody, line 62: That's really awesome, thank you so much! I would say though, we could aways use RamseyUuid for this rule, we just need to require it to be install. In that case, this branch should be based on `main`, to be released with the next major version. What do you say?

srjlewis: I was trying not to introduce any BC breaks and allow RamseyUuid to be optional.

But if you are good with making RamseyUuid required, it would clean up the code loads.

I was targeting this at 2.4 so it could be used quicker, has I dont know how far version 3.0 is from release.

I will make the chages to make RamseyUuid a required and clean up the code 😁

srjlewis: OK, just looking at the `main` version, I will create a new version for the `main` and leave the PR here if you want to merge this into `2.4`

henriquemoody:
Closing this branch as it was already implemented on 2b364764f7a78a31ee5c601722298f47e753026d and I don't plan to add new feature to version 2.4

Again, thank you for contributing, @srjlewis!

PR #1543, opened 2025-05-23
Extend UUID rule support to versions 1 through 8

srjlewis:
Updated the `Uuid` rule to validate UUID versions 1 to 8 using the `ramsey/uuid` library, replacing manual pattern checks. Updated tests and documentation accordingly, ensuring compatibility with string and `UuidInterface` inputs.

### DIFF
--- a/library/Rules/Uuid.php
+++ b/library/Rules/Uuid.php
@@ -10,14 +10,14 @@ declare(strict_types=1);
 namespace Respect\Validation\Rules;
 
 use Attribute;
+use Ramsey\Uuid\Uuid as RamseyUuid;
+use Ramsey\Uuid\UuidInterface;
 use Respect\Validation\Exceptions\InvalidRuleConstructorException;
 use Respect\Validation\Message\Template;
 use Respect\Validation\Result;
 use Respect\Validation\Rules\Core\Standard;
 
 use function is_string;
-use function preg_match;
-use function sprintf;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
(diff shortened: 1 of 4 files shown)


srjlewis:
@henriquemoody I have done the same chages as you suggest, but change the test to stringable to help with other UUID libs, it is just a idea so I dont mind if it is not merged.

library/Rules/Uuid.php
```diff
@@ -34,41 +37,44 @@ final class Uuid extends Standard
...
+            $uuid = RamseyUuid::fromString((string) $input);
+        } catch (Throwable) {
+            return Result::failed($input, $this, $parameters, $template);
+        }
 
-    private function isSupportedVersion(int $version): bool
-    {
-        return $version >= 1 && $version <= 5 && $version !== 2;
+        /** @phpstan-ignore-next-line */
+        $uuidVersion = $uuid->getFields()->getVersion();
```
henriquemoody, line 70: I'm a bit concerned about ignoring this part here, because that means that `RamseyUuid::fromString` might return something that's not what we expecting. Isn't there a programatic way to verify this?

library/Rules/Uuid.php
```diff
@@ -34,41 +37,44 @@ final class Uuid extends Standard
...
-        if (!is_string($input)) {
+
+        if (!is_string($input) && !($input instanceof UuidInterface)) {
             return Result::failed($input, $this, $parameters, $template);
         }
 
-        return new Result(preg_match($this->getPattern(), $input) > 0, $input, $this, $parameters, $template);
-    }
+        try {
+            $uuid = RamseyUuid::fromString((string) $input);
```
henriquemoody, line 64:
```suggestion
            $uuid = $input instanceof UuidInterface ? $input : RamseyUuid::fromString($input);
```

henriquemoody: `$input` will either be a string or a `UuidInterface`, it's best not to convert to a string if that's already what we're expecting

henriquemoody (changes requested):

library/Rules/Uuid.php
srjlewis: that is my mistake I was ment to remove it, well spotted

library/Rules/Uuid.php
henriquemoody: I think that solution you propose is much better than ignoring the problem! Awesome!

PR #1544, opened 2025-05-27
Extend UUID rule support to versions 1 through 8

henriquemoody:
Updated the `Uuid` rule to validate UUID versions 1 to 8 using the `ramsey/uuid` library, replacing manual pattern checks. Updated tests and documentation accordingly, ensuring compatibility with string and UuidInterface` inputs.

Duplicates #1543


henriquemoody:
Hi @srjlewis,

Thank you for https://github.com/Respect/Validation/pull/1543!

I wanted to make a few changes to it, and I thought it would be a bit easier to implement them instead of requesting changes.

I hope you don't mind, I just amended your commit.

library/Rules/Uuid.php
```diff
@@ -34,14 +37,16 @@ final class Uuid extends Standard
 {
     public const TEMPLATE_VERSION = '__version__';
 
-    private const PATTERN_FORMAT = '/^[0-9a-f]{8}-[0-9a-f]{4}-%s[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i';
-
     public function __construct(
         private readonly ?int $version = null
     ) {
+        if (!class_exists(RamseyUuid::class)) {
```
henriquemoody, line 43: This is the biggest changes, and it was my miscommunication before. I want this rule to require `ramsey/uuid`, but not the library to do so.

library/Rules/Uuid.php
```diff
@@ -51,24 +56,27 @@ public function evaluate(mixed $input): Result
...
+
+        if (!is_string($input) && !($input instanceof UuidInterface)) {
             return Result::failed($input, $this, $parameters, $template);
         }
 
-        return new Result(preg_match($this->getPattern(), $input) > 0, $input, $this, $parameters, $template);
-    }
+        if (is_string($input)) {
+            try {
+                $input = RamseyUuid::fromString($input);
```
henriquemoody, line 66: This allows us to interpret non-formatted UUIDs as valid. I _think_ that's a good thing, but I would love to hear your thoughts

tests/unit/Rules/UuidTest.php
```diff
@@ -25,27 +29,24 @@
...
-
-    #[Test]
-    public function itShouldThrowExceptionWhenVersionIsTwo(): void
-    {
-        self::expectException(InvalidRuleConstructorException::class);
-        self::expectExceptionMessage('Only versions 1, 3, 4, and 5 are supported: 2 given');
-
-        new Uuid(2);
-    }
+    private const ALL_VERSIONS = [
```
henriquemoody, line 32: I changed how the tests were structure, so we don't need to use introspection to run tham.

PR #1546, opened 2025-07-14
chore: Mark all rules as attributes

matapatos:
### Overview

Marks all Respect rules as attributes

### Why?

This way we have type-hint support when using some of these rules as attributes. See this https://github.com/Attributes-PHP/validation

### DIFF
--- a/library/Rules/AllOf.php
+++ b/library/Rules/AllOf.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 namespace Respect\Validation\Rules;
 
+use Attribute;
 use Respect\Validation\Exceptions\AllOfException;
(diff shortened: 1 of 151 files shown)


henriquemoody:
Actually, this has been already implemented in the main branch: https://github.com/Respect/Validation/blob/main/docs/rules/Attributes.md

matapatos:
That's really cool! Is there an expected release date for v3?

henriquemoody:
I was hoping I would have released it at the beginning of this year... that being said the answer is as soon as I can find time to dedicate to that. Sorry for not being more clear about the timelines, it's tough being the only active maintainer

PR #1551, opened 2025-10-31
Reset missing date information in CanValidateDateTime

davidespagnoli:
missing date information are now reset to 0 (1970-01-01T00:00:00), because otherwise they are taken from current date/time (please refer to https://www.php.net/manual/en/datetimeimmutable.createfromformat.php), so: 

Respect\Validation\Validator::date("m")->isValid('06') 

is valid if it's run on October 30
is not valid if it's run on October, 31


henriquemoody:
I'm sorry for taking so long to react! I was hyper-focusing on version 3.0 😅

I merged this manually via [fc7130d2e790978ba2db857493bf01ecd44af2ce](https://github.com/Respect/Validation/commit/93edad4b3272103ea31a5186970c1bd4bdd58b3e)

Thank you for contributing! 🐼

PR #1552, opened 2025-11-04
Update documentation to prepare for release v3

henriquemoody:


.editorconfig
```diff
@@ -0,0 +1,29 @@
...
+[*.md]
+trim_trailing_whitespace = false
+max_line_length = 120
+
+# YAML files
+[*.yml]
+indent_size = 2
+
+[*.yaml]
+indent_size = 2
```
alganet, line 25: The global rule already sets `indent_size = 2`

.specify/memory/constitution.md
```diff
@@ -0,0 +1,218 @@
...
+  - MUST validate real-world usage patterns
+  - MUST test error message rendering and exception handling
+
+**Red-Green-Refactor Cycle**: Tests MUST fail initially, then pass after implementation. No code ships without passing tests.
+
+### II. Quality Assurance Pipeline
+
+**Mandatory Pre-Commit Checks**: All changes MUST pass the complete QA pipeline (`composer qa`) before merge:
+
+1. **docheader**: Verify license headers on all source files
```
alganet, line 56: We should probably use `reuse` for that, and abandon `docheader` which is not REUSE or SPDX compliant.

https://pypi.org/project/reuse/
https://reuse.software/

`reuse lint` can check the headers, and `reuse annotate` can add them to files that are missing.

.specify/memory/constitution.md
```diff
@@ -0,0 +1,218 @@
...
+
+**Strict Typing Mandate**: Every PHP file MUST declare `strict_types=1`. Type hints are REQUIRED for all parameters and return types.
+
+**Encapsulation Defaults**:
+- Classes MUST be `final` unless explicitly designed for inheritance
+- Properties MUST be `private` unless extension is required
+- Methods MUST have explicit visibility modifiers
+
+**Documentation Requirements**:
+- License header MUST be present (enforced by docheader)
```
alganet, line 74: By aiming for REUSE compliance, we can ommit this redundant part. It will annotate and lint documentation, config files, everything.

.specify/memory/constitution.md
```diff
@@ -0,0 +1,218 @@
...
+7. Approved by maintainer
+
+**Performance Standards**:
+- Rules MUST execute efficiently for typical validation workloads
+- Avoid I/O operations within rule validation logic when possible
+- Use dependency injection for external dependencies (e.g., email validators, phone validators)
+
+## Commit Message Standards
+
+**Format**: Descriptive imperative mood without conventional commit prefixes
```
alganet, line 147: WDYT of adopting the 50/72 rule?

.specify/memory/constitution.md
```diff
@@ -0,0 +1,218 @@
...
+The `path` represents the path that a rule has taken, and we can use it
+in structural rules to identify the path of an array or object.
+```
+
+**Guidelines**:
+- First line: Concise, imperative mood (e.g., "Add", "Fix", "Improve", "Remove")
+- Body: Explain the problem, solution, and rationale
+- Include examples or before/after comparisons when helpful
+- Reference issue numbers when applicable
+- No scope prefixes (no `feat:`, `fix:`, etc.)
```
alganet, line 193: In conventional commits, scope is the part in parenthesis:

`type(this is a scope): this is a message`.

By saying "no scope prefixes", we might be introducing confusing terminology (is it about the prefix or the scope in the prefix?).

I think we should just say "No prefixes (no `feat:`, no `[FEATURE]`)" and so on.

.specify/scripts/bash/create-new-feature.sh
```diff
@@ -0,0 +1,260 @@
...
+            echo "Options:"
+            echo "  --json              Output in JSON format"
+            echo "  --short-name <name> Provide a custom short name (2-4 words) for the branch"
+            echo "  --number N          Specify branch number manually (overrides auto-detection)"
+            echo "  --help, -h          Show this help message"
+            echo ""
+            echo "Examples:"
+            echo "  $0 'Add user authentication system' --short-name 'user-auth'"
+            echo "  $0 'Implement OAuth2 integration for API' --number 5"
+            exit 0
```
alganet, line 55: Inconsistency: `check-prerequisites.sh` uses HEREDOC to dump usage info, here it's one echo per line. Let's choose one way of doing it and stick with it.

.specify/scripts/bash/setup-plan.sh
```diff
@@ -0,0 +1,61 @@
...
+for arg in "$@"; do
+    case "$arg" in
+        --json) 
+            JSON_MODE=true 
+            ;;
+        --help|-h) 
+            echo "Usage: $0 [--json]"
+            echo "  --json    Output results in JSON format"
+            echo "  --help    Show this help message"
+            exit 0
```
alganet, line 18: HEREDOC

henriquemoody:
I was just playing with SpecKit, but I didn't do a good job with it. I just created the pull request to keep it here, see what I could salvage from it. Definitely too much stuff going on here!

alganet:
@henriquemoody cool. Need any help with v3? I don't want to step in your toes, so I thought I'd better ask.

henriquemoody:
There are absolutely no toes to step on! I'd be thrilled to see you back on Validation.

I'm on vacation until the 15th of December, and I don't have my laptop on me, but the most important thing I would like to get done is #1532. Basically, I want to track paths instead of IDs, because that will allow us to have a better error message. Right now when validating nested arrays, we can't really know what failed from a single message -- tracking paths will allow us to do that. I also wanted to use an object "Path" to do that, because it's easier to handle it, and we can have a specific formatter to handle that object, allowing us to have a distinction from when the {{name}} is actually a name or a path.

I haven't gotten that to work. First of all, from my playing around with it, I cannot make it an incremental change, so I gave up on not making a big bang. Secondly, I couldn't make it work in all cases, because the paths have gaps between them and building that correctly is very difficult.

If you want to play around with it I would be very happy, but honestly, just do whatever you're interested on doing, really. I just really want to get 3.0 released, and again no toes to step on -- besides, this library carries more your name than mine anyways!

alganet:
@henriquemoody I can see the tests are passing, but I have some doubts about the intent of the change and what you are trying to achieve.

I left some comments in the PR!

If I understand this correctly, this change is indeed big, and it will take me some time to understand how the current architecture works before I make any meaningful progress. This problem is hard.

In the past, I remember using `SplObjectStorage` to leverage the fact that it can have *objects as keys* (in contrast to standard PHP arrays). In my mind, having an object (such as `Path` or something like it) as the key for the messages, would allow us to do things like `->path` or `->id` to it, or choose a default serialization when exporting to plain `->toArray()`. That approach comes with its own drawbacks though (can't use PHP functions like `array_map` on `SplObjectStorage`).

.specify/memory/constitution.md
henriquemoody: I tend to follow something something like that! But I'm so sure whether we should stick to 50/70. I think we can extend the size of the subject line, lots of people are doing that nowadays.


We can automate that with `gitlint`

.specify/memory/constitution.md
henriquemoody: I prefer that. I went to conventional commits because it's something folks use these days, but I'm not a fan, to be honest.

PR #1554, opened 2025-11-25
Fix PHP 8.5 deprecations on branch 2.4

mpesari:
Hi! This fixes some `E_DEPRECATED` issues when running unit tests on PHP 8.5.

The fixed code doesn't exist anymore on branch `main`, so I'm not sure what the current status is with that.

## Before

```
$ php8.5 vendor/bin/phpunit
...
PHP Deprecated:  Method ReflectionProperty::setAccessible() is deprecated since 8.5, as it has no effect in /home/mpesari/git/Respect/Validation/tests/unit/Rules/EmailTest.php on line 102
...
PHP Warning:  The float-string "11111111111111111111100000000000" is not representable as an int, cast occurred in /home/mpesari/git/Respect/Validation/library/Rules/Ip.php on line 109
...
PHP Deprecated:  Method ReflectionProperty::setAccessible() is deprecated since 8.5, as it has no effect in /home/mpesari/git/Respect/Validation/library/Factory.php on line 249
...
PHP Deprecated:  Method SplObjectStorage::attach() is deprecated since 8.5, use method SplObjectStorage::offsetSet() instead in /home/mpesari/git/Respect/Validation/library/Exceptions/NestedValidationException.php on line 116
...
```

## After

```
$ php8.5 vendor/bin/phpunit
...
PHP Warning:  The float-string "11111111111111111111100000000000" is not representable as an int, cast occurred in /home/mpesari/git/Respect/Validation/library/Rules/Ip.php on line 109
...
```

I'm not sure if that last Warning is something that can be fixed easily, and it remains on the `main` branch.

Whoops, there's still another error:

```
There was 1 error:                                    
                                                                                                                                                                                                                        
1) Respect\Validation\Rules\FactorTest::shouldValidateInvalidInput with data set "mt_rand is not factor PHP_INT_MAX + 1" (Respect\Validation\Rules\Factor Object (...), 9.223372036854776E+18)
The float 9.223372036854776E+18 is not representable as an int, cast occurred
                                                                                                                                                                                                                        
/home/mpesari/git/Respect/Validation/library/Rules/Factor.php:50
/home/mpesari/git/Respect/Validation/tests/library/RuleTestCase.php:117                                                                                                                                                 
/home/mpesari/git/Respect/Validation/tests/library/RuleTestCase.php:80                
```

### DIFF
--- a/tests/unit/Rules/FactorTest.php
+++ b/tests/unit/Rules/FactorTest.php
@@ -18,6 +18,7 @@ use function mt_rand;
 use function uniqid;
 
 use const PHP_INT_MAX;
+use const PHP_VERSION_ID;
 
 /**
  * @group rule
@@ -31,6 +32,32 @@ use const PHP_INT_MAX;
  */
 final class FactorTest extends RuleTestCase
 {
+    /**
+     * Some edge cases emit E_DEPRECATED on PHP >= 8.5.
+     *
+     * @test
+     *
+     * @dataProvider providerForInvalidInputEdgeCases
+     */
+    public function shouldFailOnInvalidInputEdgeCases(Factor $rule, mixed $input): void
+    {
+        if (PHP_VERSION_ID >= 80500) {
+            $this->markTestSkipped('This test emits E_DEPRECATED on PHP >= 8.5');
+        }
+
+        self::assertInvalidInput($rule, $input);
+    }
+
+    /**
+     * @return array<string, array{0:Factor, 1:mixed}>
+     */
+    public static function providerForInvalidInputEdgeCases(): array
+    {
+        return [
+            'mt_rand is not factor PHP_INT_MAX + 1' => [new Factor(mt_rand()), PHP_INT_MAX + 1],
+        ];
+    }
+
     /**
      * {@inheritDoc}
      */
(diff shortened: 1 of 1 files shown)


alganet:
Hi! Thanks for contributing 🐼 

It seems 8.5 got more strict with types, which is good.

The `main` branch is supposed to be 3.0, which is a big rewrite (I'm behind on following it as well!). It's taking a while, so maybe the 2.x branch will still have a release, and we should have 8.5 compat in mind.

What do you think of isolating the problematic assertion that made the test fail, and mark them as skipped tests? That way, your PR will have a clean build.

mpesari:
I added a skip for the failing test on PHP 8.5.

There are still other warnings that should probably be fixed, but they don't fail the tests (although I think they _should_, because PHPUnit is configured to `convertWarningsToExceptions`)

Looks like the `uopz` extension is fixed in https://github.com/krakjoe/uopz/commits/master/ but there's not a release compatible with PHP 8.5 yet in [PECL](https://pecl.php.net/package/uopz).

According to `shivammathur/setup-php` docs it's possible to [compile the extension from git](https://github.com/shivammathur/setup-php/wiki/Add-extension-from-source)...

alganet:
@mpesari thank you!

The `uopz` fix might be too much, and it's really not a PHP 8.5 thing but a transitive dependency thing.

To me, this is good as it is. I'll merge it.

PR #1565, opened 2025-12-22
Update how we handle templates

henriquemoody:
Currently, the templates that a user provides when running `assert()` can significantly impact how the message is displayed. Because of this, the formatters become complex as they all need to handle similar conditions to format results.

This commit changes this behaviour, letting only the `InterpolationRenderer` handle the templates. This makes the code simpler and allows people to use the `InterpolationRenderer` directly, without needing to figure out how to handle templates. Thinking about it further, I believe handling templates is a concern for the `Renderer` anyway, and this will open the way to other improvements using the renderer.

I also removed the exception that is thrown when the template is not a string, because I think that after validation has failed, we should not throw any other exceptions, as that could cause unexpected errors for users.


tests/feature/AssertWithKeysTest.php
```diff
@@ -41,14 +41,14 @@
...
             - the given data must pass all the rules
               - `.mysql` must pass all the rules
-                - `.host` must be a string
-                - `.user` must be present
-                - `.password` must be present
-                - `.schema` must be a string
+                - `.mysql.host` must be a string
+                - `.mysql.user` must be present
+                - `.mysql.password` must be present
+                - `.mysql.schema` must be a string
```
henriquemoody, line 47: @alganet this is happening because if I render the `Result` with an orphan path, we cannot trace the template. This change makes the formatters simpler, but it comes with this cost.

Alternatively, we could have a specific `Renderer` that always renders the result as an orphan, but we'll need to recompose the `InterpolationRenderer`. That's something we could do later, I think. But, this format doesn't bother me, in fact, it adds some sort of clarity.

What do you think?

tests/feature/Rules/EachTest.php
```diff
@@ -198,19 +208,27 @@
         ->assert([1, 2, 3]),
     fn(string $message, string $fullMessage, array $messages) => expect()
         ->and($message)->toBe('All items should not have been integers')
-        ->and($fullMessage)->toBe('- All items should not have been integers')
-        ->and($messages)->toBe(['notEach' => 'All items should not have been integers']),
+        ->and($fullMessage)->toBe(<<<'FULL_MESSAGE'
+            - All items should not have been integers
+              - `.0` must not be an integer
+              - `.1` must not be an integer
+              - `.2` must not be an integer
```
henriquemoody, line 215: @alganet This is also a consequence of not allowing templates to influence how messages are displayed.

This could be reverted to how it used to work with the changes below, but I think this is actually good to keep it as is. If the user wants a single messages, they should just use the `getMessage()` instead.

henriquemoody:
```diff
--- a/library/Message/Formatter/NestedListStringFormatter.php
+++ b/library/Message/Formatter/NestedListStringFormatter.php
@@ -47,17 +47,20 @@ final readonly class NestedListStringFormatter implements StringFormatter
             $depth++;
         }

-        foreach ($result->children as $child) {
-            $formatted .= $this->formatRecursively(
-                $displayedName === $child->name ? $child->withoutName() : $child,
-                $renderer,
-                $templates,
-                $depth,
-                ...array_filter($result->children, static fn(Result $sibling) => $sibling !== $child),
-            );
-            $formatted .= PHP_EOL;
+        if (!$result->hasCustomTemplate()) {
+            foreach ($result->children as $child) {
+                $formatted .= $this->formatRecursively(
+                    $displayedName === $child->name ? $child->withoutName() : $child,
+                    $renderer,
+                    $templates,
+                    $depth,
+                    ...array_filter($result->children, static fn(Result $sibling) => $sibling !== $child),
+                );
+                $formatted .= PHP_EOL;
+            }
         }

+
         return rtrim($formatted, PHP_EOL);
     }
```

library/Message/Formatter/TemplateResolver.php
```diff
@@ -9,80 +9,102 @@
...
-            if ($key !== null && isset($templates[$key]) && is_array($templates[$key])) {
-                return $templates[$key];
-            }
+        $nodes[] = $path->value;
+        if ($path->parent !== null) {
+            return $this->getNodes($path->parent, $nodes);
         }
 
-        return $templates;
+        return $nodes;
```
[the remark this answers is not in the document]
henriquemoody: This is not true, because we always include at least the `Path` given to the `getNodes()` method, so the return will never be an empty array.

tests/feature/AssertWithKeysTest.php
alganet: Yep, let's full commit to the idea of using full paths from now on (especially when it simplifies things). I'm good with it.

PR #1568, opened 2025-12-26
Do not overwrite existing names

henriquemoody:
This commit addresses the skipped tests by modifying certain core concepts in the library. The way names work has always been somewhat confusing, even to me. I’ve established that existing names will never be overwritten, and if a path is already defined, we will change the name to also include the path.

I’m not very happy with how I’m solving this problem, because the `FirstResultStringFormatter` is changing some behaviour in the results that seems to be something that should always happen before. But, for the moment, I will keep it as is.

### DIFF
--- a/docs/rules/KeyExists.md
+++ b/docs/rules/KeyExists.md
@@ -35,6 +35,27 @@ v::keyExists(5)->isValid(new ArrayObject(['a', 'b', 'c'])); // false
 |-------------|------------------------------------------------------------------|
 | `name`      | The validated input or the custom validator name (if specified). |
 
+## Caveats
+
+`KeyExists` defines the given `$key` as the path, and because it is a standalone rule without children, it's not possible to display a fully custom name with it.
+
+When no custom name is set, the path is displayed as `{{name}}`. When a custom name is set, the validation engine prepends the path to the custom name:
+
+```php
+v::keyExists('foo')->assert([]);
+// Message: `.foo` must be present
+
+v::keyExists('foo')->setName('Custom name')->assert([]);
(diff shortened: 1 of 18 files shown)


docs/rules/KeyExists.md
```diff
@@ -35,6 +35,27 @@ v::keyExists(5)->isValid(new ArrayObject(['a', 'b', 'c'])); // false
...
+
+```php
+v::keyExists('foo')->assert([]);
+// Message: `.foo` must be present
+
+v::keyExists('foo')->setName('Custom name')->assert([]);
+// Message: `.foo` (<- Custom name) must be present
+```
+
+If you want to display only a custom name while checking if a key exists, use [Key](Key.md) with [AlwaysValid](AlwaysValid.md):
```
alganet, line 52: I'm wondering if we shouldn't rename `Key` and `Property` to `KeyValue` and `PropertyValue`, since now they require the value to be present.

Also, while the text is informative, it might be difficult for users to understand when to pick `Key`, `KeyExists` and `KeyValue` just from reading the description.

I was thinking maybe we should do a separate markdown file with some sort of direct comparison (maybe even a table?) of these very similar rules (Key, KeyExists, KeyOptional, Property, PropertyExists, PropertyOptional).

Hopefully, by comparing them directly in a single document, we can better show their differences to users. We can then do a small link here to that doc file (same with Property).

WDYT?

henriquemoody: I'm not a big fun of renaming the rule because I find it would be too lengthy.

I do agree that users will need more information on those structural rules and I like the extra documentation and the table you suggested!

PR #1569, opened 2025-12-26
Replace placeholder "name" with "subject"

henriquemoody:
The `{{name}}` placeholder could represent different things depending on the state of the Result, and referring to it as `{{name}}` seems arbitrary. This commit changes it to `{{subject}}`, which is much more generic and it describes well what that placeholder can mean.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,20 +16,21 @@ learn that just by looking at other rules. Pick the simple ones like `ArrayType`
 to begin.
 
 Before writing anything, feature or bug fix:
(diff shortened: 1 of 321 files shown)


alganet:
Hey @henriquemoody, can I ask Copilot to do a review here? It is often good at catching possible small details in large diff PRs.

henriquemoody:
Good idea, @alganet! What do you think about this renaming? It's quite a big change

alganet:
@henriquemoody I like it, `subject` is a better name!

PR #1572, opened 2025-12-26
Ensure formatters only format, not modify results

henriquemoody:
When I changed the library to not overwrite existing names [1](https://github.com/Respect/Validation/commit/8332d28acce8c132331baaabc220215a76c6a28c), I wasn't happy with how `FirstResultStringFormatter` was changing the results, because the results should be completely ready by the time they arrive in the formatters.

This commit changes that behaviour, ensuring the results are complete with all necessary information by the time they reach the formatters.

Along with those changes, I refactored some stringifiers and simplified the `InterpolationRenderer`; we were not overwriting the "name" parameter anyway, as it was just an unnecessary overhead.

### DIFF
--- a/library/Result.php
+++ b/library/Result.php
@@ -30,6 +30,7 @@ final readonly class Result
         public array $parameters = [],
         public string $template = Rule::TEMPLATE_STANDARD,
         public bool $hasInvertedMode = false,
+        public bool $hasPrecedentName = true,
         public Name|null $name = null,
         public Result|null $adjacent = null,
         public Path|null $path = null,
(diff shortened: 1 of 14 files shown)


library/Result.php
```diff
@@ -133,40 +122,30 @@ public function withPath(Path $path): self
...
 
     public function withChildren(Result ...$children): self
     {
-        if ($this->path === null) {
-            return clone($this, ['children' => $children]);
-        }
-
-        return clone($this, ['children' => array_map(fn(Result $child) => $child->withPath($this->path), $children)]);
+        return clone($this, [
+            'children' => array_map(fn(Result $child) => $child->withSubject($this->subject), $children)
```
henriquemoody: maybe too drastic?

PR #1578, opened 2025-12-29
Create `All` rule and `all` prefix

henriquemoody:
This rule behaves similarly to ``Each, but the difference is that it presents the messages in a simpler manner. This allows users to have a simpler message when validating all items in an array or iterable value.

I’m also introducing the `all` prefix, making it easier for users to apply the `All` rule in combination with other rules.

### DIFF
--- a/docs/09-list-of-rules-by-category.md
+++ b/docs/09-list-of-rules-by-category.md
@@ -43,6 +43,7 @@
 
 ## Comparisons
 
+- [All](rules/All.md)
 - [Between](rules/Between.md)
 - [BetweenExclusive](rules/BetweenExclusive.md)
 - [Equals](rules/Equals.md)
@@ -270,6 +271,7 @@
 
 ## Transformations
 
+- [All](rules/All.md)
 - [Call](rules/Call.md)
 - [Each](rules/Each.md)
 - [Length](rules/Length.md)
@@ -302,6 +304,7 @@
 
 ## Alphabetically
 
+- [All](rules/All.md)
 - [AllOf](rules/AllOf.md)
 - [Alnum](rules/Alnum.md)
 - [Alpha](rules/Alpha.md)
--- a/docs/rules/All.md
+++ b/docs/rules/All.md
@@ -1,0 +1,55 @@
+# All
+
+- `All(Rule $rule)`
+
+Validates all items of the input against a given rule.
+
+```php
+v::all(v::intType())->isValid([1, 2, 3]); // true
+v::all(v::intType())->isValid([1, 2, '3']); // false
+```
+
+This rule is similar to [Each](Each.md), but as opposed to the former, it displays a single message when asserting an input.
+
+## Note
+
+This rule uses [Length](Length.md) with [GreaterThan][GreaterThan.md] internally. If an input has no items, the validation will fail.
+
+## Templates
+
+### `All::TEMPLATE_STANDARD`
+
+| Mode       | Template      |
+| ---------- | ------------- |
+| `default`  | Every item in |
+| `inverted` | Every item in |
+
+This template serve as message prefixes.:
+
+```php
+v::all(v::floatType())->assert([1.5, 2]);
+// Message: Every item in `[1.5, 2]` must be float
+
+v::not(v::all(v::intType()))->assert([1, 2, -3]);
+// Message: Every item in `[1, 2, -3]` must not be an integer
+```
+
+## Categorization
+
+- Comparisons
+- Transformations
+
+## Changelog
+
+| Version | Description |
+| ------: | ----------- |
+|   3.0.0 | Created     |
+
+---
+
+See also:
+
+- [Each](Each.md)
+- [Length](Length.md)
+- [Max](Max.md)
+- [Min](Min.md)
--- a/library/Rules/All.php
+++ b/library/Rules/All.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * Copyright (c) Alexandre Gomes Gaigalas <alganet@gmail.com>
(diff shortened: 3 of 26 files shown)


library/Rules/All.php
```diff
@@ -0,0 +1,39 @@
...
+        $hasPassed = true;
+        foreach ($input as $value) {
+            $result = $this->rule->evaluate($value);
+            if ($result->hasPassed === false) {
+                $hasPassed = false;
+                break;
+            }
+        }
+
+        return $result->asAdjacentOf(
```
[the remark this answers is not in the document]
henriquemoody: I will ignore this one, since I'm sure there will be at least one item here. But it's a fair point, and perhaps we should reconsider the resign.

docs/rules/All.md
```diff
@@ -0,0 +1,55 @@
...
+| `inverted` | All items in |
+
+This template serve as message prefixes.:
+
+```php
+v::all(v::floatType())->assert([1.5, 2]);
+// Message: All items in `[1.5, 2]` must be float
+
+v::not(v::all(v::intType()))->assert([1, 2, -3]);
+// Message: All items in `[1, 2, -3]` must not be an integer
```
alganet: This should be something like:

> Message: All items in `[1, 2, -3]` must not be integers

However, that would be impractical for how we build messages (would imply adding plural forms to all rules).

We can get around this by using something like:

> Message: Every item in `[1, 2, -3]` must not be an integer

This way, we use "every item" (singular) as the subject, instead of "all items" (plural) and we can reuse the messages without introducing a new form.

docs/09-list-of-rules-by-category.md
```diff
@@ -302,6 +304,7 @@
 
 ## Alphabetically
 
+- [All](rules/All.md)
```
alganet, line 307: I'm worried about introducing a confusion between `AllOf` and `All`.

They have very similar names, but very different behaviors. While `AllOf` operates on rules (the name refers to rules that must pass), `All` operates on input (the name refers to items that must pass a single rule).

The `Each` name suffers from a similar confusion  (is "each" referring to rules or input items? It's not clear just from the name).

*In the current state of naming*, I would be more comfortable with names like these:

 - `Collection` instead of `All`: it says that we're going to validate the input as a single collection, hinting that individual messages for each item are not relevant, the "whole collection" is.
 - `CollectionMap` instead of `Each`: it says we're doing a "map" on each item of the Collection, and therefore getting more detailed results.

This also puts these two together in the alphabetical list and closer in auto-completion actions, making it easier for developers.

The `all` name does sound better in chains. For example, `allPositive` is much more intuitive than `collectionPositive`. Perhaps by using `CollectionOf` instead would lead to something like `collectionOfPositive`, which is more verbose.

What we're dealing here is a category conflict. Rules such `AllOf` and `AnyOf` are referring to rules themselves and `Each` is referring to items. It's hard to be consistent, because we need to be able to target both. I considered using `LogicAnd` and `LogicOr` for those in the past. This has the benefit of being more explicit, but it results to a less readable chain (the names "all of" and "one of" could be used in english sentences, but "logic and" and "logic or" are more geeky).

I think the `Logic` prefix makes more sense. The rules that do that (AllOf, AnyOf, NoneOf, When) are the exception (they are talking about the rules, not the input). So they should get a special prefix instead, and all other rules are assuming to be talking about the input instead.

 - AllOf -> LogicAnd
 - AnyOf -> LogicOr
 - NoneOf -> LogicNand

TL;DR: the `All` name is the best, but `AllOf` is a problematic legacy naming. We need to decide, one of those categories of validators will get more verbose.

docs/rules/All.md
henriquemoody: That's a very nice catch! Indeed, the "All items in" will make messages grammatically incorrect.

docs/09-list-of-rules-by-category.md
henriquemoody: Excellent point, perhaps would be a good idea to rename those rules before we release 3.0.

I think it makes sense to keep this one as "all", it's quite intuitive, and then we could do something like this:

Old | New
----|----
AllOf | LogicAnd
AnyOf | LogicOr
OneOf | LogicXor
NoneOf | LogicNor

alganet: Oops, I previously associated NoneOf with LogicNand, but you're right, it's LogicNor!

Also, `When` could be `LogicTernary`, it also belongs to that family somehow.

There's an argument for making `Not` into `LogicNot` as well, but I would make an exception to that one for brevity.

henriquemoody: I'm not so sure about the LogicTernary, it looks very verbose even though it makes sense.

About "Not" I think it's more like "Inversion" than "Not", but brevity indeed wins there

alganet (dismissed):

docs/09-list-of-rules-by-category.md
henriquemoody: I was thinking about how this would like in the chain. Wouldn't "group" make more sense in the chain?


```php
v::oneOf(v::intType(), v::negative())->assert('string');

// vs

v::logicXor(v::intType(), v::negative())->assert('string');


// vs

v::groupXor(v::intType(), v::negative())->assert('string');

// We could also do that, via alias
v::xor(v::intType(), v::negative())->assert('string');
```

Old | New
----|----
AllOf | GroupAnd
AnyOf | GroupOr
OneOf | GroupXor
NoneOf | GroupNor

henriquemoody: Or maybe `grouped*()`

PR #1580, opened 2025-12-29
Fix date overflow when parsing year-month format without day

AmineMx:
When parsing date formats like "2026-02" that don't specify a day, DateTime::createFromFormat() was using the current day (29th) from the system date. This caused February 2026 to overflow to March 1st since February 2026 only has 28 days.

Adding the '!' prefix resets all date components to Unix Epoch before parsing, ensuring the day defaults to 1 instead of using the current day.

### DIFF
--- a/library/Helpers/CanValidateDateTime.php
+++ b/library/Helpers/CanValidateDateTime.php
@@ -11,7 +11,6 @@ namespace Respect\Validation\Helpers;
 
 use DateTime;
 use DateTimeZone;
-
 use function checkdate;
 use function date_default_timezone_get;
 use function date_parse_from_format;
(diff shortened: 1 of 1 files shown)


henriquemoody:
Thank you for the changes, @AmineMx! 🐼 

Could you fix the broken check, and add some tests to cover your changes?

henriquemoody:
Hi @AmineMx!

This was already fixed on version 2.4, and I moved the fix to the `main` branch: https://github.com/Respect/Validation/commit/a74258f1e74f372bb3ec691b13cbdc65922528eb

However, thank you for contributing! 🐼

PR #1585, opened 2026-01-02
Make `Validator` immutable

henriquemoody:
Mutable objects can be challenging to work with in larger codebases because different parts of a system may modify the same instance, making it difficult to trace where and when changes occurred. This becomes especially problematic when debugging unexpected behaviour.

By making `Validator` immutable, we ensure that adding rules via `with()` returns a new instance rather than mutating the original. This provides several benefits:

1. Predictability: A `Validator` instance will always behave the same way throughout its lifetime, regardless of what other parts of the codebase do.

2. Safe dependency injection: Users can now confidently inject a base `Validator` from a DI container, knowing that any modifications made elsewhere will not affect their instance.

3. Easier debugging: Since validators cannot be mutated after creation, there's no need to track down where an unexpected rule was added.

4. Reusability: Users can create an initial `Validator` with some base rules and reuse it by just adding new rules to the chain without affecting the base `Validator`.


henriquemoody:
This might come in handy for users, perhaps something we could release under `/Respect` https://gist.github.com/henriquemoody/acc93915f6b44e79895f19eef3a26787

alganet (approved):

PR #1591, opened 2026-01-05
Port Bash scripts to PHP

henriquemoody:
It makes more sense to use PHP to generate PHP code than to use Bash. I love writing Bash scripts, but I know it’s not for everyone, and they can become quite complex. Porting them to PHP code also lowers the barrier for people to change them.

While I was making those changes, I also noticed a problem with how we save the domain suffixes. We’re converting all of them to ASCII, so we’re not preserving languages such as Chinese, Thai, and Hebrew, which use non-ASCII characters.


henriquemoody:
This code is awful, but I'm not willing to prioritize refactoring it, since it's not some critical part of the library.

data/domain/postal-code.php
```diff
@@ -0,0 +1,717 @@
+<?php declare(strict_types=1);
+// Copyright (c) https://download.geonames.org/export/dump/countryInfo.txt
+// SPDX-License-Identifier: CC-BY-4.0
+return [
+    'AD' => [
+        '/^AD\\d\\d\\d$/',
```
henriquemoody: Those regexes are incorrect, we shouldn't space those characters with `\`

alganet:
Looks great!

> While I was making those changes, I also noticed a problem with how we save the domain suffixes. We’re converting all of them to ASCII, so we’re not preserving languages such as Chinese, Thai, and Hebrew, which use non-ASCII characters.

Can we add a test that ensures validation for those new languages work? Just a few for the most stable international domains.

PR #1594, opened 2026-01-08
Create a "Get started" page

henriquemoody:
That allows users to quickly start playing with Validation.

This page should contain only the very basic information for users to get started with Validation, the "Feature Guide" page should be the one that has more details.


alganet:
I think "Getting Started" is more appropriate for documentation. "Get Started" is what I would use in a call-to-action button.

docs/getting-started.md
```diff
@@ -0,0 +1,98 @@
+---
+weight: -100
```
alganet, line 2: What does `weight` do?

docs/getting-started.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Installation
+
+To install Respect\Validation, use [Composer](http://getcomposer.org):
+
+```shell
+composer require respect/validation:^3.0
+```
+
+Ensure you have PHP 8.5 or above installed.
```
alganet, line 19: Composer will check PHP version, we don't need to put it here.

docs/getting-started.md
```diff
@@ -0,0 +1,98 @@
...
+v::between(1, 256)->assert($input, '{{subject}} is not what I was expecting');
+```
+
+### Custom exceptions
+
+Throw your own exceptions when the validation fails:
+
+```php
+v::between(1, 256)->assert($input, new DomainException('Not within the expected range'));
+```
```
alganet, line 86: This example should probably include catching the custom exception too.

docs/getting-started.md
henriquemoody: It changes the order in which the page is shown when we render it in `mkdocs`.

I don't like this frontmatter very much, but the alternative is that every time we add/change markdown from our documentation we need to manually add them to `mkdocs.yaml`

docs/getting-started.md
henriquemoody: I think it's a good information for users to know which version is minimum one without having to run composer in their machine. It doesn't hurt, right?

alganet: Fair enough!

docs/getting-started.md
alganet: It looks really strange when viewing the markdown from GitHub:

image

I did a search right now, and it seems that this comes from the [mkdocs-nav-weight plugin](https://github.com/shu307/mkdocs-nav-weight/), however, I could not find it in the repo or the PR. Will this order change work without it?

henriquemoody: We have it here, actually: https://github.com/Respect/Validation/commit/35ea95c6f03b7fed1b903dfbcc0922cb4ea962ee#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2R4

alganet: Thanks! I haven't seen it before.

I see there are pages already with the weight metadata. I'll be honest, I don't like that it renders :D But I won't block the PR. I'll investigate a better way of not having those though, even if I have to write my own mkdocs plugin using comments instead of visible metadata! But let's leave that for later.

Have you seen the comment about [the title](https://github.com/Respect/Validation/pull/1594#issuecomment-3726650862)? Sorry it wasn't a diff comment, it ended up as an isolated PR comment.

henriquemoody: Sorry, I didn't see the title comment. I just changed it!

PR #1595, opened 2026-01-08
Update "Feature guide" documentation

henriquemoody:
The documentation was quite outdated after the many changes I made in the main branch. This commit will make sure it's up-to-date.

I'm also creating a "Placeholder pipes" documentation, so there's not too much content in the "Feature guide"

### DIFF
--- a/docs/feature-guide.md
+++ b/docs/feature-guide.md
@@ -8,47 +8,133 @@ The `ValidatorBuilder` class is the core of Respect\Validation, offering a fluen
 
 For convenience, the `ValidatorBuilder` class is aliased as `v`. This means you can write `v::intType()` instead of `\Respect\Validation\ValidatorBuilder::intType()`.
 
-## Validating using booleans
+## Validation methods
+
+### Validating using booleans
 
 With the `isValid()` method, determine if your input meets a specific validator.
 
 ```php
-if (v::intType()->positive()->isValid($input)) {
-    echo 'The input you gave me is a positive integer';
-} else {
+if (!v::intType()->positive()->isValid($input)) {
     echo 'The input you gave me is not a positive integer';
 }
 ```
 
 Note that you can combine multiple validators for a complex validation.
 
-## Validating using exceptions
+### Validating using exceptions
 
 The `assert()` method throws an exception when validation fails. You can handle those exceptions with `try/catch` for more robust error handling.
-
-### Basic example
 
 ```php
 v::intType()->positive()->assert($input);
 ```
 
+### Validating using results
+
+You can validate data and handle the result manually without using exceptions:
+
+```php
+$result = v::numericVal()->positive()->between(1, 255)->validate($input);
+if (!$result->isValid()) {
+    echo $result;
+}
+```
+
+The `validate()` method returns a `ResultQuery` object that has the following methods to output messages:
+
+ - `getMessage()`: Returns the first message from the deepest failed result in chain.
+ - `getFullMessage()`: Returns the full message including all failed results.
+ - `getMessages()`: Returns an array of all messages from failed results.
+
 ## Smart validation
 
-Respect\Validation offers over 150 validators, many of which are designed to address common scenarios. Here's a quick guide to some specific use cases and the validators that make validation straightforward.
+Respect\Validation offers over 150 validators, many of which are designed to address common scenarios.
 
-- Using validators as **PHP Attributes**: [Attributes](validators/Attributes.md).
-- Validating **Arrays**: [Key](validators/Key.md), [KeyOptional](validators/KeyOptional.md), [KeyExists](validators/KeyExists.md).
-- Validating **Array structures**: [KeySet](validators/KeySet.md).
-- Validating **Object properties**: [Property](validators/Property.md), [PropertyOptional](validators/PropertyOptional.md), [PropertyExists](validators/PropertyExists.md).
-- Using **Conditional validation**: [NullOr](validators/NullOr.md), [UndefOr](validators/UndefOr.md), [When](validators/When.md).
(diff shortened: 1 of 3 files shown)


docs/feature-guide.md
```diff
@@ -8,35 +8,129 @@ The `ValidatorBuilder` class is the core of Respect\Validation, offering a fluen
...
+}
+```
+
+The `validate()` method returns a `ResultQuery` object that has the following methods to output messages:
+
+| Method             | Description                                                        |
+| ------------------ | ------------------------------------------------------------------ |
+| `getMessage()`     | Returns the first message from the deepest failed result in chain. |
+| `getFullMessage()` | Returns the full message including all failed results.             |
+| `getMessages()`    | Returns an array of all messages from failed results.              |
```
alganet: Nitpick: do we really need a table? Maybe a bullet list is enough. It also reads better on plain text.

```suggestion
 - `getMessage()`: Returns the first message from the deepest failed result in chain.
 - `getFullMessage()`: Returns the full message including all failed results.
 - `getMessages()` : Returns an array of all messages from failed results.
```

docs/feature-guide.md
```diff
@@ -8,35 +8,129 @@ The `ValidatorBuilder` class is the core of Respect\Validation, offering a fluen
...
+
+The `ResultQuery` also has methods to query nested results:
+
+```php
+$mysqlUserResult = $result->findByPath('mysql.user');
+if ($mysqlUserResult !== null) {
+    echo $mysqlUserResult;
+}
+```
```
alganet: This is missing some context. The last time we used the `$result` variable in this page was an int validation. Perhaps we need to introduce the database connection settings example more explicitly, or something smaller that can still be relevant for the path example.

docs/feature-guide.md
```diff
@@ -8,35 +8,129 @@ The `ValidatorBuilder` class is the core of Respect\Validation, offering a fluen
...
+v::size('MB', v::not(v::greaterThan(5)))->assert('path/to/file.zip');
+// → The size in megabytes of "path/to/file.zip" must not be greater than 5
+
+v::dateTimeDiff('years', v::greaterThan(18))->assert('2025');
+// → The number of years between now and "2025" must be greater than 18
+```
+
+### Prefixed shortcuts
+
+For convenience, several prefixed shortcuts are available for validators that wrap other validators:
```
alganet, line 105: Perhaps we can link to the mixin class so the user can see all available shortcuts.

docs/feature-guide.md
```diff
@@ -48,15 +142,19 @@ Respect\Validation offers over 150 validators, many of which are designed to add
...
+### Using custom templates
 
 Define your own error message when the validation fails:
 
 ```php
 v::between(1, 256)->assert($input, '{{subject}} is not what I was expecting');
 ```
 
-### Custom templates per validator
+### Custom templates per rule
```
alganet: Are going back? :D

docs/feature-guide.md
```diff
@@ -111,11 +254,65 @@ $validator->assert('validation');
 $validator->assert('alexandre gaigalas');
 ```
 
-## Customising validator names
+Every time you add a new rule to the chain, a new validator is created. That means you can do things like:
```
alganet:
```suggestion
Every time you add a new node to the chain, a new immutable instance is created. That means you can do things like:
```

docs/feature-guide.md
```diff
@@ -111,11 +254,65 @@ $validator->assert('validation');
...
-Template messages include the placeholder `{{subject}}`, which defaults to the input. Use `setName()` to replace it with a more descriptive label.
+```php
+$baseValidator = v::intType()->between(1, 155);
+
+$baseValidator->even()->assert($input1);
+
+$baseValidator->odd()->assert($input2);
+```
+
+Both the `even()` and `odd()` calls created a new validator, which means that the `$baseValidator` remained unchanged.
```
alganet:
```suggestion
Both the `even()` and `odd()` calls created a new instance, which means that the `$baseValidator` remained unchanged.
```

docs/feature-guide.md
```diff
@@ -111,11 +254,65 @@ $validator->assert('validation');
...
+
+Use the `not` prefix to invert a validator:
+
+```php
+v::notEquals('main')->assert($input);
+```
+
+### Nested validation paths
+
+Validation can trace the path of nested structures and display helpful messages:
```
alganet: There was a short introduction to paths before in this document, I left a comment about lacking context. Maybe we can remove it and replace for an anchor to this more in-depth part.

docs/feature-guide.md
henriquemoody: That would be difficult, because there are multiple mixing classes. I think we could create a page with the prefixes, though. What do you think?

alganet: That would be cool. They're generated aren't they? Perhaps we can generate the docs too.

henriquemoody: I just added a markdown with it. I haven't fully reviewed it though.

docs/feature-guide.md
```diff
@@ -85,23 +217,22 @@ use Respect\Validation\Exceptions\ValidationException;
...
+v::named('Your email', v::email())->assert($input);
+// → Your email must be a valid email
 ```
 
-For more details, check the [Not](validators/Not.md) validator documentation.
-
 ## Reusing validators
 
-Validators can be created once and reused across multiple inputs.
+The `ValidatorBuilder` is immutable—every time you add a new validator to the chain, you get a new instance while the original remains unchanged. This makes validators safe to create once and reuse:
```
alganet: We almost never use em dashes, it feels weird to have one here.

```suggestion
The `ValidatorBuilder` is immutable. Every time you add a new validator to the chain, you get a new instance while the original remains unchanged. This makes validators safe to create once and reuse:
```

alganet (approved):
Small nitpick on the emdash. You can ignore it and merge if you want.

Issue #1589, opened 2026-01-04
Rename logical validators


henriquemoody:
Related to the discussion in https://github.com/Respect/Validation/pull/1578#discussion_r2658221972

alganet:
Regarding the new names (continuing the discussion here).

If we could use just `v::and, v::or, v::xor, etc` it would be better, but there is a reserved word issue here: 

https://www.php.net/manual/en/reserved.keywords.php

The previous names `allof` and `oneof` that I have chosen previously came from RDF terminology:

https://www.w3.org/2004/12/q/doc/rdf-rulesets.html

They also are present in JSON Schema. I'm actually fond of them, but the conflict with the `All` rule is an issue for that naming convention.

Since we drifted a lot from RDF since 0.9/1.0, I'm willing to part with AllOf/OneOf, but I think we should still latch to an existing estabilished convention (even if it's not an official standard). The "Logic" prefix comes from that, since lots of codebases use it (especially parsers who need to circumvent `and` and `or` being reserved words):

https://github.com/search?q=logicand+logicor&type=code

The names `groupedAnd` and `groupedOr` make sense, and there is a precedence:

https://github.com/search?q=groupedand+groupedor&type=code

However, it seems way less common than the `Logic` prefix.

Another option is using the `op` prefix, for "operator": `opAnd`, `opOr` and so on. These are also very common:

https://github.com/search?q=opand+opor&type=code
https://github.com/search?q=op_and+op_or&type=code

henriquemoody:
We could use `v::and`, `v::or`, `v::xor`, as method names, but not as rule names. We could solve that by introducing an `Alias` transformer that we already had. The issue is the inconsistency.

My concern with the `logic*` is that it doesn't express that there's a group of rules to be accounted for. Form the links your sent, the logic* ones didn't seem linked to groups.

I'm not against the `logic*`, though.

alganet:
The examples searches I linked to, as I mentioned, are mostly from parsers that have to circumvent reserved words. Those parsers would like to use raw `and` and `or`, but those are reserved, so they used `LogicAnd` and `LogicOr`. The convention in that case is *how to circumvent reserved words*, not how to group things.

If I do `v::a()->logicAnd()->b()`, someone might be inclined to believe that I'm saying `A && B`, but I'm actually saying `A && () && B`, so yes, there is a world in which this can cause confusion.

[...]

PR #1599, opened 2026-01-11
Fix message overriding bug in NestedArrayFormatter

alganet:
This commit resolves an issue where validation messages would overwrite each other when multiple validators failed on the same path or key (e.g., within an `Each` or `Key` validator).

Changes to `NestedArrayFormatter`:
- Implemented a merge strategy: Key collisions now result in a list of messages instead of the last message winning.
- Improved handling of mixed key types: When both numeric and string keys are present (common in composite validators), numeric keys are now replaced by the validator's ID (e.g., `arrayType`, `equals`) to provide meaningful, distinct keys.
- Preserved list behavior: Purely numeric key sets are treated as lists, maintaining their sequence without re-keying logic.
- Refactored the class to use smaller, single-purpose methods and `array_reduce` for clarity.

Tests:
- Updated feature tests (`EachTest`, `AttributesTest`, etc.) to expect the full set of validation errors.
- Enhanced `NestedArrayFormatterTest` with scenarios for key collisions, mixed keys, and ID substitution.

### DIFF
--- a/library/Message/Formatter/NestedArrayFormatter.php
+++ b/library/Message/Formatter/NestedArrayFormatter.php
@@ -13,8 +13,11 @@ use Respect\Validation\Message\ArrayFormatter;
 use Respect\Validation\Message\Renderer;
 use Respect\Validation\Result;
 
+use function array_reduce;
 use function count;
 use function current;
+use function is_array;
+use function is_numeric;
 
 final readonly class NestedArrayFormatter implements ArrayFormatter
 {
@@ -25,31 +28,117 @@ final readonly class NestedArrayFormatter implements ArrayFormatter
      */
     public function format(Result $result, Renderer $renderer, array $templates): array
     {
-        if (count($result->children) === 0) {
-            return [
-                $result->path->value ?? $result->id->value => $renderer->render($result, $templates),
-            ];
+        if ($result->children === []) {
+            return [$this->getKey($result) => $renderer->render($result, $templates)];
         }
 
-        $messages = [];
-        foreach ($result->children as $child) {
-            $key = $child->path->value ?? $child->id->value;
-            $messages[$key] = $this->format(
-                $child->withoutName(),
+        [$children, $hasString] = $this->prepareChildren($result->children);
+
+        $messages = array_reduce(
+            $children,
+            fn(array $messages, array $item) => $this->appendMessage(
+                $messages,
+                $item['key'],
+                $item['child'],
                 $renderer,
                 $templates,
-            );
-            if (count($messages[$key]) !== 1) {
-                continue;
-            }
-
-            $messages[$key] = current($messages[$key]);
-        }
+                $hasString,
+            ),
+            [],
+        );
 
         if (count($messages) > 1) {
             return ['__root__' => $renderer->render($result, $templates)] + $messages;
         }
 
         return $messages;
+    }
+
+    /**
+     * @param array<Result> $children
+     *
+     * @return array{0: array<array{key: string|int, child: Result}>, 1: bool}
+     */
+    private function prepareChildren(array $children): array
+    {
+        $mapped = [];
+        $hasString = false;
+
+        foreach ($children as $child) {
+            $key = $this->getKey($child);
+            if (!is_numeric($key)) {
+                $hasString = true;
+            }
+
+            $mapped[] = ['key' => $key, 'child' => $child];
+        }
+
+        return [$mapped, $hasString];
+    }
+
+    /**
+     * @param array<string|int, mixed> $messages
+     * @param array<string|int, mixed> $templates
+     *
+     * @return array<string|int, mixed>
+     */
+    private function appendMessage(
+        array $messages,
+        string|int $key,
+        Result $child,
+        Renderer $renderer,
+        array $templates,
+        bool $hasString,
+    ): array {
+        if ($hasString && is_numeric($key)) {
+            $key = $child->id->value;
+        }
+
+        $message = $this->renderChild($child, $renderer, $templates);
+
+        if (!$hasString) {
+            $messages[] = $message;
+
+            return $messages;
+        }
+
+        if (isset($messages[$key])) {
+            if (!is_array($messages[$key])) {
+                $messages[$key] = [$messages[$key]];
+            }
+
+            $messages[$key][] = $message;
+
+            return $messages;
+        }
+
+        $messages[$key] = $message;
+
+        return $messages;
+    }
+
+    /**
+     * @param array<string|int, array<string>> $templates
+     *
+     * @return array<string>|string
+     */
+    private function renderChild(Result $child, Renderer $renderer, array $templates): array|string
+    {
+        $formatted = $this->format(
+            $child->withoutName(),
+            $renderer,
+            $templates,
+        );
+
+        if (count($formatted) === 1) {
+            return current($formatted);
+        }
+
+        return $formatted;
+    }
+
+    private function getKey(Result $result): string|int
+    {
+        return $result->path->value ?? $result->id->value;
     }
 }
--- a/tests/feature/Issues/Issue1376Test.php
+++ b/tests/feature/Issues/Issue1376Test.php
@@ -29,7 +29,10 @@ test('https://github.com/Respect/Validation/issues/1376', catchAll(
             '__root__' => '`stdClass { +$author="foo" }` must pass all the rules',
             'title' => '`.title` must be present',
             'description' => '`.description` must be present',
-            'author' => 'The length of `.author` must be between 1 and 2',
+            'author' => [
+                '`.author` must be an integer',
+                'The length of `.author` must be between 1 and 2',
+            ],
             'user' => '`.user` must be present',
         ]),
 ));
--- a/tests/feature/Validators/AttributesTest.php
+++ b/tests/feature/Validators/AttributesTest.php
@@ -57,7 +57,10 @@ test('Multiple attributes, all failed', catchAll(
         ->and($messages)->toBe([
             '__root__' => '`Respect\Validation\Test\Stubs\WithAttributes { +$name="" +$birthdate="not a date" #$phone="not a phone number" + ... }` must pass the rules',
             'name' => '`.name` must be defined',
-            'birthdate' => 'For comparison with now, `.birthdate` must be a valid datetime',
+            'birthdate' => [
+                '`.birthdate` must be a valid date in the format "2005-12-30"',
+                'For comparison with now, `.birthdate` must be a valid datetime',
+            ],
             'phone' => '`.phone` must be a valid telephone number or must be null',
             'email' => '`.email` must be a valid email address or must be null',
         ]),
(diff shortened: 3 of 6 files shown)


tests/feature/Validators/AttributesTest.php
```diff
@@ -57,7 +57,10 @@
         ->and($messages)->toBe([
             '__root__' => '`Respect\Validation\Test\Stubs\WithAttributes { +$name="" +$birthdate="not a date" #$phone="not a phone number" + ... }` must pass the rules',
             'name' => '`.name` must be defined',
-            'birthdate' => 'For comparison with now, `.birthdate` must be a valid datetime',
+            'birthdate' => [
+                '`.birthdate` must be a valid date in the format "2005-12-30"',
+                'For comparison with now, `.birthdate` must be a valid datetime',
+            ],
```
alganet, line 63: I don't like that this now is more verbose, however, I'm not sure how to fix.

This is a consequence of us validating the dates when using the diff validator. In the real world, maybe people wont use Date+DateTimeDiff together, since one kind of implies the other by consequence.

tests/feature/Issues/Issue1376Test.php
```diff
@@ -29,7 +29,10 @@
             '__root__' => '`stdClass { +$author="foo" }` must pass all the rules',
             'title' => '`.title` must be present',
             'description' => '`.description` must be present',
-            'author' => 'The length of `.author` must be between 1 and 2',
+            'author' => [
+                '`.author` must be an integer',
+                'The length of `.author` must be between 1 and 2',
+            ],
```
alganet, line 35: This is an unfortunate consequence of implementing a conflict-solving non-overriding mechanism. Some validators that previously produced a more flat list now might have mixed scalar/complex values.

This is a known issue all across Validation (mixed scalar/complex responses), and it is still an open problem. This change just exposes how this problem was not affecting just `Each`, it was being hidden by the overriding behavior.

library/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -25,31 +28,120 @@
...
+            }
+
+            $messages[$key][] = $message;
+
+            return $messages;
         }
 
+        $messages[$key] = $message;
+
         return $messages;
```
henriquemoody:
```suggestion
        return [$key => $message] + $messages;
```
Nitpick

library/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -25,31 +28,120 @@
...
+        }
+
+        if (isset($messages[$key])) {
+            if (!is_array($messages[$key])) {
+                $messages[$key] = [$messages[$key]];
+            }
+
+            $messages[$key][] = $message;
+
+            return $messages;
```
henriquemoody:
```suggestion
            return [$key => [$message]] + $messages;
```
Nitpick

tests/feature/Issues/Issue1376Test.php
henriquemoody: I'm not even sure this a problem as much as an expected behavior. What we should be more clear about is that `ValidatorBuilder` will generate failing results for all validators that were passed to it.

I'm starting to think we should maybe have a `CircuitValidatorBuilder` that would build a inner `Circuit` validator instead of a `Reducer` (that uses `AllOf` if there's more than 1 validator).

We could have something like this, even:

```php
// From the chain
v::intVal()->lengthBetween(1, 2)->asCircuit();

// As starting point
v::asCircuit()->intVal()->lengthBetween(1, 2);

// from a different alias
cv::intVal()->lengthBetween(1, 2);
```

So, in a case like that, if the user doesn't want all messages, they can make a choice.

tests/feature/Validators/EachTest.php
```diff
@@ -301,15 +309,17 @@
...
         FULL_MESSAGE)
         ->and($messages)->toBe([
             '__root__' => 'Each item in `[["not_int": "wrong"], ["my_int": 2], "not an array"]` must be valid',
             0 => '`.0.my_int` must be present',
             1 => '`.1.my_int` must be an odd number',
             2 => [
                 '__root__' => '`.2` must pass all the rules',
-                2 => '`.2` must be an array',
+                'arrayType' => '`.2` must be an array',
                 'my_int' => '`.2.my_int` must be present',
```
henriquemoody, line 321: Unrelated (more of a mental note)

This should probably be:

```
`.my_int` must be present
```

Now that we have the `Subject` it should be easy to follow the same logic as in `NestedListStringFormatter`

library/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -25,31 +28,120 @@
...
-            $key = $child->path->value ?? $child->id->value;
-            $messages[$key] = $this->format(
-                $child->withoutName(),
+        [$children, $hasString, $hasNumeric] = $this->prepareChildren($result->children);
+
+        $messages = array_reduce(
+            $children,
+            fn(array $messages, array $item) => $this->appendMessage(
+                $messages,
+                $item,
```
henriquemoody:
```suggestion
                $item['key'],
                $item['child'],
```

library/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -25,31 +28,120 @@
...
+    private function appendMessage(
+        array $messages,
+        array $item,
+        Renderer $renderer,
+        array $templates,
+        bool $hasString,
+        bool $hasNumeric,
+    ): array {
+        $key = $item['key'];
+        $child = $item['child'];
```
henriquemoody:
```suggestion
    /**
     * @param array<string|int, mixed> $messages
     * @param array<string|int, mixed> $templates
     *
     * @return array<string|int, mixed>
     */
    private function appendMessage(
        array $messages,
        string|int $key,
        Result $child,
        Renderer $renderer,
        array $templates,
        bool $hasString,
        bool $hasNumeric,
    ): array {
```

library/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -25,31 +28,120 @@
...
+     *
+     * @return array<string|int, mixed>
+     */
+    private function appendMessage(
+        array $messages,
+        array $item,
+        Renderer $renderer,
+        array $templates,
+        bool $hasString,
+        bool $hasNumeric,
```
henriquemoody: The `$hasNumeric` is redundant, you can achieve the same with only the `$hasString`

library/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -25,31 +28,120 @@
...
+            return $messages;
         }
 
+        $messages[$key] = $message;
+
         return $messages;
     }
+
+    /** @param array<string|int, mixed> $templates */
+    private function renderChild(Result $child, Renderer $renderer, array $templates): mixed
```
henriquemoody:
```suggestion
    private function renderChild(Result $child, Renderer $renderer, array $templates): array|string
```

Dealing with the recursive array using docblocks will be annoying, but we could use a `@psalm-type` at the header of the class with that:

```
 * @psalm-type Message = string|array<Message>
```

Same for the templates

tests/feature/Issues/Issue1376Test.php
alganet: The `asCircuit` idea is related to the previous check/assert behavior we discussed in #1589. Basically, "all leaves" or "stops at first leave of each branch".

I still believe the best architecture solves this as an evaluation choice, not a building choice. If you're interested in only true/false, then is `asCircuit` should be the default mode, for example. It's not only a message choice, it's a performance choice too (in the sense that you only want all leaves if you need their messages).

tests/feature/Validators/EachTest.php
alganet: We can still have many entries inside `Each` that could have `my_int` in it. Perhaps `.2.my_int` is better.

henriquemoody: But if you're already in a nested structure, wouldn't it be better to just see which key failed inside that structure?

I think it's more of a cosmetic choice, really, maybe you prefer with `.2.my_int` and me `my_int`. If that's the case, then there's no better or worse, so not changing what's already there is the best choice, IMO 😅

library/Message/Formatter/NestedArrayFormatter.php
alganet: Done!

library/Message/Formatter/NestedArrayFormatter.php
alganet: Done!

library/Message/Formatter/NestedArrayFormatter.php
alganet: Removed it! Nice catch.

library/Message/Formatter/NestedArrayFormatter.php
alganet: Didn't needed the psalm type. Have I missed anything?

henriquemoody: I think PHPStan will understand how it works with no problem because of how you use it, but because the actual type is a recursive array, I _think_ the correct type is what I described.

alganet: I'm not very familiar with psalm-type. As I understand, they're similar to type aliases in Python.

I tried, but got this:

image

(and a bunch of other errors too due to the unresolved type)

alganet:
@henriquemoody I want to be type-pedantic (types are cool). I'll try for a little while to remove the `mixed` annotations. If by tomorrow I can't fix it, I'll merge :D

library/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -25,31 +28,117 @@
...
+    private function prepareChildren(array $children): array
+    {
+        $mapped = [];
+        $hasString = false;
+
+        foreach ($children as $child) {
+            $key = $this->getKey($child);
+            if (!is_numeric($key)) {
+                $hasString = true;
             }
```
henriquemoody, line 71: You don't need that anymore

henriquemoody: I was seeing old code in GitHub mobile

library/Message/Formatter/NestedArrayFormatter.php
henriquemoody: Well, if doesn't work, it doesn't work 😅 -- sorry for wasting your time!

tests/feature/Validators/EachTest.php
alganet: That's a good point. I'm going to think about that for a while, and the use cases!

tests/feature/Issues/Issue1376Test.php
henriquemoody: Although this is related to #1589, I think this is a discussion on itself. One of the advantages of v3 is the ability to get insights on everything that happens inside each validator. Doing a short-circuit on all validations feels like a step back, and it will again restrict validators that wrap other validators. Making it a choice, though, allows the to user decide what will happen.

I understand the performance implications you're mentioning, but that's the trade off of the v3. If we want results that are more concise, perhaps that's more up to a `ResultFilter` implementation.

PR #1602, opened 2026-01-11
Introduce phpbench benchmarks and profiles

alganet:
This commit is the first step in setting up *Continuous Performance* for the repository.

 - Adds phpbench/phpbench to dev dependencies.
 - Adds an initial `ValidatorBench.php` file with validate benchmarks for several validators.
 - Adds `composer bench` script to run benchmarks.
 - Adds `composer bench:profile` script to generate profiles.

---

**phpbench** can do much more than how it is currently configured, but these are future steps. For now, it can help developers to catch performance and memory usage regressions.

The `composer bench:profile`, in conjunction with tools like `kcachegrind` or `webgrind`, is particularly useful to find hotspots and do more detailed analysis.

Example benchmark profile loaded with kcachegrind:

Screenshot 2026-01-11 202749

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,11 @@
 .couscous/
 .iso-codes-cache/
+.phpbench/
 .phpcs.cache
 .phpunit.cache/
 composer.lock
 Makefile
+phpbench.json
 phpcs.xml
 phpstan.neon
 phpunit.xml
--- a/phpbench.json.dist
+++ b/phpbench.json.dist
@@ -1,0 +1,6 @@
+{
+    "$schema": "./vendor/phpbench/phpbench/phpbench.schema.json",
+    "runner.bootstrap": "vendor/autoload.php",
+    "runner.path": "tests/benchmark",
+    "runner.file_pattern": "*Bench.php"
+}
--- a/tests/benchmark/ValidatorBench.php
+++ b/tests/benchmark/ValidatorBench.php
@@ -1,0 +1,113 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Respect\Validation\Benchmarks;
+
+/*
+ * Copyright (c) Alexandre Gomes Gaigalas <alganet@gmail.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+use Generator;
+use PhpBench\Attributes as Bench;
+use Respect\Validation\Validator;
+use Respect\Validation\ValidatorBuilder as v;
+
+class ValidatorBench
+{
+    /** @param array<Validator, mixed> $params */    #[Bench\Iterations(10)]
+    #[Bench\RetryThreshold(10)]
+    #[Bench\Revs(5)]
+    #[Bench\ParamProviders(['provideValidatorInput'])]
+    public function benchValidate(array $params): void
+    {
+        [$v, $input] = $params;
+        $v->validate($input);
+    }
+
+    public function provideValidatorInput(): Generator
+    {
+        yield 'AllOf' => [v::allOf(v::intVal(), v::greaterThan(0)), 5];
+        yield 'Alnum' => [v::alnum(), 'abc123'];
+        yield 'Alpha' => [v::alpha(), 'abc'];
+        yield 'AnyOf' => [v::anyOf(v::intVal(), v::stringVal()), 5];
+        yield 'ArrayType' => [v::arrayType(), []];
+        yield 'ArrayVal' => [v::arrayVal(), []];
+        yield 'Between' => [v::between(1, 10), 5];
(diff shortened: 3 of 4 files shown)


tests/benchmark/ValidatorBench.php
```diff
@@ -0,0 +1,120 @@
...
+use Generator;
+use PhpBench\Attributes as Bench;
+use Respect\Validation\Validator;
+use Respect\Validation\ValidatorBuilder as v;
+
+use function fopen;
+
+class ValidatorBench
+{
+    /** @param array<Validator, mixed> $params */
```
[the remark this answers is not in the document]
alganet: No, it's not.

.gitignore
```diff
@@ -1,7 +1,9 @@
 .couscous/
 .iso-codes-cache/
+.phpbench/
 .phpcs.cache
 .phpunit.cache/
+.phpunit.result.cache
```
henriquemoody: I think that's there because of switching from branch 2.4 to main, that happens to me too. The current version on master doesn't generate this file, so, I think it's better not to add this noise here. Perhaps for the time being we should both add that to our global git ignore file

phpbench.json.dist
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "./vendor/phpbench/phpbench/phpbench.schema.json",
+    "runner.bootstrap": "vendor/autoload.php",
+    "runner.path": "tests/benchmark",
+    "runner.file_pattern": "*Bench.php"
+}
```
henriquemoody, line 6: It would be best to name it `phpbench.json.dist` and add `phpbench.json` to the git ignore list, that way we can play around with it without having to avoid committing it

tests/benchmark/ValidatorBench.php
```diff
@@ -0,0 +1,113 @@
...
+use PhpBench\Attributes as Bench;
+use Respect\Validation\Validator;
+use Respect\Validation\ValidatorBuilder as v;
+
+class ValidatorBench
+{
+    /** @param array<Validator, mixed> $params */    #[Bench\Iterations(10)]
+    #[Bench\RetryThreshold(10)]
+    #[Bench\Revs(5)]
+    #[Bench\ParamProviders(['provideValidatorInput'])]
```
henriquemoody, line 22: How about we add an assertion that's around 10% higher to whatever number we're getting now and we add that to the Continuous Integration workflow?

henriquemoody (dismissed):
Wrote a few comments but having this here would be awesome as is

tests/benchmark/ValidatorBench.php
alganet: It's not that simple. This 10% needs to be relative to a previous result.

This means we need to create a storage branch, and keep the phpbench storage file there to use it as baseline. It's quite a complicated workflow, and it requires some preparation.

Therefore, I separated this effort in at least 4 parts that are smaller atomic deliverables that are easier to review:

 - This PR.
 - Add remaining validators and other code paths (getting messages, etc). Also, perhaps fixing non-serializable validators (the data provider showed me this bug, that's why some of them are missing, I don't know if I will be able to fix them all). Having serialized validators is not mandatory (we could not use data providers in the bench files), but it's a nice to have. 
 - Continuous Performance workflow (just run and commit history to dedicated branch)
 - Continuous Performance summary (on workflow page), PR comment, tresholds.

The treshold is the last step.

henriquemoody: I like the plan you laid out!

As I said before, I'm happy with this step as it is, I was more looking for a low hanging fruit, but as is this is awesome!

tests/benchmark/ValidatorBench.php
```diff
@@ -0,0 +1,113 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Respect\Validation\Benchmarks;
+
+/*
+ * Copyright (c) Alexandre Gomes Gaigalas <alganet@gmail.com>
+ * SPDX-License-Identifier: MIT
+ */
```
henriquemoody, line 10:
```suggestion
<?php

/*
 * Copyright (c) Alexandre Gomes Gaigalas <alganet@gmail.com>
 * SPDX-License-Identifier: MIT
 */
 
 declare(strict_types=1);

namespace Respect\Validation\Benchmarks;
```

Nit picking... 

We should probably use reuse soon

henriquemoody (approved):

tests/benchmark/ValidatorBench.php
alganet: If it weren't for the planned reuse switch, I would fix this right now! But yep, it's coming soon.

PR #1603, opened 2026-01-12
Setup GitHub actions to check documentation files

henriquemoody:
This pull request configures GitHub actions to check flaws in the documentation. Many documents we have are related to actual code, and sometimes they get out of sync. By running this command in the workflow, we make sure that the changes we made will be reflected in the documentation files.


henriquemoody:
That's a great idea! Perhaps something for 3.1, what do you say?

PR #1605, opened 2026-01-15
Fix factory when validator has no constructor

alganet:
This was affecting both v::version and v::attributes, making them impossible to use except for the concrete interface.

This change allows these validators to be built using the chain.

### DIFF
--- a/library/NamespacedValidatorFactory.php
+++ b/library/NamespacedValidatorFactory.php
@@ -17,6 +17,7 @@ use Respect\Validation\Transformers\Transformer;
 use Respect\Validation\Transformers\ValidatorSpec;
 
 use function array_merge;
+use function Respect\Stringifier\stringify;
 use function sprintf;
 use function trim;
 use function ucfirst;
@@ -57,6 +58,8 @@ final readonly class NamespacedValidatorFactory implements ValidatorFactory
     /** @param array<int, mixed> $arguments */
     private function createRule(string $ruleName, array $arguments = []): Validator
     {
+        $reflection = null;
+
         foreach ($this->rulesNamespaces as $namespace) {
             try {
                 /** @var class-string<Validator> $name */
@@ -72,12 +75,22 @@ final readonly class NamespacedValidatorFactory implements ValidatorFactory
                     throw new InvalidClassException(sprintf('"%s" must be instantiable', $name));
                 }
 
-                return $reflection->newInstanceArgs($arguments);
+                break;
             } catch (ReflectionException) {
                 continue;
             }
         }
 
-        throw new ComponentException(sprintf('"%s" is not a valid rule name', $ruleName));
+        if (!$reflection) {
+            throw new ComponentException(sprintf('"%s" is not a valid rule name', $ruleName));
+        }
+
+        try {
+            return $reflection->newInstanceArgs($arguments);
+        } catch (ReflectionException) {
+            throw new InvalidClassException(
+                sprintf('"%s" could not be instantiated with arguments %s', $ruleName, stringify($arguments)),
+            );
+        }
     }
 }
--- a/tests/unit/NamespacedRuleFactoryTest.php
+++ b/tests/unit/NamespacedRuleFactoryTest.php
@@ -60,6 +60,19 @@ final class NamespacedRuleFactoryTest extends TestCase
     }
 
     #[Test]
+    public function shouldThrowsAnExceptionOnConstructorReflectionFailure(): void
+    {
+        $constructorArguments = ['a', 'b'];
+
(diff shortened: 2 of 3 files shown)


tests/unit/NamespacedRuleFactoryTest.php
```diff
@@ -59,6 +60,17 @@ public function shouldDefineConstructorArgumentsWhenCreatingRule(): void
         self::assertSame($constructorArguments, $validator->validations);
     }
 
+    #[Test]
+    public function shouldCreateRuleWhenNoConstructorIgnoringArguments(): void
+    {
+        $constructorArguments = ['a', 'b'];
```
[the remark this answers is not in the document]
alganet: That won't be necessary.

library/NamespacedValidatorFactory.php
```diff
@@ -72,6 +72,11 @@ private function createRule(string $ruleName, array $arguments = []): Validator
                     throw new InvalidClassException(sprintf('"%s" must be instantiable', $name));
                 }
 
+                $constructor = $reflection->getConstructor();
+                if ($constructor === null) {
+                    return $reflection->newInstance();
+                }
```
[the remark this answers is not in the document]
alganet: That is out of scope for this, I just want to recover the ability for the chain to build v::attributes and v::version cleanly for now.

henriquemoody:
This is really curious. How come the tests didn't pick this up?

tests/unit/NamespacedRuleFactoryTest.php
```diff
@@ -59,6 +60,17 @@ public function shouldDefineConstructorArgumentsWhenCreatingRule(): void
...
+
+        $factory = new NamespacedValidatorFactory(new StubTransformer(), [self::TEST_RULES_NAMESPACE]);
+        $validator = $factory->create('noConstructor', $constructorArguments);
+
+        self::assertInstanceOf(NoConstructor::class, $validator);
```
henriquemoody: Should this really be acceptable? In this case, it seems like the arguments didn't match the contract established by the constructor. It's kinda how PHP works, I know, but if that's the fix I wonder if we should have it

alganet: PHP allows passing constructor arguments to a class that does not have constructors.

henriquemoody: That's what I said, I know. But how come is this fix needed? There are several tests building validations without constructor, and that wasn't ever a problem. We built with reflection many versions ago, I'm curious to how you found this issue

alganet:
@henriquemoody  Now I see the problem better.

I was doing `v::version("argument")`, and got the `"version" is not a valid rule name` error. I rushed a fix and it mislead me. The fix works, but perhaps it's not the best choice of what to do (accidentally works).

Perhaps we should throw "`"version" has 0 arguments, 1 passed instead.`. That would be less misleading than saying it doesn't exist, and still informative to the user.

EDIT: It's kind of a pain to manage mandatory/optional arguments though, and we shouldn't be checking that kind of stuff during runtime (ReflectionArgument and so on are slow).

tests/unit/NamespacedRuleFactoryTest.php
alganet: See my comment on the PR root. I think you're right.

henriquemoody:
@alganet I think maybe we just need better exception handling. Right now there are too many things in the same try block. We could improve that without managing arguments.

An exception saying that a validator is not valid is one thing, but in this case we should say we were not able to create a validator with the given arguments.

Since you are at it already, perhaps you could put the creation of the validator in another try block and handle that exception with a more descriptive message

alganet:
@henriquemoody makes sense, good idea!

library/NamespacedValidatorFactory.php
```diff
@@ -72,12 +74,22 @@ private function createRule(string $ruleName, array $arguments = []): Validator
...
-        throw new ComponentException(sprintf('"%s" is not a valid rule name', $ruleName));
+        if (!$reflection) {
+            throw new ComponentException(sprintf('"%s" is not a valid rule name', $ruleName));
+        }
+
+        try {
+            return $reflection->newInstanceArgs($arguments);
+        } catch (ReflectionException) {
+            throw new InvalidClassException(
+                sprintf('"%s" could not be instantiated', $ruleName),
```
henriquemoody:
```suggestion
                sprintf('Could not create validator "%s" with the %s arguments', $ruleName, stringify($arguments)),
```

I think it's best to make that more geared towards what the user is doing. I  think "could not be instantiated" describes exactly what happens, but it might not be so useful to the user.

PR #1606, opened 2026-01-15
Introduce ContainsCount validator

alganet:
This validator is similar to Contains, but also checks how many times the needle appears.

Additionally, the Domain validator was changed to use it instead of relying on an unserializable callback, thus, making it serializable.


alganet:
I wanted to make the `$count` parameter accept a wrapped validator, but I could not do it and also have template messages. I also found out some other validators don't have that, such as `Between`, so I decided to stop banging my head and just deliver the simpler version.

library/Validators/ContainsCount.php
```diff
@@ -0,0 +1,93 @@
...
+
+use function in_array;
+use function is_array;
+use function is_scalar;
+use function mb_strtolower;
+use function mb_substr_count;
+
+#[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
+#[Template(
+    '{{subject}} must contain {{containsValue}} {{count}} time(s)',
```
henriquemoody, line 25: You could also create two templates: singular and plural.

library/Mixins/Builder.php
```diff
@@ -84,6 +84,8 @@ public static function contains(mixed $containsValue, bool $identical = false):
     /** @param non-empty-array<mixed> $needles */
     public static function containsAny(array $needles, bool $identical = false): Chain;
 
+    public static function containsCount(mixed $containsValue, int $count, bool $identical = false): Chain;
```
henriquemoody, line 87: Run `bin/console create:mixin`, it will generate for all prefixes as well.


We should probably add that to the workflow as well.

If you don't run it will show in the next diff when someone runs it

library/Validators/ContainsCount.php
```diff
@@ -0,0 +1,93 @@
...
+        $count = 0;
+        foreach ($input as $item) {
+            if (!in_array($this->containsValue, [$item], $this->identical)) {
+                continue;
+            }
+
+            $count++;
+        }
+
+        return $count;
```
henriquemoody: Feels like an array_reduce kinda job, but you don't need to change anything

henriquemoody (approved):

library/Mixins/Builder.php
alganet: Just add it to composer scripts so people (like me!) know it exists.

library/Validators/ContainsCount.php
alganet: Done!

library/Mixins/Builder.php
alganet: Done in the last force-push!

PR #1607, opened 2026-01-16
Add more benchmarks, smoke tests, make FilterVar serializable

alganet:
This commit introduces a new feature test: SerializableTest, that
tests several validators for their ability to be serialized and
unserialized.

It also makes it so that the same list of validators can be used
by both simple benchmarks and "smoke tests" of all kinds, including
this serialize/unserialize one.

Additionally, the FilterVar validator was modified. Previously, due
to the use of Callback, it was not serializable, but now it is.

### DIFF
--- a/library/Validators/FilterVar.php
+++ b/library/Validators/FilterVar.php
@@ -12,12 +12,11 @@ namespace Respect\Validation\Validators;
 use Attribute;
 use Respect\Validation\Exceptions\InvalidRuleConstructorException;
 use Respect\Validation\Message\Template;
-use Respect\Validation\Validators\Core\Envelope;
+use Respect\Validation\Result;
+use Respect\Validation\Validator;
 
 use function array_key_exists;
 use function filter_var;
-use function is_array;
-use function is_int;
 
 use const FILTER_VALIDATE_BOOLEAN;
 use const FILTER_VALIDATE_DOMAIN;
@@ -33,7 +32,7 @@ use const FILTER_VALIDATE_URL;
     '{{subject}} must be valid',
     '{{subject}} must not be valid',
 )]
-final class FilterVar extends Envelope
+final readonly class FilterVar implements Validator
 {
     private const array ALLOWED_FILTERS = [
         FILTER_VALIDATE_BOOLEAN => 'is_bool',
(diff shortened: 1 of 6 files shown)


henriquemoody:
I would be great if you could add the information you have in the pull request to the commit message, it helps with adding more context when looking at the git logs.

Personally, I prefer when the commit tells what it does in the subject line, and explain why in the body. This subject line tells 3 different things, and it would be best to have a subject line that expresses (in imperative mood) the overarching subject of the commit.

library/Validators/FilterVar.php
```diff
@@ -46,21 +45,22 @@ final class FilterVar extends Envelope
...
-
-        parent::__construct(new Callback(static function ($input) use ($filter, $arguments) {
-            return (self::ALLOWED_FILTERS[$filter])(
-                filter_var($input, ...$arguments),
-            );
-        }));
+    public function evaluate(mixed $input): Result
+    {
+        return Result::of(
+            (self::ALLOWED_FILTERS[$this->filter])(match ($this->options) {
```
henriquemoody, line 58: Is `match` the new ternary? 😂

Not asking to change, I just thought it was an interesting way to use it.

tests/unit/SerializableTest.php
```diff
@@ -0,0 +1,32 @@
...
+use function serialize;
+use function unserialize;
+
+#[Group('core')]
+final class SerializableTest extends TestCase
+{
+    use SmokeTestProvider;
+
+    #[DataProvider('provideValidatorInput')]
+    public function testValidatorSerialization(Validator $v, mixed $input): void
```
henriquemoody:
```suggestion
    [#Test]
    public function validatorSerialization(Validator $v, mixed $input): void
```

I wish there was a rule for that in PHP_CodeSniffer

library/Validators/FilterVar.php
```diff
@@ -33,7 +32,7 @@
     '{{subject}} must be valid',
     '{{subject}} must not be valid',
 )]
-final class FilterVar extends Envelope
+final class FilterVar implements Validator
```
henriquemoody:
```suggestion
final readonly class FilterVar implements Validator
```

We can make it readonly now. I will look for a PHP_CodeSniffer rule that does that when possible.

tests/unit/SerializableTest.php
```diff
@@ -0,0 +1,32 @@
...
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\TestCase;
+use Respect\Validation\Test\SmokeTestProvider;
+
+use function serialize;
+use function unserialize;
+
+#[Group('core')]
+final class SerializableTest extends TestCase
```
henriquemoody: This is not a unit test, and although it's arguable not a feature test either, it's best to put it there, maybe turn into a Pest test?

tests/unit/SerializableTest.php
henriquemoody: We might not even need to change that if we turn that into a feature test.

library/Validators/FilterVar.php
alganet: Our combination of static analyzers doesn't allow multi-line ternaries, which I do in fact prefer, like this:

```php
$isSomething
    ? whenTrue()
    : whenFalse()
```

It forces them to be inline, and then the line becomes too big for the line length rule. In this case, it doesn't justify creating methods to wrap the conditions. 

A match circumvents that, and I used it as a hack definitely! It will lead to the same opcodes as a ternary though.

tests/unit/SerializableTest.php
alganet: Can I reuse the data provider in the Pest test cleanly? If you know how to do it, let me know.

henriquemoody: You can use Datasets: https://pestphp.com/docs/datasets, maybe an extra array_map() around it, but about that

library/Validators/FilterVar.php
henriquemoody: Not criticising, I genuinely found it interesting, and I can see the benefits

tests/unit/SerializableTest.php
alganet: I found a way of reusing the trait, but it's kind of awkward and phpstan complains about it. I notice phpstan complains about a lot of Pest things, so I decided to go with it.

henriquemoody: PHPstan indeed complains a lot about Pest, I wish they had better ways of handling static analysis...

PR #1608, opened 2026-01-16
Update 02-feature-guide.md: correct validator name

bicpi:
The description refers to code above and there a `date()` validator was used, not `dateTime()`


alganet:
Nice catch! Thanks for your contribution 🐼

PR #1611, opened 2026-01-18
Prevent validators from generating `validatorBuilder` Ids

henriquemoody:
When calling `v::validatorName()`, a `ValidatorBuilder` is instantiated to manage the validation chain. However, users generally expect the ID to reflect the specific validator invoked rather than the internal builder class.

This commit updates `Id::fromValidator()` to resolve the ID from the underlying validator when it is the sole member of the `ValidatorBuilder`, ensuring more intuitive identification.

### DIFF
--- a/tests/unit/IdTest.php
+++ b/tests/unit/IdTest.php
@@ -1,0 +1,86 @@
+<?php
+
+/*
+ * SPDX-License-Identifier: MIT
+ * SPDX-FileCopyrightText: (c) Respect Project Contributors
+ * SPDX-FileContributor: Henrique Moody <henriquemoody@gmail.com>
+ */
+
+declare(strict_types=1);
+
+namespace Respect\Validation;
+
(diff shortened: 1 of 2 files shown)


tests/unit/IdTest.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+/*
+ * Copyright (c) Alexandre Gomes Gaigalas <alganet@gmail.com>
+ * SPDX-License-Identifier: MIT
+ */
```
alganet, line 7: `reuse lint` accepts this header because it's valid SPDX, but it's not up to the repository new standards.

Perhaps I should look into making a more comprehensive checker that would catch that sort of thing.

henriquemoody: It would be great if you take a look into that. Indeed, that should not have been allowed, it's only there because it was an old branch.

PR #1612, opened 2026-01-19
Make all remaining validators serializable

alganet:
This commit concludes the effort to make all current validators serializable by fixing the remaining ones.

The ability to use `finfo` instances on some filesystem validators was removed. `Image` was refactored to be a readonly class.

A command was added to the main `composer qa` flow that checks if all validators are covered by smoke tests (which are currently used by benchmarks and serialization tests). Therefore, this commit also ensures that every validator has a benchmark.


composer.json
```diff
@@ -73,22 +73,24 @@
         }
     },
     "scripts": {
+        "bench-profile": "vendor/bin/phpbench xdebug:profile",
```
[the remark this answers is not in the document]
alganet: It's fine, I'm trying to make it more consistent because we used `docs-fix` and not `docs:fix` before this one.

library/Validators/NoneOf.php
```diff
@@ -38,17 +34,24 @@ final class NoneOf extends Composite
...
         }
 
-        return Result::of($valid, $input, $this, [], $template)->withChildren(...$children);
+        return Result::of(
+            $failedCount === 0,
+            $input,
+            $this,
+            [],
+            count($children) === $failedCount ? self::TEMPLATE_ALL : self::TEMPLATE_SOME,
+        )->withChildren(...$children);
```
henriquemoody, line 55: What was the deal with NoneOf?

alganet: There were closures used by `array_map` and `array_reduce` that relied on the state of `$this->validators`. This behavior is not serializable.

alganet: By the way, we can totally use array_map and array_reduce and so on. In fact I could have refactored it to make the closures pure. However, this implementation is also better (only counts failures instead of mapping them, etc). Sometimes the functional stuff leads to overhead.

PR #1613, opened 2026-01-19
Introduce REUSE compliance

alganet:
This commit introduces REUSE compliance by annotating all files with SPDX information and placing the reused licences in the LICENSES folder.

We additionally removed the docheader tool which is made obsolete by this change.

The main LICENSE and copyright text of the project is now not under my personal name anymore, and it belongs to "The Respect Project Contributors" instead.

This change restores author names to several files, giving the appropriate attribution for contributions.


data/postal-code.php
```diff
@@ -222,9 +246,17 @@
...
+    ],
+    'HM' => [
+        '/^\d{4}$/',
+        '/^(\d{4})$/',
+    ],
     'HN' => [
-        '/^\w{2}\d{4}$/',
-        '/^([A-Z]{2}\d{4})$/',
+        '/^\d{5}$/',
+        '/^(\d{6})$/',
```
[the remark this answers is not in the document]
alganet: These come from the updated run of the geonames script. It's fine.

data/postal-code.php
```diff
@@ -515,8 +571,8 @@
         '/^(\d{6})$/',
     ],
     'RS' => [
-        '/^\d{6}$/',
-        '/^(\d{6})$/',
+        '/^\d{5}$/',
+        '/^(\d{5})$/',
```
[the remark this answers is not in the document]
alganet: These come from the updated run of the geonames script. It's fine.

docs/messages/placeholder-pipes.md
```diff
@@ -1,3 +1,8 @@
+<!--
+SPDX-FileCopyrightText: 2010-2026 Respect Project Contributors
+SPDX-License-Identifier: MIT
+-->
```
henriquemoody, line 4: Do we really need that here as well? I think it's really okay not to have this here.

src-dev/Markdown/SpdxHeader.php
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Respect\Dev\Markdown;
+
+use function array_shift;
+use function preg_match;
+
+trait SpdxHeader
```
henriquemoody: I don't think this belongs in this namespace

.gitattributes
```diff
@@ -1,3 +1,6 @@
+# SPDX-FileCopyrightText: 2010-2026 Respect Project Contributors
+# SPDX-License-Identifier: MIT
```
henriquemoody: I would say we don't need this here, not in the other .* files or configuration files

.gitignore
```diff
@@ -1,3 +1,6 @@
+# SPDX-FileCopyrightText: 2010-2026 Respect Project Contributors
+# SPDX-License-Identifier: MIT
```
henriquemoody: I would say we don't need this here

README.md
```diff
@@ -1,3 +1,7 @@
+<!--
+SPDX-FileCopyrightText: 2010-2026 Respect Project Contributors
+SPDX-License-Identifier: MIT
+-->
```
henriquemoody, line 4: I think we should not have that in the Markdown files.

REUSE.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[[annotations]]
+path = "composer.json"
+SPDX-FileCopyrightText = "2010-2026 Respect Project Contributors"
+SPDX-License-Identifier = "MIT"
+
+[[annotations]]
+path = "phpbench.json.dist"
+SPDX-FileCopyrightText = "2010-2026 Respect Project Contributors"
+SPDX-License-Identifier = "MIT"
```
henriquemoody: Often your files don't have a EOF, maybe your editor is misconfigured?

.github/dependabot.yml
```diff
@@ -1,3 +1,6 @@
+# SPDX-FileCopyrightText: 2010-2026 Respect Project Contributors
+# SPDX-License-Identifier: MIT
+
```
henriquemoody: I think we don't need this in the files in `.github` in fact, any file that configures a service/app/package.

src-dev/Markdown/SpdxHeader.php
alganet: Where would you put it?

docs/messages/placeholder-pipes.md
alganet: I don't make the choices here. It's `reuse lint` that does. There is no `.reuseignore`, you cannot omit licensing information (that's by design in the reuse tool).

You must think of this tool as `phpstan`, it removes friction by eliminating personal opinions.

What we can do is chose how to encode the licensing information. For example, this could be as an annotation in the `REUSE.toml` file. Either way, it must be somewhere.

.gitattributes
alganet: Again, not my choice. You should see `reuse` as a tool like `phpstan` or `phpcs`. It lints, I fix. No personal opinions matter.

src-dev/Markdown/SpdxHeader.php
henriquemoody: Since we're using that to handle the `Content`, I think this should be part of the content. You would then not even have a `prependSpdxToContent`, you would just do something like:

```diff
- $content = new Content();
+ $content = $file->content->extractSpdxHeader();
```

henriquemoody:
Resolved the comments about the Markdown files. I had a conversation with @alganet, and he made me realize that it's as important to add that to the documentation files as to the code.

src-dev/Markdown/SpdxHeader.php
alganet: I was not able to make this work in a single method.

There are changes to `$lines` that occour outside of the `Content` class. For example, in `ValidatorHeaderLinter`:

```php
        while (($line = array_shift($lines)) !== false) {
            if (preg_match('/^(# .+|- .+|)$/', $line) === 0) {
                array_unshift($lines, $line);
                break;
            }
        }
```

I could not make it a single method because of it. 

So, I definitely can move the methods to `Content`, but I was not able to make it a single clean API without further extensive refactorings.

Perhaps I didn't understood your idea or I'm missing something. If that's the case, can you clarify it?

REUSE.toml
```diff
@@ -0,0 +1,6 @@
+version = 1
+
+[[annotations]]
+path = [ "*.yml", "*.yaml", ".git*", "*.dist", "composer.json", "tests/fixtures/*", ".github/*.yml", ".github/workflows/**.yml" ]
```
henriquemoody, line 4: Why the tests/fixtures/?

henriquemoody (approved):

src-dev/Markdown/SpdxHeader.php
henriquemoody: Yeah, I'm not a big fan of those changes outside the Content, but I decided to be less picky with it since we don't ship this code. You can merge as is. I was [trying to make this a bit better](https://gist.github.com/henriquemoody/31903b92a38055a2207d2e4ee289ef40), but I don't know how much better it is since we're making changes outside the Content as you pointed out.

REUSE.toml
alganet: When you run `reuse lint`, all files need to have a license. Even binaries, or unrecognized files.

Stuff like our `valid-image.png` would need either this annotation, or a `valid-image.png.license` companion file just with the headers. I opted for the annotation instead of the companion file.

I just noticed that because of that, `data-provider.php` got a pass (it can not have SPDX headers). However, the alternative would be to list individual files here in the annotations. 

One good way of answering those "why this?" questions is to just remove (for example, remove the `tests/fixtures/*` item) and run `reuse lint`. You'll see it generates a list of files that are not REUSE compliant. Same with removing an SPDX header and so on.

I also recommend experimenting with `reuse annotate`, which is the tool for automatically adding SPDX headers to stuff. It's what I used to annotate most of the repo. It has a `--skip-unrecognized` option which I selected in order to not generate several kinds of `.license` companion files. Once I did that, I refined what it did manually, ensuring the linter would pass.

src-dev/Markdown/SpdxHeader.php
alganet: Nice, thanks for the diff! I decided to apply it because it leads to less lines of code.

PR #1615, opened 2026-01-21
Replace template rendering code with Respect\StringFormatter

henriquemoody:
I've moved almost all the code for placeholder replacement and parameter modifiers into an external library called Respect\StringFormatter. This approach allows us to evolve the template capabilities without making major changes to the Validation code.

This commit will introduce another dependency, `respect/string-formatter`, and will upgrade the version of `respect/string-formatter`, which simplifies our internal API greatly.

While making this change, I also updated how we generate exceptions. Instead of rendering the full message and the array of messages, we delegate that creation to the `ResultQuery`, which improves performance because we don’t need to render those big messages unless the user actually needs them.


library/Exceptions/ValidationException.php
```diff
@@ -18,22 +18,18 @@
...
-    /**
-     * @param array<string|int, mixed> $messages
-     * @param array<string>            $ignoredBacktracePaths
-     */
     public function __construct(
         string $message,
-        private readonly string $fullMessage,
-        private readonly array $messages,
-        array $ignoredBacktracePaths = [],
+        private readonly ResultQuery $resultQuery,
```
henriquemoody, line 31: I don't like this very much, to be honest, but it is a waste of compute to render those messages prematurely. Besides, these can be lots of messages. Perhaps if we introduce a cache later, we could put it back.

library/Message/InterpolationRenderer.php
```diff
@@ -10,38 +10,32 @@
...
+        private PlaceholderFormatter $formatter,
         private TemplateResolver $templateResolver,
     ) {
     }
 
     /** @param array<string|int, mixed> $templates */
     public function render(Result $result, array $templates): string
     {
-        $parameters = ['path' => $result->path, 'input' => $result->input, 'subject' => Subject::fromResult($result)];
+        $parameters = ['path' => $result->path, 'input' => $result->input, 'subject' => $result];
```
henriquemoody, line 30: The name `{{subject}}` make less sense now, as we don't have a `Subject` object anymore. But I found that the `Subject` class was really just proxying the `Result`, so I decided to just remove the overhead.

tests/feature/Validators/CallableTypeTest.php
```diff
@@ -27,5 +27,5 @@
     fn() => v::not(v::callableType())->assert(function (): void {
         // Do nothing
     }),
-    fn(string $fullMessage) => expect($fullMessage)->toBe('- `Closure {}` must not be a callable'),
+    fn(string $fullMessage) => expect($fullMessage)->toBe('- `Closure { fn(): void }` must not be a callable'),
```
henriquemoody, line 30: I think this looks much better.

tests/feature/TranslatorTest.php
```diff
@@ -9,25 +9,33 @@
 declare(strict_types=1);
 
 use Respect\Validation\ContainerRegistry;
-use Respect\Validation\Message\Translator;
-use Respect\Validation\Message\Translator\ArrayTranslator;
+use Symfony\Component\Translation\Loader\ArrayLoader;
+use Symfony\Component\Translation\Translator;
+use Symfony\Contracts\Translation\TranslatorInterface;
 
+$translator = new Translator('en');
```
henriquemoody, line 16: So, instead of keeping our own way of handling translations, I decided to use Synfony's one. It's very well known, and commonly used across different projects, so users will find the path of translating much easier.

library/Message/Stringifier/Handlers/PathHandler.php
```diff
@@ -1,30 +1,30 @@
...
 /*
  * SPDX-License-Identifier: MIT
  * SPDX-FileCopyrightText: (c) Respect Project Contributors
  * SPDX-FileContributor: Henrique Moody <henriquemoody@gmail.com>
  */
 
-namespace Respect\Validation\Message\Stringifier;
+declare(strict_types=1);
+
+namespace Respect\Validation\Message\Stringifier\Handlers;
```
alganet: Nitpick: I don't like the name "handler". I think it's too generic.

The `Placeholder` namespace is now free, and I think it actually represents what these classes (NameHandler, PathHandler, ResultHandler) do: they handle what happen to those placeholders in the message.

library/Exceptions/ValidationException.php
alganet: I actually prefer it this way.

alganet (dismissed):

library/Message/Stringifier/Handlers/PathHandler.php
henriquemoody: I moved them to `Respect\Validation\Message\Parameters`, I think it makes more sense now. What do you think?

PR #1622, opened 2026-01-26
Update message placeholder conversion documentation

henriquemoody:
This documentation had become outdated after significant changes to the codebase, including the replacement of the template rendering code with Respect\StringFormatter.


henriquemoody:
Merged manually

PR #1625, opened 2026-01-26
Adhere to GitHub Community Standards

alganet:
Makes the project more friendly to GitHub users by providing conventional files.

Those changes will improve our score under the GitHub community tab:

https://github.com/Respect/Validation/community

 - ~The CONTRIBUTING.md file was moved to the .github/ folder to remove clutter. This change is also in preparation to create more comprehensive developer documentation under a dedicated section in `docs/`~
 - The CHANGELOG.md file was removed. It was outdated and not being maintained.
 - Minimalistic .github/PULL_REQUEST_TEMPLATE.md
 - Minimalistic .github/SECURITY.md
 - Minimalistic .github/CODE_OF_CONDUCT.md
 - Minimalistic .github/ISSUE_TEMPLATES


.github/ISSUE_TEMPLATES/bug_report.md
```diff
@@ -0,0 +1,18 @@
...
+
+**To Reproduce**
+Steps to reproduce.
+
+**Expected behavior**
+What should happen.
+
+**Environment**
+- OS: 
+- PHP Version:
```
henriquemoody: Validation version, maybe?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
+## Type of Change
+
+- [ ] Bug fix
+- [ ] Refactoring
+- [ ] New feature
+- [ ] Breaking change
+- [ ] Documentation update
```
henriquemoody: I'm not sure we really want that, though? I find those are usually more annoying than helpful

alganet: I'll confess that I also find it annoying, and I added just to improve the community score.

I tried to be the least annoying as possible. For example, not having placeholders like "Describe the changes etc" that require deleting or reviewing the template.

This small checklist is added at the end of the PR commit message, and requires almost no interaction.

I haven't tested if GitHub accepts an empty template. It might be a worth test to do.

henriquemoody: I saw that Composer, for example, just has an empty file 🤷‍♂️

alganet:
@henriquemoody

I, personally, don't like these files very much. That's why I'm adding them to the `.github` folder. I see them as interfaces to the GitHub workflow, not actual docs.

As I mentioned in the commit description, my plan is to have a `docs/development` folder with richer information that we can link `.github -> docs` when needed, keeping the mandatory files to score high according to GitHub very short.

Do you prefer them in the root folder? I find that to add too much clutter, with UPPERCASED files that are too distracting.

.github/ISSUE_TEMPLATES/bug_report.md
alganet: Good catch.

henriquemoody:
Personally, I tend to look at the CONTRIBUTING.md file in repositories, and I think some AI agents also look for them.

If you feel strongly about having all those files at the root, I can compromise with all of them, but I would really want to keep the CONTRIBUTING.md file in the root folder.

.github/PULL_REQUEST_TEMPLATE.md
alganet: I'll use an empty file! I did not know that was allowed :D

henriquemoody:
I don't mid the uppercase files because LICENSE and README are also standard, and they're uppercase. More of the same to mee

alganet:
@henriquemoody makes sense, I moved the `CONTRIBUTING.md` file back.

henriquemoody:
You gotta edit your commit message, BTW

PR #1627, opened 2026-01-26
Lint Changelog format in validator docs

alganet:
Introduces a Markdown linter for checking the Changelog format.

"See Also" was transformed into a section to make it easier to handle it with the `Content` class. The "Related" linter was simplified to reflect that change too.

An additional "alignment" parameter was added to markdown table generators, allowing the padding and headers to be explicitly marked with a specific left (-1), middle (0) or right(1) alignment.

Existing files were fixed using the `fix` option after the changes.

### DIFF
--- a/src-dev/Markdown/Content.php
+++ b/src-dev/Markdown/Content.php
@@ -14,7 +14,9 @@ use ArrayIterator;
 use IteratorAggregate;
 use UnexpectedValueException;
 
+use function array_fill;
 use function array_find_key;
+use function array_keys;
 use function array_map;
 use function array_slice;
 use function count;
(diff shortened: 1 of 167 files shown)


src-dev/Markdown/Content.php
```diff
@@ -203,9 +207,10 @@ public function getSection(string $text): self
 
     /**
      * @param array<int, string>             $headers
+     * @param array<int, int>                $alignment
      * @param array<int, array<int, string>> $rows
      */
-    public function table(array $headers, array $rows): void
+    public function table(array $headers, array $alignment, array $rows): void
```
henriquemoody: I would have gone for the $alignment as a third, optional argument, that way one doesn't always need to be concerned about the alignment

alganet: That makes sense. I will update it.

PR #1629, opened 2026-01-26
Simplify how we load and save files in `data/`

henriquemoody:
We had different ways of saving and loading files from `data/`, so I decided to unify them to simplify things. I repurposed the `DomainInfo` class and named it `DataLoader`, so we can use the same class to load anything from the `data/` directory.


src-dev/Helpers/DataSaver.php
```diff
@@ -0,0 +1,44 @@
...
+    /** @param array<string|int, mixed> $data */
+    public function save(array $data, string $fileCopyrightText, string $licenseIdentifier, string $path): void
+    {
+        ksort($data);
+
+        $spdx = '// SPDX';
+        $fileContent = implode(PHP_EOL, [
+            '<?php declare(strict_types=1);',
+            $spdx . '-FileCopyrightText: ' . $fileCopyrightText,
+            $spdx . '-License-Identifier: ' . $licenseIdentifier,
```
alganet: FYI: I've been concatenating the SPDX prefix to avoid failing the linter.

The other way is to use `// REUSE-IgnoreStart` and `// REUSE-IgnoreEnd`, but I think it's silly.

henriquemoody: I know, I concatenated it for the same reason 😅

henriquemoody: I didn't know you could use the `// REUSE-Ignore*`, I think it's better so it's more explicit why we're doing this here

tests/unit/Helpers/DataLoaderTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+    #[Test]
+    public function shouldReturnDifferentDataWhenDifferentFilesAreLoaded(): void
+    {
+        $data1 = DataLoader::load('postal-code.php');
+        $data2 = DataLoader::load('domain/tld.php');
+
+        self::assertNotSame($data1, $data2);
+    }
+}
```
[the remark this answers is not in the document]
henriquemoody: I decided to settle for not testing this because otherwise I would either create functionality I don't need, or test using reflection, which I don't like either. I think this is not idea, but it's not the end of the world.

PR #1630, opened 2026-01-27
Create migration guide from v2 to v3

henriquemoody:
As we approach the release of version 3.0, it is important to provide our users with a clear guide to migrating from version 2.x to 3.0. This commit provides guidance by creating a document that will remain in the repository, so users can always access it.

### DIFF
--- a/docs/migrating-from-v2-to-v3.md
+++ b/docs/migrating-from-v2-to-v3.md
@@ -1,0 +1,941 @@
+<!--
+SPDX-FileCopyrightText: (c) Respect Project Contributors
+SPDX-License-Identifier: MIT
+-->
+
+# Migrating from v2.x to v3.x
+
+This guide will help you migrate your code from Respect\Validation 2.x to 3.0.
+
+Version 3.0 is a major release with breaking changes. This document is organized into two main sections: **Breaking Changes** (what you must update) and **New Features** (what you can now use).
+
+### Installation
+
+Update your `composer.json` to require version 3.0:
+
+```shell
+composer require respect/validation:^3.0
+```
+
+Respect\Validation 3.0 requires PHP 8.5 or above (2.x required PHP 8.1+).
+
+---
+
+## Breaking changes
+
+### Class and namespace changes
+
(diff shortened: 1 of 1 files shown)


docs/migrating-from-v2-to-v3.md
```diff
@@ -0,0 +1,919 @@
...
+    {
+        return $this->checkSomething($input);
+    }
+}
+```
+
+**After (3.0):** Single file with `#[Template]` attribute:
+
+```php
+use Respect\Validation\Message\Template;
```
henriquemoody, line 463: Shall we just move this to the `Respect\Validation` namespace?

docs/migrating-from-v2-to-v3.md
```diff
@@ -0,0 +1,941 @@
...
++ if (v::intType()->isValid($input)) {
+      // validation passed
+  }
+```
+
+#### `check()` removed, `assert()` unified
+
+In 2.x, there were two exception-based methods:
+
+- `check()` threw rule-specific exceptions (e.g., `IntTypeException`)
```
alganet, line 74: The main goal of `check` wasn't the fact that it returned a rule-specific exception, it was the fact that it failed fast. As soon as it got one exception, it stopped probing for more.

We should point out that global fail fast is not available, and we use `Circuit` to apply the same idea to individual chains instead.

docs/migrating-from-v2-to-v3.md
```diff
@@ -0,0 +1,941 @@
...
++} catch (ValidationException $exception) {
+    echo $exception->getMessage();
+}
+```
+
+The `ValidationException` provides all methods previously split between exceptions:
+
+```diff
+-use Respect\Validation\Exceptions\NestedValidationException;
++use Respect\Validation\Exceptions\ValidationException;
```
alganet, line 96: We should probably point out that all specific exceptions were removed, and you cannot catch `IntValException` (for example) directly anymore. In practice, it's like always having `NestedValidationException`.

PR #1631, opened 2026-01-27
Fix ResultQuery::findByPath() for nested paths

henriquemoody:
The `findByPath()` method was failing to return results when using nested dot-notation paths such as `user.email` or `items.1`. However, it’s returning `null` instead of the expected result in some cases.

The root cause was a mismatch between how paths are stored vs searched:

- Storage: Validators like Key and Each create results where the path is stored as a linked list. For `user.email`, the "email" result has `path="email"` with `parent="user"`.

- Search (old): The method expected a tree structure where it would find a child with `path="user"`, then search that child for `path="email"`. But no child had `path="user"` - only "email" (with "user" as its parent).

The fix computes each result's full path by walking up the parent chain and compares it against the search path. Also converts numeric strings to integers when parsing paths (e.g., `items.1` → `['items', 1]`) since array indices are stored as integers.

While working on this fix, I also realised that to expose the result's status, it’s best to use `hasFailed()` instead of `isValid()` in `ResultQuery`, since users will mostly use results when validation failed, not when it passed.

### DIFF
--- a/src/ResultQuery.php
+++ b/src/ResultQuery.php
@@ -15,9 +15,11 @@ use Respect\Validation\Message\Renderer;
 use Respect\Validation\Message\StringFormatter;
 use Stringable;
 
-use function array_shift;
+use function array_find;
+use function array_map;
+use function array_reverse;
+use function ctype_digit;
 use function explode;
-use function implode;
 use function is_string;
 
 final readonly class ResultQuery implements Stringable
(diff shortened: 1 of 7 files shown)


src/ResultQuery.php
```diff
@@ -71,32 +73,18 @@ public function findByName(string $name): self|null
...
-        return null;
+        return match ($result) {
+            null => null,
+            $this->result => $this,
+            default => clone ($this, ['result' => $result]),
+        };
     }
 
-    public function isValid(): bool
+    public function hasFailed(): bool
```
henriquemoody, line 85: I'm coming to the realization that we shouldn't even have that method, instead `validate()` returns null or the failed result. What do you think, @alganet?

alganet: It makes interfaces awkward:

```
// Would return null, which juggles to false, misleading since I expect `validate` to enter the if passed
if ($obj->validate()) {
}
```

IMHO `ResultQuery` should be compatible with `Result`. All things you could do with `Result`, you could also do with `ResultQuery`, like getting `->hasPassed`. This makes implementation weirder (likely a proxy of some kind), but the interface cleaner and would unlock things (if we could use them interchangeably).

There are tools we could use to achieve that, such as PHP's new property hooks: https://www.php.net/manual/en/language.oop5.property-hooks.php

I honestly don't know what the best course of action is here.

henriquemoody: I gotta think a bit more about that. I'm not sure what's the best course of action either. Perhaps we should just give it time, and see what users want from `ResultQuery` after we rollout 3.0.

PR #1632, opened 2026-01-27
Streamline validators.md index

alganet:
Makes it so the index looks more like a cheatsheet, condensing information instead of making long lists that require lots of
scrolling to explore.

Additionally, the happy path for each validator was also added, providing a quick reference use for comparison.

The direct markdown links were replaced by titled markdown references, offering mouse-over tooltips over links that
display the validator one-line description.

To ensure a proper source of truth for these new index goodies, the AssertionMessageLinter was modified to verify that the first assertion in each doc is a single-line validator that passes (a happy path), further making our documentation conventions more solid.

---

### Cheatsheet-style and tooltips on mouse-over

image

### Alphabetical list with happy path examples

image

### Plain text

Viewing the document in plain-text gives almost the same feeling as the rendered markdown, respecting the true spirit of the format.

image

### DIFF
--- a/src-dev/Markdown/Linters/AssertionMessageLinter.php
+++ b/src-dev/Markdown/Linters/AssertionMessageLinter.php
@@ -42,6 +42,7 @@ use function str_contains;
 use function str_ends_with;
 use function str_replace;
 use function str_starts_with;
+use function strpos;
 use function substr;
 use function substr_count;
 use function trim;
@@ -73,6 +74,7 @@ final readonly class AssertionMessageLinter implements Linter
         $newContent = new Content();
         $inCodeBlock = false;
         $currentCodeBlock = [];
+        $isFirstCodeBlock = true;
 
         foreach ($content as $line) {
             if (str_starts_with($line, '```php')) {
@@ -83,7 +85,7 @@ final readonly class AssertionMessageLinter implements Linter
 
             if ($inCodeBlock && str_starts_with($line, '```')) {
                 $currentCodeBlock[] = $line;
-                $processedLines = $this->processCodeBlock($currentCodeBlock, $context);
+                $processedLines = $this->processCodeBlock($currentCodeBlock, $context, $isFirstCodeBlock);
 
                 foreach ($processedLines as $processedLine) {
                     $newContent->raw($processedLine);
@@ -91,6 +93,7 @@ final readonly class AssertionMessageLinter implements Linter
 
                 $inCodeBlock = false;
                 $currentCodeBlock = [];
+                $isFirstCodeBlock = false;
                 continue;
             }
 
@@ -110,14 +113,14 @@ final readonly class AssertionMessageLinter implements Linter
      *
      * @return string[]
      */
-    private function processCodeBlock(array $lines, ArrayObject $context): array
+    private function processCodeBlock(array $lines, ArrayObject $context, bool $isFirstCodeBlock): array
     {
         $openingLine = $lines[0];
         $closingLine = $lines[count($lines) - 1];
         $codeLines = array_slice($lines, 1, -1);
         $code = implode("\n", $codeLines);
 
-        $processedCodeLines = $this->processCode($code, $context);
+        $processedCodeLines = $this->processCode($code, $context, $isFirstCodeBlock);
 
         return [
             $openingLine,
@@ -127,7 +130,7 @@ final readonly class AssertionMessageLinter implements Linter
     }
 
     /** @return string[] */
-    private function processCode(string $code, ArrayObject $context): array
+    private function processCode(string $code, ArrayObject $context, bool $isFirstCodeBlock): array
     {
         $cleanedCode = $this->stripOutputComments($code);
 
@@ -146,6 +149,7 @@ final readonly class AssertionMessageLinter implements Linter
         $processedLines = [];
         $lastEndPos = $phpTagLength;
         $lastWasAssertion = false;
+        $firstAssertionEncountered = false;
 
         foreach ($statements as $statement) {
             $startPos = $statement->getStartFilePos();
@@ -174,6 +178,34 @@ final readonly class AssertionMessageLinter implements Linter
                 $processedLines[] = $statementCode;
                 $contextHasClass = $this->contextContainsClass($context);
                 $result = $this->executeStatement($statementCode, $context, $printer);
+
+                if ($isFirstCodeBlock && !$firstAssertionEncountered) {
+                    // Enforce one-liner for the first assertion in the first code block
+                    if (strpos($statementCode, "\n") !== false) {
+                        throw new UnexpectedValueException(sprintf(
+                            'The first assertion must be a single-line assertion. Statement: %s',
+                            trim($statementCode),
+                        ));
+                    }
+
+                    $validated = false;
+                    foreach ($result as $line) {
+                        if (trim($line) === '// Validation passes successfully') {
+                            $validated = true;
+                            break;
+                        }
+                    }
+
+                    if (!$validated) {
+                        throw new UnexpectedValueException(sprintf(
+                            'The first assertion in the first code block must validate successfully. Statement: %s',
+                            trim($statementCode),
+                        ));
+                    }
+
+                    $firstAssertionEncountered = true;
+                }
+
                 $processedLines = array_merge($processedLines, $result);
 
                 if ($contextHasClass) {
--- a/src-dev/Markdown/Linters/ValidatorIndexLinter.php
+++ b/src-dev/Markdown/Linters/ValidatorIndexLinter.php
@@ -17,7 +17,9 @@ use Symfony\Component\Finder\Finder;
 
 use function array_keys;
 use function dirname;
+use function implode;
 use function ksort;
+use function preg_match;
 use function sort;
 use function sprintf;
 use function str_ends_with;
@@ -38,15 +40,19 @@ final readonly class ValidatorIndexLinter implements Linter
         $validators = $this->getSortedListOfValidators();
 
         $validatorsByCategory = [];
+        $validatorsDescriptions = [];
+        $validatorsExamples = [];
         foreach ($validators as $validator) {
-            foreach ($this->getCategoriesByValidator($validator) as $category) {
+            $validatorInfo = $this->getValidatorInfo($validator);
+            foreach ($validatorInfo['categories'] as $category) {
                 $validatorsByCategory[$category] ??= [];
                 $validatorsByCategory[$category][] = $validator;
+                $validatorsDescriptions[$validator] = $validatorInfo['description'];
+                $validatorsExamples[$validator] = $validatorInfo['example'];
             }
         }
 
         ksort($validatorsByCategory);
-        $validatorsByCategory['Alphabetically'] = $validators;
 
         $categories = array_keys($validatorsByCategory);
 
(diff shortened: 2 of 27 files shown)


src-dev/Markdown/Linters/ValidatorIndexLinter.php
```diff
@@ -88,10 +111,22 @@ private function getSortedListOfValidators(): array
...
+        foreach ($content->getSection(sprintf('# %s', $validator)) as $line) {
+            if ($description === '' && preg_match('/^[A-Z]/', $line) === 1) {
+                $description = $line;
+            }
+
+            if (preg_match('/^v::/', $line) === 1) {
+                $example = $line;
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
alganet: That has been fixed in the last commit revision by ensuring all docs have proper source of truth.

src-dev/Markdown/Linters/AssertionMessageLinter.php
```diff
@@ -174,6 +178,34 @@ private function processCode(string $code, ArrayObject $context): array
                 $processedLines[] = $statementCode;
                 $contextHasClass = $this->contextContainsClass($context);
                 $result = $this->executeStatement($statementCode, $context, $printer);
+
+                if ($isFirstCodeBlock && !$firstAssertionEncountered) {
+                    // Enforce one-liner for the first assertion in the first code block
+                    if (strpos($statementCode, "\n") !== false) {
```
henriquemoody, line 184: I think those changes shouldn't be in this linter, because this linter is supposed to only handle assertion executions, it doesn't need to be validator specific. You don't need executions to do this check, so I would prefer to have it as a whole other linter.

At some point I would like to run it in all documentation files, to make sure all examples produce the expected results, not only with `assert()` but maybe event with `echo`.

henriquemoody (approved):
I love the new `validators.md`, and I the linter for the first assertion is really great, regardless of how they look in `validators.md`.

I made a comment about the first code block change, but I won't block this merge -- nothing we couldn't change afterwards, and since that's not code we ship in the package we can change it at any time.

src-dev/Markdown/Linters/AssertionMessageLinter.php
alganet: That would be a nice improvement! I'm all up for it.

If you look closely, my changes are not validator specific. They are *assertion specific* though.

What I'm doing is checking if the first statement in the page is a one-liner and a happy path. 

For example, if in the future you expand it to run in other pages, such as `comparable-values.md`, it would still work, as long as the first example in that page is also "a simple one" (one-liner and a happy path).

The idea here is that the document should start in a simple way, then it can increase in complexity afterwards.

If you change this to also consider `echo $message` assertions, we can change what "happy path" means without many structural modifications. Perhaps just expand `if (trim($line) === '// Validation passes successfully') {` to more conventional scenarios.

If you are talking about checking all kinds of code blocks, not only assertions, then I think *that* is another completely different linter that doesn't have `Assertion` in its name.

henriquemoody: I think they're validator-specific because we would like that the first example _in the validator docs_ would be a happy path.

In other documentation files, that might not be the case, because we wouldn't be demonstrating how a specific validator works, but more how the code behaves. Take the migration guide, or the feature guide. There's so much code there, and we can guarantee it's working properly.

I could agree that checking all kinds of code blocks is another linter, but it seems like it closer to what the assertion does now (because it executes the code and ensure the comment bellow matches the expectation), than checking if the first example is a happy path. We can disagree on that, but I hope we can agree that we don't want _every_ document to have a happy as the first example, because that won't work with any document.

alganet: > I think they're validator-specific because we would like that the first example in the validator docs would be a happy path.

That is what the commit message says, but look at the code instead. There is nothing validator-specific about it except what it was already there.

> Take the migration guide, or the feature guide.

Both of these pages have a simple validation as their first assertion `php` code block. These could easily pass my code without many changes.

> we don't want every document to have a happy as the first example

Have you actually ran my changes in those scenarios you're implying? Modify the previously existing constructor to run on `translation.md`. You will see that non-assertions are skipped. It does not complain about the block having the container registry setup (because that's not the first assertion). In fact, that page has no assertions so my code wouldn't even trigger.

---

I believe this idea to process all non-assertion code blocks is a major refactor and out of scope.

Furthermore, my changes do not block that refactor to happen. They're localized and work within existing structures, such as `isValidationAssertion` (existing method). 

In implementation terms, it would be a waste to replicate all the code-running infrastructure into another linter. Making another linter, therefore, is an architectural change (it would require abstracting out the php runner code), that is super large in scope and makes no sense for this PR.

In terms of usage and documentation conventions, not Linter implementation, I also think my changes are sound. They already skip blocks that are not assertions (your main complaint), and do not block these non-assertion blocks to be inspected in the future.

henriquemoody: The fact that it runs, doesn't mean that's what we want, it's about the intention of the linter. Apart from a validator's documentation, I don't want to constraint myself into having the first assertion to be a happy path.

You do no need to run the code to lint that the first assertion is a happy path for the validators docs: you can just assume that the first example needs to be a single line and have a `// Validation passes successfully` bellow it. Where the `AssertionMessageLinter` really shines is on running the code, and validating whether the assertions run as we expect them to run.

In any case, you already have my approval, you can just merge this anyways

alganet: > I don't want to constraint myself into having the first assertion to be a happy path

I was skeptical of this choice myself, but I could not think of a scenario that would hurt by having this.

Can you explain better what are you afraid of? I understand the feeling, but actual examples speak more than words.

> you can just assume that the first example needs to be a single line and have a // Validation passes successfully bellow it.

I see what you mean, but that creates a soft dependency between the two linters. This new linter would "depend" on the code runner one, but that dependency would not be explicit in the code.

If you run it _standalone_, then all sorts of things would pass:

```php
foo(
$bar);
```

Personally, I think that's bad design, and linters should have encapsulated behavior. That way, we can change one without having to worry about its soft dependencies with another. I know we don't have unit tests for the linters yet, but think from this testable approach. The only think I could really guarantee is that the second line is a comment. Without some deeper introspection on the first line (at least ensuring it's a valid statement), this linter would not stand on its own.

> In any case, you already have my approval, you can just merge this anyways

I appreciate the vote of confidence, but my priority is getting a good design, not pushing the change forward. I am ok to wait and sort this out.

henriquemoody: > that creates a soft dependency between the two linters. This new linter would "depend" on the code runner one, but that dependency would not be explicit in the code.

I see that more of a dependency upon a convention then a dependency between two linters.

In the example you gave, the linter would not pass, because the code above `// Validation passes successfully` is not an assertion, but a variable declaration.

The intent of the linters are different because:

* `AssertionMessageLinter`: verifies if the assertions in the documentation have the expected behavior when they ran with PHP (maybe the current name is just not descriptive for what it does)
* `AssertionOrderLinter`: verifies if the first assertion in the documentation is a single line, and provides a successful assertion (not sure about the name either, but that's the one I'm suggesting to split)

> Can you explain better what are you afraid of?

I'm afraid that I'm writing a documentation about about something and I get a message that the linter is failing because the first assertion in my documentation is not a single line with a passing case. I would then need to reframe how I'm thinking about this document, just because the linter says so, or add it to the ignore list.

alganet: > In the example you gave, the linter would not pass, because the code above // Validation passes successfully is not an assertion, but a variable declaration.

That is **precisely what I'm saying** that does not sit right with me.

Again, think of the unit test scenario I mentioned. My `foo( // comment` example would pass the unit tests, showcasing the issue with behavior encapsulation.

Linters are a class of software that often do not rely on violations depending on each other. 

In PHPStan, for example, you can disable one rule and keep other rules running. That is also implied in our architecture by `CompositeLinter` accepting a list of rules you can pick and choose.

That can only be guaranteed if each rule stands on on its own. If we disable `AssertionMessageLinter`, then `AssertionOrderLinter` would stop working as intended (and would "pass" my foo comment counter-example).

Again, I understand your way introduces modularity, but that modularity comes at the cost of rule isolation.

> I'm afraid that I'm writing a documentation about about something and I get a message that the linter is failing because the first assertion in my documentation is not a single line with a passing case.

And you think extracting a linter into another class is the only solution for that? 

Let's not mix implementation discussion with our repository conventions. There are many ways to prevent that from happening that do not imply having a separate linter class. I hinted at one of them: having a better abstraction that reuses code runs.

It seems this decision is for later (a scenario that hasn't happened yet), and an easy fix. I'm just not convinced that having a separate linter preemptively is a good fix.

henriquemoody: > and would "pass" my foo comment counter-example

Whether it would ou would not pass would depend on the implementation we have.

I understand that on `AssertionOrderLinter` you want to parse the code, and make sure it's an assertion. Where we're diverging, I guess, is that you want to run the code to make sure it's a passing assertion (while I would be happy to just parse the code), and that the `AssertionMessageLinter` already does, so it seems like the logical place to keep it, but even though they do similar things, they could still be different linters and have no dependency with each other at all.

* AssertionMessageLinter: execute all assertions, make sure each one of them match the expected behavior
* AssertionOrderLinter: execute only the first assertion, make sure it's a 1-line & passing assertion

I enjoy the discussion, but this seems to be a decision for later indeed. I like having a good design, but I think we don't need to be so concerned with the design of those linters since they're not code we shit to our users, but things we use. The code your shipping solves the problem we have right now, and if things happen to change later, we can always revisit.

alganet: > we can always revisit.

I totally agree. We do have to keep aligned, however. This kind of discussion shapes how we think about future linters, and prevents us from creating implementations with different disparing competing visions.

If I were to merge as it is without making my opinion explicit, you might have assumed that I was aligned with the idea of a separate linter. However, as you can see, I'm not (or at least I don't see it as that simple). I don't want any decision to float implicitly. 

I take PR comments seriously, especially on OSS software that might have lots of eyes on it. This also helps future potential contributors to understand our goals.

henriquemoody: > If I were to merge as it is without making my opinion explicit, you might have assumed that I was aligned with the idea of a separate linter

I might have, but that wouldn't be the case. In fact, I assumed the opposite, that we were _not_ aligned. To me, the weight of decisions are proportional how irreversible they are, and I consider this case to be a very light one.

Alignment is important, and we need to keep aligned with the decisions we make here, but for the linters we have quite some wiggle room since the design of the linters would scarcely change the developer experience, which is the reason why they exist in the first place.

The design of linters would influence how we contributors create/change them, but that's secondary to this project. We had other linters and other developer-specific code running in Bash script for years and only you and I changed those.

> I take PR comments seriously, especially on OSS software that might have lots of eyes on it. This also helps future potential contributors to understand our goals.

I appreciate you take them seriously, and so do I. To me, though, in this case, this discussion influences Validation very little at its core. But, had we decided to turn those linters into a standalone project, I would say the weight of the decisions was higher, and we would have to come to an agreement (or commitment).

I see that the discussion about the linters helps future potential contributors to understand our goals about how to run the project, but not so much about the goals we have for Validation itself.

PR #1633, opened 2026-01-28
Make ValidatorBuilder::isValid fail fast implicitly

alganet:
The evaluation methods of `ValidatorBuilder`, namely `evaluate` and `isValid`, are very similar.

Each of them performs a full evaluation, only differing in their return.

Whereas `evaluate` returns a full `Result` object, `isValid` only returns a boolean.

As it turns out, if the user is interested only in the boolean, we do not need to gather all `Result` data.

Here we introduce a new interface, `IsValid`, that defines a method to skip all kinds of message generation for such cases.

Fail-fast scenarios such as `isValid` can leverage early failed results to stop the chain when only a boolean is needed.

This results in 70% performance improvement for chains with 10 nodes, and the gains increase with chain size. The more nodes,
the faster this change makes. Chains with 100 nodes can gain up to 90% performance compared to the previous implementation.

A benchmark was added to ensure these gains remain in future library iterations.

Furthermore, this change is only internal and backwards-compatible, making no public interface changes that would affect how users interact with the library.

TL;DR makes ValidatorBuilder::isValid super fast.

---

### Local benchmark results

**BEFORE**
image

**AFTER**
image

### DIFF
--- a/src/ValidatorBuilder.php
+++ b/src/ValidatorBuilder.php
@@ -56,13 +56,7 @@ final readonly class ValidatorBuilder implements Validator, Nameable
 
     public function evaluate(mixed $input): Result
     {
-        $validator = match (count($this->validators)) {
-            0 => throw new ComponentException('No validators have been added.'),
-            1 => current($this->validators),
-            default => new AllOf(...$this->validators),
-        };
-
-        return $validator->evaluate($input);
+        return $this->getEvaluationTarget()->evaluate($input);
     }
 
     /** @param array<string|int, mixed>|string|null $template */
@@ -73,7 +67,13 @@ final readonly class ValidatorBuilder implements Validator, Nameable
 
     public function isValid(mixed $input): bool
     {
-        return $this->evaluate($input)->hasPassed;
+        $validator = $this->getEvaluationTarget();
+
+        if ($validator instanceof IsValid) {
+            return $validator->isValid($input);
+        }
+
+        return $validator->evaluate($input)->hasPassed;
     }
 
     /** @param array<string|int, mixed>|callable(ValidationException): Throwable|string|Throwable|null $template */
@@ -116,6 +116,15 @@ final readonly class ValidatorBuilder implements Validator, Nameable
         }
 
         return null;
+    }
+
+    private function getEvaluationTarget(): Validator
+    {
+        return match (count($this->validators)) {
+            0 => throw new ComponentException('No validators have been added.'),
+            1 => current($this->validators),
+            default => new AllOf(...$this->validators),
+        };
     }
 
     /** @param array<string|int, mixed>|string|null $template */
--- a/src/Validators/AnyOf.php
+++ b/src/Validators/AnyOf.php
@@ -15,6 +15,7 @@ declare(strict_types=1);
 namespace Respect\Validation\Validators;
 
 use Attribute;
+use Respect\Validation\IsValid;
 use Respect\Validation\Message\Template;
 use Respect\Validation\Result;
 use Respect\Validation\Validator;
@@ -28,7 +29,7 @@ use function array_reduce;
     '{{subject}} must pass at least one of the rules',
     '{{subject}} must pass at least one of the rules',
 )]
-final class AnyOf extends Composite
+final class AnyOf extends Composite implements IsValid
 {
     public function evaluate(mixed $input): Result
     {
@@ -40,5 +41,24 @@ final class AnyOf extends Composite
         );
 
         return Result::of($valid, $input, $this)->withChildren(...$children);
+    }
+
+    public function isValid(mixed $input): bool
+    {
+        foreach ($this->validators as $validator) {
+            if ($validator instanceof IsValid) {
+                if ($validator->isValid($input)) {
+                    return true;
+                }
+
+                continue;
+            }
+
+            if ($validator->evaluate($input)->hasPassed) {
+                return true;
+            }
+        }
+
+        return false;
     }
 }
--- a/src/Validators/NoneOf.php
+++ b/src/Validators/NoneOf.php
@@ -15,6 +15,7 @@ declare(strict_types=1);
 namespace Respect\Validation\Validators;
 
 use Attribute;
+use Respect\Validation\IsValid;
 use Respect\Validation\Message\Template;
 use Respect\Validation\Result;
 use Respect\Validation\Validators\Core\Composite;
@@ -32,7 +33,7 @@ use function count;
     '{{subject}} must pass all the rules',
     self::TEMPLATE_ALL,
 )]
-final class NoneOf extends Composite
+final class NoneOf extends Composite implements IsValid
 {
     public const string TEMPLATE_ALL = '__all__';
     public const string TEMPLATE_SOME = '__some__';
@@ -41,8 +42,8 @@ final class NoneOf extends Composite
     {
         $failedCount = 0;
         $children = [];
-        foreach ($this->validators as $validator) {
-            $child = $validator->evaluate($input)->withToggledModeAndValidation();
+        foreach ($this->validators as $child) {
+            $child = $child->evaluate($input)->withToggledModeAndValidation();
             $children[] = $child;
             if ($child->hasPassed) {
                 continue;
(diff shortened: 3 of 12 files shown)


src/ValidatorBuilder.php
```diff
@@ -118,6 +118,15 @@ public function getName(): Name|null
         return null;
     }
 
+    private function getEvaluationTarget(): Validator
+    {
+        return match (count($this->validators)) {
+            0 => throw new ComponentException('No validators have been added.'),
+            1 => current($this->validators),
+            default => new AllOf(...$this->validators),
```
henriquemoody, line 126: If the goal is improving the performance (or simplify) of the `isValid()` method, we could do this simply by using `Circuit` instead of `AllOf` here.

I like what you're doing with `Generative`, and I see that with it we would still have performance improvements for other "composite" rules (like `AllOf`, `AnyOf`), but then I would argue that what we're having in `Generative:::isValid()` should be something we use in `Circuit` uses

We could just have a different Rule interface (`FailFastRule`) what would have another method (`evaluateWithFailFast()`) so validators would implement that method and `Circuit` would use them. That way we make `Circuit` a true fail-fast validator.

PS.: The names I suggested are just illustrative, I don't think they're good.

alganet: `Circuit` does not apply itself recursively through all its children. If there's an AllOf inside it, that `AllOf` will be fully evaluated.

`Generative` goes deep.

henriquemoody: I know, that's why I'm suggesting this solution. Something like this:

```php
// FailFastRule.php
interface FailFastRule
{
    public function evaluateWithFailFast(mixed $input): Result;
}

// AllOf.php
final class AllOf extends Composite implements FailFastRule
{
    // ...
    public function evaluateWithFailFast(mixed $input): Result
    {
        foreach ($this->validators as $validator) {
            $result = $validator instanceof Validator
                ? $validator->evaluateWithFailFast($input)
                : $validator->evaluate($input);

            if (!$result->hasPassed) {
                return $result;
            }
        }

        return $result;
    }
}

// Circuit.php
final class Circuit extends Composite
{
    public function evaluate(mixed $input): Result
    {
        foreach ($this->validators as $validator) {
            $result = $validator instanceof Validator
                ? $validator->evaluateWithFailFast($input)
                : $validator->evaluate($input);

            if (!$result->hasPassed) {
                return $result;
            }
        }

        return $result;
    }
}


// ValidatorBuilder.php

final readonly class ValidatorBuilder implements Validator, Nameable
{
    // ...

    public function isValid(mixed $input): bool
    {
        if ($this->validators === []) {
            throw new ComponentException('No validators have been added.');
        }

        $validator = new Circuit(...$this->validators);

        return $validator->evaluate($input)->hasPassed;
    }
    // ...
```

src/Validators/AnyOf.php
```diff
@@ -41,4 +41,19 @@ public function evaluate(mixed $input): Result
...
+    public function isValid(mixed $input): bool
+    {
+        foreach ($this->validators as $validator) {
+            if ($validator instanceof ResultGenerator && $validator->isValid($input)) {
+                return true;
+            }
+
+            if ($validator->evaluate($input)->hasPassed) {
+                return true;
+            }
```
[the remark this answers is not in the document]
alganet: This has been fixed.

src/Validators/NoneOf.php
```diff
@@ -59,4 +60,19 @@ public function evaluate(mixed $input): Result
...
+    public function isValid(mixed $input): bool
+    {
+        foreach ($this->validators as $validator) {
+            if ($validator instanceof ResultGenerator && $validator->isValid($input)) {
+                return false;
+            }
+
+            if ($validator->evaluate($input)->hasPassed) {
+                return false;
+            }
```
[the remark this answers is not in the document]
alganet: This has been fixed.

src/ValidatorBuilder.php
henriquemoody: The major gain is that `Circuit` becomes a REAL fail fast validator, and it's not so different from what you're doing.

alganet: > `public function evaluateWithFailFast(mixed $input): Result;`

I see your suggestion returns `Result`. What is the goal here? Build a partial `Result` instance?

I solved this by interfacing to boolean, so I don't have to deal with partials.

> Circuit: `? $validator->evaluateWithFailFast($input)`
> [...]
> ValidtorBuilder: `$validator = new Circuit(...$this->validators);`

Seems like you're only wrapping the first level of chains in this solution. Circuit can see other Circuits, but it's up to the user to wrap every single of the inner ones, or make sure they're at the beginning of the chain.

Consider this:

```
v::arrayVal()->allOf(...
```

In this case, the composite is not the first node in the chain. It would be wrapped by `Circuit`, but then, how can you fail fast the second `allOf` node?

I'm not sure we need an extra wrapping around `Circuit` every time we find a `FailFastRule`. I just decided to make that mechanism part of the interface itself.

This way, `Generative` becomes an inner path for getting booleans fast. `Circuit` remains as an useful tool for managing individual fast-fail reports with messages. There is no way to combine both (getting true fail fast and full Report instances at the same time).

---

Maybe I misunderstood your suggestion. My main blocker here is dealing with potentially partial `Result` instances. That is what I'm most curious about your suggestion, how to deal with that.

alganet: I finished the implementation and tests, and the PR is not an incomplete draft anymore. I added you as a reviewer.

henriquemoody: I'm suggesting to use `Circuit` as a default only on `isValid()`, not on `assert()`, because indeed, it's up to the user to define how they'd like the results to be shown.

If we consider the following chain:
```php
v::arrayVal()->allOf(...)
```

when we would use `isValid()` (not `assert()`), it would be translated into something like:

```php
new Circuit(new ArrayVal(), new AllOf(...))
```

BUT, because `Circuit` calls `FailFastValidator:: evaluateWithFailFast()`, `AllOf` would fail-fast if one of its validators fail. Admittedly, that works best with `AllOf` and `NoneOf` than with the other ones, because returning partial results in `AnyOf` and `OneOf` can confusing for the user, because we would indeed return "partial results", and the message might not make sense with `AnyOf` and `OneOf`.

I meant "partial" because in the context of `AllOf` and `NoneOf` each child result makes sense on its own, not so much for `AnyOf` and `OneOf`

> What is the goal here? Build a partial `Result` instance?

Your changes sparkle some ideas, and I thought we could indeed return "partial" `Result` instances. A better way of saying it is to return the child result directly, not the parent (in some cases)

With the changes I'm suggesting `Circuit` can fail-fast not only with the validators we give it, but those validators themselves could also fail-fast (even the `Each` validator). If someone has a `v::circuit(v::intVal(), v::allOf(v::positive(), v::greaterThan(3)))->assert(-1)`, the `AllOf` validator would never evaluate `v::greaterThan(3)`, and instead return the result of `Positive()`, but only because `Circuit` is the main wrapper.

henriquemoody: It's fair if we don't want to have a `Circuit` as a wrapper the whole time, we could call `evaluateWithFailFast()` directly on `isValid()`

alganet: I experimented locally with partial `Result` instances. It's a deep rabbit hole.

Ultimately, we could make `Result` completely lazy (it already carries the Validator instance it was obtained from), using property hooks for `$hasPassed` as I mentioned previously. This PR advances the internal engine towards that goal. The immutable list of children is a design issue though. It only goes as far as partials.

The truth is that getting a boolean will _always_ be faster, even than partial or lazy `Result` instances. And it's simpler to implement, so it makes a lot of sense (to me) to start with a small measurable deliverable that does that. No changes to public interface, no complications, clear use case for all sorts of scenarios.

> even the Each validator

That is a separate refactor, improving `FilteredNonEmptyArray` to deal with iterables (basically, making it `FilteredNonEmptyIterable`). There are challenges with no-rewind style iterables though.

alganet: @henriquemoody

I have updated the PR to only feature the boolean `IsValid` interface, and not deal with any generators, leaving space for us to decide on how to deal with result streaming later.

This still offers a significant performance improvement for obtaining boolean only results, 

`if (v::foo()->bar()->baz()->validate()` is very a common usage pattern seeing in our users across GitHub. Examples:

https://github.com/search?q=%22Respect%5CValidation%22+AND+%22v%3A%3A%22+AND+%22validate%28%22&type=code

This also opens up possibilities for modes, such as  `--verbose/--quiet` in (but not limited to) CLI apps, getting both performance in the quiet mode and throroughness in the verbose mode.

With this, you are free to implement your ideas around `Circuit`, as long as the benchmarks for single boolean do not display any performance regression (I would accept a small performance regression actually).

alganet:
This was superseeded by #1663

PR #1636, opened 2026-01-28
Fixes git workspace after checking out benchmarks

alganet:
With the introduction of composite actions in the GitHub workflows, the ci-perf.yml workflow broke.

This happens because the setup-action configures a Post Run (a hook to be executed after the workflow).

Since the benchmarks checkout a different orphan branch, when that Post Run executes, it cannot find the action anymore.

To fix it, I introduce a new step that restores the git workspace, making it available for that hook.

---

Example failure:

https://github.com/Respect/Validation/actions/runs/21448771339/job/61772437584

image

### DIFF
--- a/.github/workflows/ci-perf.yml
+++ b/.github/workflows/ci-perf.yml
@@ -77,3 +77,12 @@ jobs:
           git add .phpbench latest.md
           git commit -m "Store benchmark results [skip ci]" || echo "No changes to commit"
           git push origin benchmarks
+
+      - name: Restore workspace
+        if: always()
+        run: |
+          set -euo pipefail
+          echo "Restoring workspace to original commit: ${GITHUB_SHA}"
+          git fetch origin || true
+          git checkout --detach "${GITHUB_SHA}"
+          git reset --hard "${GITHUB_SHA}"


alganet:
@henriquemoody 

If we checkout only the `benchmarks` branch, there will be no `vendor/bin/phpbench` or benchmark classes to run.

I need to first checkout the codebase, run the benchmarks, then checkout the `benchmarks` and update it with the results.

Perhaps there are better ways of doing it, but I could not find them.

I know for sure this restore works because I did it in the PHL repository (it has a similar `coverage` branch also with composite actions): https://github.com/alganet/PHL/blob/main/.github/workflows/build.yml#L170

Perhaps we can do another checkout at the end, but I never tried that.

henriquemoody:
I suggested using `actions/checkout@v6` so you didn't have to run those Git commands in so many places, and just using `actions/checkout@v6` to handle that. You can use it into different paths with different branches, but it's not an issue, using the git commands is just fine.

henriquemoody (approved):

alganet:
I can confirm the fix worked as intended:

image

PR #1637, opened 2026-01-28
Remove now unecessary fallback to json_decode

alganet:
Since PHP 8.3, we don't need that fallback anymore. Now that the library is set to a minimum PHP 8.5, we can safely remove it.


alganet:
Thanks! Waiting for #1636 so the CI will run cleaner on main.

PR #1638, opened 2026-01-28
Containerize sokil databases

alganet:
The main focus of this change is to make those optional dependencies more testable.

Unfortunately, some phpstan-ignores had to be included, since ::set is not a PsrContainer method. We're only using it on tests though, so it's fine. It targets our php-di container for testing purposes only. The real implementation only relies on ::get.

This change also has the side effect of improving the performance of those validators by not instantiating their databases each time a iso validator is built, achieving massive improvements in those scenarios. A small benchmark with no assertions was added to track that improvement.

---

This brings the total coverage to almost 99%, addressing one of the key uncoverable paths that existed before.

### DIFF
--- a/src/Validators/CurrencyCode.php
+++ b/src/Validators/CurrencyCode.php
@@ -15,6 +15,8 @@ declare(strict_types=1);
 namespace Respect\Validation\Validators;
 
 use Attribute;
+use Psr\Container\NotFoundExceptionInterface;
+use Respect\Validation\ContainerRegistry;
 use Respect\Validation\Exceptions\InvalidValidatorException;
 use Respect\Validation\Exceptions\MissingComposerDependencyException;
 use Respect\Validation\Message\Template;
@@ -22,7 +24,6 @@ use Respect\Validation\Result;
 use Respect\Validation\Validator;
 use Sokil\IsoCodes\Database\Currencies;
 
-use function class_exists;
 use function in_array;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
@@ -39,14 +40,6 @@ final readonly class CurrencyCode implements Validator
         private string $set = 'alpha-3',
         Currencies|null $currencies = null,
     ) {
-        if (!class_exists(Currencies::class)) {
-            throw new MissingComposerDependencyException(
-                'CurrencyCode rule requires PHP ISO Codes',
-                'sokil/php-isocodes',
-                'sokil/php-isocodes-db-only',
-            );
-        }
-
         $availableSets = ['alpha-3', 'numeric'];
         if (!in_array($set, $availableSets, true)) {
             throw new InvalidValidatorException(
@@ -56,7 +49,15 @@ final readonly class CurrencyCode implements Validator
             );
         }
 
-        $this->currencies = $currencies ?? new Currencies();
+        try {
+            $this->currencies = $currencies ?? ContainerRegistry::getContainer()->get(Currencies::class);
+        } catch (NotFoundExceptionInterface) {
+            throw new MissingComposerDependencyException(
+                'CurrencyCode rule requires PHP ISO Codes',
+                'sokil/php-isocodes',
+                'sokil/php-isocodes-db-only',
+            );
+        }
     }
 
     public function evaluate(mixed $input): Result
--- a/src/Validators/SubdivisionCode.php
+++ b/src/Validators/SubdivisionCode.php
@@ -12,6 +12,8 @@ declare(strict_types=1);
 namespace Respect\Validation\Validators;
 
 use Attribute;
+use Psr\Container\NotFoundExceptionInterface;
+use Respect\Validation\ContainerRegistry;
 use Respect\Validation\Exceptions\InvalidValidatorException;
 use Respect\Validation\Exceptions\MissingComposerDependencyException;
 use Respect\Validation\Helpers\CanValidateUndefined;
@@ -20,8 +22,6 @@ use Respect\Validation\Result;
 use Respect\Validation\Validator;
 use Sokil\IsoCodes\Database\Countries;
 use Sokil\IsoCodes\Database\Subdivisions;
-
-use function class_exists;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
@@ -41,7 +41,11 @@ final readonly class SubdivisionCode implements Validator
         Countries|null $countries = null,
         Subdivisions|null $subdivisions = null,
     ) {
-        if (!class_exists(Countries::class) || !class_exists(Subdivisions::class)) {
+        try {
+            $container = ContainerRegistry::getContainer();
+            $countries ??= $container->get(Countries::class);
+            $this->subdivisions = $subdivisions ?? $container->get(Subdivisions::class);
+        } catch (NotFoundExceptionInterface) {
             throw new MissingComposerDependencyException(
                 'SubdivisionCode rule requires PHP ISO Codes',
                 'sokil/php-isocodes',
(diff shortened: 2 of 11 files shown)


src/Validators/CurrencyCode.php
```diff
@@ -56,7 +49,15 @@ public function __construct(
...
-        $this->currencies = $currencies ?? new Currencies();
+        try {
+            $this->currencies = $currencies ?? ContainerRegistry::getContainer()->get(Currencies::class);
+        } catch (NotFoundExceptionInterface) {
+            throw new MissingComposerDependencyException(
+                'CurrencyCode rule requires PHP ISO Codes',
+                'sokil/php-isocodes',
+                'sokil/php-isocodes-db-only',
+            );
+        }
```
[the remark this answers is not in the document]
alganet: Catching `MissingClassException` is a blunder, because only *our* container throws it.

The correct way to fix this is to change the message, indicating that sokil/php-isocodes must be installed *and registered in the container* (which is something we do by default, but anyone having a custom container would need to do on their own if they want to use those rules).

src/Validators/SubdivisionCode.php
```diff
@@ -41,22 +41,24 @@ public function __construct(
         Countries|null $countries = null,
         Subdivisions|null $subdivisions = null,
     ) {
-        if (!class_exists(Countries::class) || !class_exists(Subdivisions::class)) {
+        try {
+            $container = ContainerRegistry::getContainer();
+            $countries ??= $container->get(Countries::class);
+            $this->subdivisions = $subdivisions ?? $container->get(Subdivisions::class);
```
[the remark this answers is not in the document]
alganet: Good bot. I'll fix it.

src/ContainerRegistry.php
```diff
@@ -94,6 +100,10 @@ public static function createContainer(array $definitions = []): Container
                 ),
                 $container->get(TranslatorInterface::class),
             )),
+            Subdivisions::class => factory(static::optional(Subdivisions::class)),
+            Countries::class => factory(static::optional(Countries::class)),
+            Languages::class => factory(static::optional(Languages::class)),
+            Currencies::class => factory(static::optional(Currencies::class)),
```
henriquemoody: It would be best to autowire those dependencies instead, so maybe an easier way is not to add those definitions unless the class exists rather than creating the object if the class exists.

Those classes depend on `Sokil\IsoCodes\TranslationDriver\TranslatorInterface`, and it could be useful for uses if they also use `sokil/php-isocodes-db-i18n`.

We could talk to the maintainer of `sokil/php-isocodes` to use a `Symfony\Contracts\Translation\TranslatorInterface` directly. They made many changes to their packages as per my request -- they're open to it.

alganet: > It would be best to autowire those dependencies instead, so maybe an easier way is not to add those definitions unless the class exists rather than creating the object if the class exists.

Is this something that you see as being possible right now? Can you be more specific on how you would do it? (small pseudocode snippet).

> We could talk to the maintainer of sokil/php-isocodes to use a Symfony\Contracts\Translation\TranslatorInterface directly. They made many changes to their packages as per my request -- they're open to it.

That sounds great! May I suggest we open an issue to track that idea?

henriquemoody: >  Can you be more specific on how you would do it? (small pseudocode snippet).

Something _like_ this

```diff
     public static function createContainer(array $definitions = []): Container
     {
+        foreach ([Subdivisions::class, Countries::class, Languages::class, Currencies::class] as $class) {
+            if (class_exists($class)) {
+                $definitions[$class] = autowire($class);
+            }
+        }
+
         return new Container($definitions + [
```

> I suggest we open an issue to track that idea?

Here it is: https://github.com/sokil/php-isocodes/issues/86

alganet: @henriquemoody How would you approach the tests for that implementation?

henriquemoody: With the current setup, we cannot tell PHP that the class doesn't exist, so we can only test that the container created in `createdContainer()` can create build those objects.

alganet: I see, so, that would imply removing the newly added tests and going back to those paths being uncoverable.

Is that what you mean here, or am I missing something?

henriquemoody: No, you would keep your tests! What would change is how you define the container in the container registry.

I think just defining an empty container should be enough for testing each validator.

```php
ContainerRegistry::setContainer(new DI\Container());
```

src/Validators/SubdivisionCode.php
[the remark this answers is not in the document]
alganet: This is actually fine though. We are supposed to always have a container.

src/ContainerRegistry.php
alganet: I implemented your suggestion. However, as you can see in the [failed tests](https://github.com/Respect/Validation/actions/runs/21484209580/job/61888647038?pr=1638), the exception is not thrown when the test is setup that way.

What am I missing here?

henriquemoody: I checked it here, apparently PHP-DI has Autowiring enabled by default, but your can change the behaviour 


```php
$containerBuilder = new \DI\ContainerBuilder();
$containerBuilder->useAutowiring(false);

ContainerRegistry::setContainer($containerBuilder->build());
```

alganet: Nice, that did the trick! Thanks.

src/Exceptions/MissingClassException.php
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-FileCopyrightText: (c) Respect Project Contributors
+ */
+
+declare(strict_types=1);
+
+namespace Respect\Validation\Exceptions;
+
+use Psr\Container\NotFoundExceptionInterface;
+
+final class MissingClassException extends ComponentException implements Exception, NotFoundExceptionInterface
```
henriquemoody: I think you can get rid of this one

tests/unit/Validators/LanguageCodeTest.php
```diff
@@ -35,6 +41,31 @@ public function itShouldThrowAnExceptionWhenSetIsInvalid(): void
...
+    {
+        $mainContainer = ContainerRegistry::getContainer();
+        $container = ContainerRegistry::createContainer();
+        $container->set(Languages::class, factory(
+            static function (): void {
+                throw new MissingClassException();
+            },
+        ));
+        // @phpstan-ignore-next-line
+        ContainerRegistry::setContainer($container);
```
henriquemoody: I guess this one needs fixing

tests/unit/Validators/CurrencyCodeTest.php
```diff
@@ -36,6 +39,24 @@ public function itShouldThrowsExceptionWhenInvalidFormat(): void
...
+            new CurrencyCode('alpha-3');
+            $this->fail('Expected MissingComposerDependencyException was not thrown.');
+        } catch (MissingComposerDependencyException $e) {
+            $this->assertStringContainsString(
+                'CurrencyCode rule requires PHP ISO Codes',
+                $e->getMessage(),
+            );
+        } finally {
+            ContainerRegistry::setContainer($mainContainer);
+        }
```
henriquemoody, line 57: Nitpick: You might want to create a test helper for this one

src/ContainerRegistry.php
```diff
@@ -51,6 +56,21 @@ final class ContainerRegistry
...
+                Languages::class,
+                Currencies::class,
+            ] as $requiredClass
+        ) {
+            if (!class_exists($requiredClass)) {
+                continue;
+            }
+
+            $definitions[$requiredClass] = autowire($requiredClass);
+        }
```
henriquemoody: Since PHP-DI autowires by default, perhaps we can just get rid of this piece of code here.

PR #1640, opened 2026-01-29
Fix misplaced Outer/Inner names when using named() on Each

### LINKED ISSUE #1639: Each inverts names
In 3.0, the `Each` validator is inverting names.

Consider this example:

```php
test('Inner and outer names.', catchAll(
    fn() => v::named('my_file.txt', v::each(v::named('line', v::intType())))->assert(['a', 'b', 'c']),
    fn(string $message, string $fullMessage, array $messages) => expect()
        ->and($fullMessage)->toBe(<<<'FULL_MESSAGE'
        - Each item in my_file.txt must be valid
          - `.0` must be an integer
          - `.1` must be an integer
          - `.2` must be an integer
        FULL_MESSAGE)
));
```

In this scenario, `['a', 'b', 'c']` is the return of some `file('my_file.txt')`, a text file with 3 lines containing each of those letters.

Currently, it does not pass, and generates `Each item in line must be valid`.

This example is clearer than the ones in the codebase that use `Wrapper` and `Wrapped` and are easy to overlook. I'm using `my_file.txt` and `line` as a real example (each line of my_file.txt must be XXX, a common type of validation).

Ideally, if the validator passed to `Each` has a name, we want this:

```php
test('Inner and outer names.', catchAll(
    fn() => v::named('my_file.txt', v::each(v::named('line', v::intType())))->assert(['a', 'b', 'c']),
    fn(string $message, string $fullMessage, array $messages) => expect()
        ->and($fullMessage)->toBe(<<<'FULL_MESSAGE'
        - Each item in my_file.txt must be valid
          - `.0` (<- line) must be an integer
          - `.1` (<- line) must be an integer
          - `.2` (<- line) must be an integer
        FULL_MESSAGE)
));
```

We could omit `(<- line)` in the messages if the validator passed does not have an explicit name.

That would allow us to have a descriptive message of the whole (in this case, the file) and also the part (each line) with clear keys visible as well.

alganet:
The Each validator was using "Outer" names (the name applied to the parent) in child results and vice-versa.

This commit fixes it, and also streamlines the Result class, introducing helper `mapChildren` and `mapAdjacent` methods and removing unecessary recursive array_maps.

Named children now also display their names, allowing users to create more meaningful messages that do not spam nested `.0`...`.0`...etc numeric keys which could be confusing. If the user does not name them, the previous behavior is kept.

---

This fixes #1639

### DIFF
--- a/src/Message/Formatter/NestedArrayFormatter.php
+++ b/src/Message/Formatter/NestedArrayFormatter.php
@@ -45,6 +45,8 @@ final readonly class NestedArrayFormatter implements ArrayFormatter
                 $renderer,
                 $templates,
                 $hasString,
+                $result,
+                isParentVisible: count($children) > 1,
             ),
             [],
         );
@@ -91,12 +93,14 @@ final readonly class NestedArrayFormatter implements ArrayFormatter
         Renderer $renderer,
         array $templates,
         bool $hasString,
+        Result $parent,
+        bool $isParentVisible,
     ): array {
         if ($hasString && is_numeric($key)) {
             $key = $child->id->value;
         }
 
-        $message = $this->renderChild($child, $renderer, $templates);
+        $message = $this->renderChild($child, $renderer, $templates, $parent, $isParentVisible);
 
         if (!$hasString) {
             $messages[] = $message;
@@ -124,10 +128,15 @@ final readonly class NestedArrayFormatter implements ArrayFormatter
      *
      * @return array<string>|string
      */
-    private function renderChild(Result $child, Renderer $renderer, array $templates): array|string
-    {
+    private function renderChild(
+        Result $child,
+        Renderer $renderer,
+        array $templates,
+        Result $parent,
+        bool $isParentVisible,
+    ): array|string {
         $formatted = $this->format(
-            $child->withoutName(),
+            $isParentVisible && $child->name === $parent->name ? $child->withoutName() : $child,
             $renderer,
             $templates,
         );
--- a/src/Result.php
+++ b/src/Result.php
@@ -77,7 +77,7 @@ final readonly class Result
         if ($this->allowsAdjacent()) {
             return clone ($result, [
                 'id' => $this->id->withPrefix($prefix),
-                'adjacent' => $this->withInput($result->input),
+                'adjacent' => clone($this, ['input' => $result->input]),
             ]);
         }
 
@@ -112,12 +112,17 @@ final readonly class Result
         return clone($this, ['id' => Id::fromValidator($validator)]);
     }
 
+    public function withPrecedentName(bool $hasPrecedentName): self
+    {
+        return clone($this, [
+            'hasPrecedentName' => $hasPrecedentName,
+            'adjacent' => $this->adjacent?->withPrecedentName($hasPrecedentName),
+            'children' => $this->mapChildren(static fn(Result $r) => $r->withPrecedentName($hasPrecedentName)),
+        ]);
+    }
+
     public function withPath(Path $path): self
     {
-        if ($this->path === $path) {
-            return $this;
-        }
-
         if ($this->path !== null) {
             $this->path->parent = $path;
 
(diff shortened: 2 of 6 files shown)


src/Result.php
```diff
@@ -180,25 +176,7 @@ public function withNameFrom(Validator $validator): self
...
-            'hasPrecedentName' => true,
-            'adjacent' => $this->adjacent?->withNameFrom($validator),
-            'children' => array_map(
-                static fn(Result $child) => $child->withNameFrom($validator),
-                $this->children,
-            ),
-        ]);
-    }
-
-    public function withInput(mixed $input): self
```
henriquemoody, line 192: The `withInput()` was buggy, I was going to send a fix somewhere this week. This change just keeps the buggy behavior, though.

The `withInput()` should also overwrite the input of adjacent result (see [this](https://github.com/henriquemoody/Validation/commit/507b46603c846143adb274d591c5a441077166d4#diff-ef70a67aac7e66a56477071c98893e9698e5c2b8fabd1f92adb7890dd8cfe764R196)). I think it should overwrite the input of the children, but the conditions should not only take the input into account, but also if the path is the same or not.

It's okay to keep the code as is, but we should provide a `withInput()` in v3

src/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -45,6 +45,8 @@ public function format(Result $result, Renderer $renderer, array $templates): ar
                 $renderer,
                 $templates,
                 $hasString,
+                $result,
+                hasRoot: count($children) > 1,
```
henriquemoody: For a better understanding, it would be great if this boolean is saved in a variable that can also be used in line 54

src/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -91,12 +93,14 @@ private function appendMessage(
         Renderer $renderer,
         array $templates,
         bool $hasString,
+        Result $parent,
+        bool $hasRoot,
```
henriquemoody:
```suggestion
        bool $isParentVisible,
```
Nitpick, hasRoot is understandable in context

src/Result.php
```diff
@@ -247,4 +221,15 @@ public function allowsAdjacent(): bool
 
         return count($childrenThatAllowAdjacent) === 1;
     }
+
+    /** @return array<Result> */
+    private function mapChildren(callable $callback): array
```
henriquemoody, line 226: This is great! I was going to do that next time I had to do anything with the children again

src/Result.php
```diff
@@ -247,4 +221,15 @@ public function allowsAdjacent(): bool
...
+    {
+        return $this->children === [] ? [] : array_map($callback, $this->children);
+    }
+
+    private function mapAdjacent(callable $callback): Result|null
```
henriquemoody: I see less of a reason for this one, though. The nullsafe gives that for us for free, why creating this method?

henriquemoody (commented):
Awesome work, and great catch! I think we should get to an agreement regarding the `withInput()` before merging this one, though.

src/Result.php
alganet: @henriquemoody I didn't knew!

Can you open an issue for that one so we can track it?

src/Message/Formatter/NestedArrayFormatter.php
alganet: Line 54 is `count($messages)`, not `count($children)`. Those differ!

src/Result.php
alganet: It was doing more during a local refactor, but then it didn't, and I forgot to put this back how it was. Thanks!

src/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -45,6 +45,8 @@ public function format(Result $result, Renderer $renderer, array $templates): ar
                 $renderer,
                 $templates,
                 $hasString,
+                $result,
+                hasParentVisible: count($children) > 1,
```
henriquemoody: I think it should be either `isParentVisible` or `hasVisibleParent`

src/Result.php
henriquemoody: I opened #1641 instead

src/Message/Formatter/NestedArrayFormatter.php
henriquemoody: You're absolutely right, there's an `array_reduce()`!

henriquemoody (approved):

PR #1641, opened 2026-01-29
Fix input overwrite not propagating to adjacent results

henriquemoody:
Adjacent results are results that treat the same input. When overwriting the input of a result, we should also overwrite the input of its adjacent result to maintain consistency. Currently, there are no cases where this has caused issues, but this change prevents potential problems.

### DIFF
--- a/src/Result.php
+++ b/src/Result.php
@@ -191,12 +191,17 @@ final readonly class Result
 
     public function withInput(mixed $input): self
     {
-        $currentInput = $this->input;
-
         return clone($this, [
             'input' => $input,
+            'adjacent' => $this->adjacent?->withInput($input),
             'children' => array_map(
-                static fn(Result $child) => $child->input === $currentInput ? $child->withInput($input) : $child,
+                function (Result $child) use ($input): Result {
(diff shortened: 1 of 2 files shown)


henriquemoody:
You're right, I will try working on that. In any case, go merge #1640 first, I will handle the conflicts before merging this one.

alganet:
@henriquemoody I merged it. BTW you had this one open and approved before mine. I like the merge etiquette 😄 , but this change was small (not lots of conflicts to resolve), I could have handled it.

henriquemoody:
I think it's best to close this one and ship this code with the Masked rule.

I think it makes sense to ship the code with some actually use-case, otherwise I might ship something that does not meet our needs. Let me know if you disagree, @alganet.

alganet:
@henriquemoody I kind of agree. I was trying to refactor `Contains` to use `withInput` somehow (I had an idea, but it didn't panned out).

Maybe you can think of a smaller validator that we could add that exercises the paths we're aiming to fix. I couldn't think of any (that were easy to implement).

henriquemoody:
I will ship this method with the `Masked()` validator.

PR #1645, opened 2026-01-29
Refactor case sensitiveness support

alganet:
This is a mid-size refactor that affects several validators.

Most prominently, the ones that had an `$identical` parameter to deal with case sensitiveness.

This parameter was confusing, effectively making validators such as `Contains` behave very differently for arrays versus strings.

In arrays, `$identical` meant "the same type", while it in strings it meant "case sensitive".

That parameter was removed, and the default behavior is now to always compare **case sensitive** and strict typing.

A document explaining how to combine other validators in order to achieve case _insensitive_ comparisons was added.

Additionally, the `Call` validator was refactored back to be suitable to take on the task of being a fast, quick composable validator.

With the introduction of `Circuit`, we can shift the responsibility of dealing with possible mismatches to the user. This kind of type handling is demonstrated in how I refactored `Tld` to account for the type mismatch without setting error handlers.

### DIFF
--- a/docs/validators/Call.md
+++ b/docs/validators/Call.md
@@ -52,22 +52,20 @@ v::call(
 // Validation passes successfully
 ```
 
-## Templates
+Call does not handle possible errors (type mismatches). If you need to
+ensure that your callback is of a certain type, use [Circuit](Circuit.md) or 
+handle it using a closure:
 
-### `Call::TEMPLATE_STANDARD`
+```php
+v::call('strtolower', v::equals('ABC'))->assert(123);
+// 𝙭 strtolower(): Argument #1 ($string) must be of type string, int given
 
-|       Mode | Template                                                   |
-| ---------: | :--------------------------------------------------------- |
-|  `default` | {{input}} must be a suitable argument for {{callable}}     |
-| `inverted` | {{input}} must not be a suitable argument for {{callable}} |
+v::circuit(v::stringType(), v::call('strtolower', v::equals('abc')))->assert(123);
+// → 123 must be a string
 
-## Template placeholders
-
-| Placeholder | Description                                                      |
-| ----------- | ---------------------------------------------------------------- |
-| `callable`  |                                                                  |
-| `input`     |                                                                  |
-| `subject`   | The validated input or the custom validator name (if specified). |
+v::circuit(v::stringType(), v::call('strtolower', v::equals('abc')))->assert('ABC');
+// Validation passes successfully
+```
 
 ## Categorization
 
@@ -77,9 +75,10 @@ v::call(
 
 ## Changelog
 
-| Version | Description |
-| ------: | :---------- |
-|   0.3.9 | Created     |
+| Version | Description                   |
+| ------: | :---------------------------- |
+|   3.0.0 | No longer sets error handlers |
+|   0.3.9 | Created                       |
 
 ## See Also
 
--- a/src/Validators/Call.php
+++ b/src/Validators/Call.php
@@ -17,21 +17,12 @@ declare(strict_types=1);
 namespace Respect\Validation\Validators;
 
 use Attribute;
-use ErrorException;
-use Respect\Validation\Message\Template;
 use Respect\Validation\Result;
 use Respect\Validation\Validator;
-use Throwable;
 
 use function call_user_func;
-use function restore_error_handler;
-use function set_error_handler;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
-#[Template(
-    '{{input}} must be a suitable argument for {{callable}}',
-    '{{input}} must not be a suitable argument for {{callable}}',
-)]
 final class Call implements Validator
 {
     /** @var callable */
@@ -46,18 +37,6 @@ final class Call implements Validator
 
     public function evaluate(mixed $input): Result
     {
-        set_error_handler(static function (int $severity, string $message, string|null $filename, int $line): void {
-            throw new ErrorException($message, 0, $severity, $filename, $line);
-        });
-
-        try {
-            $result = $this->validator->evaluate(call_user_func($this->callable, $input));
-        } catch (Throwable) {
-            $result = Result::failed($input, $this, ['callable' => $this->callable]);
-        }
-
-        restore_error_handler();
-
-        return $result;
+        return $this->validator->evaluate(call_user_func($this->callable, $input));
     }
 }
--- a/src/Validators/ContainsAny.php
+++ b/src/Validators/ContainsAny.php
@@ -28,13 +28,13 @@ use function count;
 final class ContainsAny extends Envelope
 {
     /** @param non-empty-array<mixed> $needles */
-    public function __construct(array $needles, bool $identical = false)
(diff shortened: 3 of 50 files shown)


docs/validators/Call.md
```diff
@@ -52,14 +52,29 @@ v::call(
 // Validation passes successfully
 ```
 
+Call does not handle possible errors (type mismatches). If you need to
+ensure that your callback is of a certain type, use `Circuit` or handle it
+using a closure:
+
+```php
+v::call('strtolower', v::equals('ABC'))->assert(123);
+// 𝙭 strtolower(): Argument #1 ($string) must be of type string, int given
```
henriquemoody, line 61: Accidental feature from the linter 😂

src/Validators/Call.php
henriquemoody, line 31:
```suggestion
    /** @param callable(mixed): mixed $callable */
    public function __construct(
```

I think it might be good to do that, I'm not sure if PHPStan will start to complain, though.

src/Validators/ContainsAny.php
```diff
@@ -47,11 +47,11 @@ public function __construct(array $needles, bool $identical = false)
...
      */
-    private function getValidators(array $needles, bool $identical): array
+    private function getValidators(array $needles): array
     {
         return array_map(
-            static function ($needle) use ($identical): Contains {
-                return new Contains($needle, $identical);
+            static function ($needle): Contains {
+                return new Contains($needle);
             },
```
henriquemoody:
```suggestion
            static fn ($needle) => new Contains($needle),
```
Nitpick

src/Validators/Tld.php
```diff
@@ -23,6 +23,9 @@ final class Tld extends Envelope
 {
     public function __construct()
     {
-        parent::__construct(new Call('mb_strtoupper', new In(DataLoader::load('domain/tld.php'))));
+        parent::__construct(new Circuit(
+            new StringType(),
+            new Call('mb_strtoupper', new In(DataLoader::load('domain/tld.php'))),
```
henriquemoody, line 28:
```suggestion
            new Lowercase(),
            new Call(mb_strtoupper(...), new In(DataLoader::load('domain/tld.php'))),,
```
Do we also want to make this one case-insensitive?

tests/unit/Validators/CallTest.php
```diff
@@ -28,16 +26,15 @@ final class CallTest extends RuleTestCase
     public static function providerForValidInput(): iterable
     {
         return [
-            'valid rule and valid callable' => [new Call('trim', Stub::pass(1)), ' input '],
+            'callback true' => [new Call('strtolower', new Equals('abc')), 'ABC'],
```
henriquemoody, line 31: We're missing a test to how the validator behaves when the callback fail, it's a good one to have to establish that this is the expected behavior.

docs/validators/Call.md
```diff
@@ -52,14 +52,29 @@ v::call(
 // Validation passes successfully
 ```
 
+Call does not handle possible errors (type mismatches). If you need to
+ensure that your callback is of a certain type, use `Circuit` or handle it
```
henriquemoody:
```suggestion
ensure that your callback is of a certain type, use [Circuit](Circuit.md) or handle it
```

src/Validators/Call.php
```diff
@@ -17,20 +17,16 @@
...
 use Respect\Validation\Result;
 use Respect\Validation\Validator;
-use Throwable;
 
 use function call_user_func;
-use function restore_error_handler;
-use function set_error_handler;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
 #[Template(
```
henriquemoody: You can just remove this template, we'll never have one now.

henriquemoody (commented):
I love those changes!

Could you also update the migration guide? It's important information to be there.

src/Validators/Call.php
alganet: `phpcbf` just removes the whole comment. I guess it's truly redundant.

image

src/Validators/Tld.php
alganet: Wait, it's already case-insensitive.

When we generate `domain/tld.php`, we uppercase the domains. Then, before calling the input, we uppercase it as well. That makes it insensitive to case, which is what we want (`EXAMPLE.COM == example.com`).

tests/unit/Validators/CallTest.php
alganet: I added a test.

The user is responsible for handling any potential errors. For example, if he wants a type-safe `strtolower`, he can do so by doing something like this:

```php
v::call(static fn ($i) => strtolower((string) $i), v::equals('foo'))
```

Or, if we wants messages and not conversion behavior, using `Circuit` as documented.

src/Validators/Tld.php
henriquemoody: With the current implementation `EXAMPLE.COM`, or `EXAMPLE.com` all pass as valid because we force them to be uppercase. You're know more about this stuff than I do, but I think, as input: 

* EXAMPLE.COM: is invalid
* EXAMPLE.com: is invalid
* example.com: valid

Am I missing something?

docs/validators/Call.md
```diff
@@ -52,14 +52,29 @@ v::call(
...
+
 ## Templates
 
 ### `Call::TEMPLATE_STANDARD`
 
-|       Mode | Template                                                   |
-| ---------: | :--------------------------------------------------------- |
-|  `default` | {{input}} must be a suitable argument for {{callable}}     |
-| `inverted` | {{input}} must not be a suitable argument for {{callable}} |
+|       Mode | Template                    |
```
henriquemoody: That's odd, the linter should have not allowed this to be here. Could you remove it?

src/Validators/Tld.php
henriquemoody: I typed it wrong. My question was if we also want to remove the case-insensitivity in this validador.

tests/unit/Validators/CallTest.php
henriquemoody: Yes, I got that, I think it's the best way forward, not to hide what could go wrong.

src/Validators/Tld.php
alganet: Regarding correctedness:

https://www.rfc-editor.org/rfc/rfc3986.html Section 6.2.2.1 "Case Normalization"

> scheme and host are case-insensitive and therefore should be normalized to lowercase.  For example, the URI <HTTP://www.EXAMPLE.com/> is equivalent to <http://www.example.com/>. 

The top-level domain, being a part of the host, should therefore be compared case-insensitively.

---

If our data was public (people would consume it directly from the data/ folder), there's an argument for making the comparison happen in lowercase, and save our data in lowercase too.

However, since it's not public and an implementation detail, I think the uppercase convention is better

Uppercase letters appear first in the ASCII table and were historically "faster" in earlier times. The performance benefits are not true anymore, but the tradition continues: if you find data in UPPERCASE somewhere, it usually means a case-insensitive comparison is being done. I think that communicates intent well.

henriquemoody (approved):

PR #1649, opened 2026-01-30
Allow empty values in iterables for All, Each, Max, Min

alganet:
Now empty values are again allowed in FilteredArray-style validators.

To solve the issue with negation, a Result attribute was added to signal indeciseveness (when a result cannot be reliably inverted). On such cases, we consider that result to be valid.

For example, `v::not(v::min(v::equals(10)))` says "The lowest value of the iterable input should not be equal 10".

If the input is empty, we cannot decide whether its minimum is equal to 10 or not, so the validator essentially becomes a null-op.

Users that want to ensure these validators have a valid decidable target must use it in combination with `Length` or other similar validators to achieve the same result.

### DIFF
--- a/docs/validators/Each.md
+++ b/docs/validators/Each.md
@@ -20,6 +20,22 @@ v::each(v::dateTime())->assert($releaseDates);
 // Validation passes successfully
 ```
 
+This validator is similar to [All](All.md), but while `All` displays a single message 
+generic to all failed entries, `Each` will display a message for each failed 
+entry instead.
+
+```php
+v::each(v::startsWith('2010'))->assert($releaseDates);
+// → - Each item in `["validation": "2010-01-01", "template": "2011-01-01", "relational": "2011-02-05"]` must be valid
+// →   - `.template` must start with "2010"
+// →   - `.relational` must start with "2010"
+
+v::named('Release Dates', v::each(v::startsWith('2010')))->assert($releaseDates);
+// → - Each item in Release Dates must be valid
+// →   - `.template` must start with "2010"
+// →   - `.relational` must start with "2010"
+```
+
 You can also validate array keys combining this validator with [Call](Call.md):
 
 ```php
@@ -29,7 +45,15 @@ v::call('array_keys', v::each(v::stringType()))->assert($releaseDates);
 
 ## Note
 
-This validator uses [Length](Length.md) with [GreaterThan][GreaterThan.md] internally. If an input has no items, the validation will fail.
+This validator will pass if the input is empty. Use [Length](Length.md) with [GreaterThan][GreaterThan.md] to perform a stricter check:
+
+```php
+v::each(v::equals(10))->assert([]);
+// Validation passes successfully
+
+v::length(v::greaterThan(0))->each(v::equals(10))->assert([]);
+// → The length of `[]` must be greater than 0
+```
 
 ## Templates
 
@@ -54,11 +78,11 @@ This validator uses [Length](Length.md) with [GreaterThan][GreaterThan.md] inter
 
 ## Changelog
 
-| Version | Description                                                 |
-| ------: | :---------------------------------------------------------- |
-|   3.0.0 | Rejected `stdClass`, non-iterable. or empty iterable values |
-|   2.0.0 | Remove support for key validation                           |
-|   0.3.9 | Created                                                     |
+| Version | Description                        |
+| ------: | :--------------------------------- |
+|   3.0.0 | Rejected `stdClass`, non-iterables |
+|   2.0.0 | Remove support for key validation  |
+|   0.3.9 | Created                            |
 
 ## See Also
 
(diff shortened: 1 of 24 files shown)


src/Validators/Core/FilteredArray.php
```diff
@@ -32,28 +28,23 @@ public function evaluate(mixed $input): Result
...
         $array = $this->toArray($input);
-        $validator = new Circuit(
-            new Not(new Undef()),
-            new Length(new GreaterThan(0)),
-        );
-        $result = $validator->evaluate($array);
-        if (!$result->hasPassed) {
-            return $result->withIdFrom($this);
+
+        if (!count($array)) {
```
henriquemoody:
```suggestion
        if ($array === []) {
```

No function call, I think it's better

src/Validators/Core/FilteredArray.php
```diff
@@ -32,28 +28,23 @@ public function evaluate(mixed $input): Result
...
+    /** @param array<mixed> $input */
+    abstract protected function evaluateArray(array $input): Result;
 
     /**
      * @param iterable<mixed> $input
      *
      * @return array<mixed>
      */
-    private function toArray(iterable $input): array
+    protected function toArray(iterable $input): array
```
henriquemoody: Why that?

alganet:
Known draft issues:

 - Terminology: I'm considering `isIndeterminate` instead of `wasSkipped`
 - Tests: I'm working towards 100% coverage
 - Docs and migration guide: on their way

src/Result.php
```diff
@@ -222,6 +223,11 @@ public function allowsAdjacent(): bool
         return count($childrenThatAllowAdjacent) === 1;
     }
 
+    public function withWasSkipped(bool $wasSkipped): self
```
henriquemoody:
```suggestion
    public function asSkipped(): self
```

Just wondering about names, but I can only see us skipping it, not the other way around. Maybe it's too early to restrict, it feels a _bit_ off. Not a blocker, though

src/Validators/Not.php
```diff
@@ -28,6 +28,10 @@ public function evaluate(mixed $input): Result
     {
         $result = $this->validator->evaluate($input);
 
+        if ($result->wasSkipped) {
+            return Result::passed($input, $this);
```
henriquemoody: Because of that, you will have to create a template for Not()

henriquemoody: What happens if you just return the result?

src/Validators/Core/FilteredArray.php
alganet: Leftover from refactor process. At some point, I still had FilteredNonEmptyArray extending from the newly introduced FilteredArray, so I needed the method visible there. I'll switch it back to private.

src/Validators/Not.php
alganet: Just returning the result works!

src/Result.php
alganet: I'm leaning towards the `indeterminate` terminology. Agree that this is only setting true, we would never set it to false during runtime (not for a foreseable future, unless we start doing some statistical validation).

src/Validators/Core/FilteredArray.php
```diff
@@ -32,28 +28,23 @@ public function evaluate(mixed $input): Result
         }
 
         $array = $this->toArray($input);
-        $validator = new Circuit(
```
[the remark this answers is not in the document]
alganet: We could also remove the `Iterable` check, treating scalars as an array with one item 😈 In that scenario, the user needs to ->arrayVal beforehand.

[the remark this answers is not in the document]
henriquemoody: That would only work if the user is running it in a Circuit, otherwise it will trigger PHP errors

src/Result.php
henriquemoody: indeterminate sounds good to me.

src/Validators/Core/FilteredArray.php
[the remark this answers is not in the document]
alganet: What I mean is: 

```
is_iterable($input) ? $input : [$input]
```

Literally treating non-iterables as an array with one item.

How would that lead to PHP errors?

docs/validators/Each.md
```diff
@@ -20,6 +20,23 @@ v::each(v::dateTime())->assert($releaseDates);
...
+// →   - `.template` must start with "2010"
+// →   - `.relational` must start with "2010"
+
+v::named('Release Dates', v::each(v::startsWith('2010')))->assert($releaseDates);
+// → - Each item in Release Dates must be valid
+// →   - `.template` must start with "2010"
+// →   - `.relational` must start with "2010"
+```
+
+
```
henriquemoody:
```suggestion

```

henriquemoody (approved):
I merged my PR before seeing this one was requested for review. You'll need to resolve conflicts. Sorry, dude

docs/validators/Each.md
alganet: Since there was no comment text, just a suggestion that I implemented, I'm chosing to resolve this.

PR #1650, opened 2026-01-30
Rename Core\Reducer to Composite and expose it as a public validator

henriquemoody:
The term "Reducer" is widely used in functional programming to describe a specific pattern: a function that takes an accumulator and a value, returning a new accumulator (e.g., array_reduce, Redux reducers). This class does not perform that operation. Instead, it consolidates multiple validators into a single one, which is the definition of composition.

The new name "Composite" accurately reflects what this validator does: it composes zero or more validators into a unified validation unit. This aligns with the Composite design pattern terminology and avoids semantic confusion with functional programming concepts.

The class has also been moved from the internal Core namespace to the public Validators namespace. This change acknowledges that composing validators dynamically is a legitimate use case for library consumers, not just an internal implementation detail. The library itself uses this pattern in Attributes and KeySet validators, demonstrating its practical value.

The signature has been extended to accept zero or more validators:

- Zero validators: Returns AlwaysValid (useful for conditional chains)
- One validator: Pass-through without wrapping (avoids unnecessary nesting)
- Multiple validators: Combines using AllOf (all must pass)

This flexibility is particularly useful when building validation chains dynamically, where the number of validators may vary at runtime based on configuration or user input.

Assisted-by: Claude Code (Claude Opus 4.5)


src/Validators/Composite.php
```diff
@@ -0,0 +1,30 @@
...
+namespace Respect\Validation\Validators;
+
+use Attribute;
+use Respect\Validation\Validator;
+use Respect\Validation\Validators\Core\Wrapper;
+
+use function count;
+
+#[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
+final class Composite extends Wrapper
```
alganet: With this change, we would have a `Validators\Composite` and a `Validators\Core\Composite`. Two distinct classes that are very similar.

They are actually related:

 - When you instantiate `Validators\Composite`, if there are two or more parameters, it will create an `AllOf`, which is a subtype of `Validators\Core\Composite`.

 - Both receive a list of validators.

However, the abstract Core Composite is actually *more strict* than the public one (it exists in order to make the variadic function accept two mandatory arguments).

So, as part of this refactor, I would suggest renaming `Core\Composite` to something else that explicitly tells that it needs two or more arguments.

This "check for two arguments" exists in two places. In the function signature of Core\Composite and in the match body of this new validator. Is there a way we can abstract that so that is expressed in a single place instead?

henriquemoody: That didn't come unnoticed, my mistake for not mentioning. I was planning to do that in another PR to explain it a bit better since there were no conflicts, just two classes with the same name in different namespaces. Would you be okay with another merge request or do you think it should be in the same one?

As for the name of the other one, I'm not sure how to call it, yet

alganet: If you could make the rename in this PR (it's a simple change), then the more complicated changes in other, that would be great! That's not a blocker though.

I also don't know about the names.

Here's some terminology I was researching:

 - Unary: a function that takes exactly one argument (our ::evaluate is unary)
 - Binary: a function that takes exactly two arguments.
 - N-ary or variadic: indefinite number of arguments.

This is call "arity": https://en.wikipedia.org/wiki/Arity

I was thinking of something like `MinBinaryComposite`, but it sounds very weird.

PR #1654, opened 2026-01-30
Create "Patterned" validator

henriquemoody:
The Patterned validator formats input values using a pattern before displaying them in error messages, while still validating the original unformatted input. This is the complement to the Masked validator:

- Masked: Hides sensitive parts of the input (e.g., "4111****1111")
- Patterned: Adds formatting to raw input (e.g., "4111 1111 1111 1111")

This validator solves a common UX problem: when users submit data without formatting characters (like spaces or punctuation), error messages show the raw input, which can be harder to read and verify. For example:

  - Credit card "4111111111111112" → "4111 1111 1111 1112"
  - CPF "12345678900" → "123.456.789-00"
  - Phone "11987654321" → "(11) 98765-4321"

The validator is particularly useful in scenarios where:

1. Form inputs strip formatting characters before submission
2. Data comes from external systems without formatting
3. Users need to visually verify the value they entered
4. Error messages should match the expected display format

It uses the PatternFormatter from respect/string-formatter, which supports digit placeholders (0), quantifiers ({n}), and transformations (\u for uppercase, \l for lowercase, etc.).

### DIFF
--- a/docs/migrating-from-v2-to-v3.md
+++ b/docs/migrating-from-v2-to-v3.md
@@ -581,6 +581,7 @@ Version 3.0 introduces several new validators:
 | `Lazy`             | Creates validators dynamically based on input              |
 | `Masked`           | Masks sensitive input values in error messages             |
 | `Named`            | Customizes the subject name in error messages              |
+| `Patterned`        | Formats input values using a pattern in error messages     |
 | `PropertyExists`   | Checks if an object property exists                        |
 | `PropertyOptional` | Validates an object property only if it exists             |
 | `Templated`        | Attaches custom error message templates                    |
@@ -716,6 +717,18 @@ v::masked('1-@', v::email(),v::email())->assert('invalid@example.com');
 
 v::masked('6-12', v::creditCard(), 'X')->assert('4111111111111211');
 // → "41111XXXXXXX1211" must be a valid credit card number
+```
+
+#### Patterned
+
+Decorates a validator to format input values using a pattern in error messages while still validating the original unformatted data. This is useful for displaying formatted values when the original input lacks formatting characters:
+
+```php
+v::patterned('0{3}.0{3}.0{3}-0{2}', v::cpf())->assert('12345678900');
+// → "123.456.789-00" must be a valid CPF number
+
+v::patterned('$00.00', v::phone())->assert(1297);
+// → "$12.97" must be a valid telephone number
 ```
 
 #### Named
--- a/tests/feature/Validators/PatternedTest.php
+++ b/tests/feature/Validators/PatternedTest.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * SPDX-License-Identifier: MIT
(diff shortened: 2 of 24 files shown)


docs/migrating-from-v2-to-v3.md
```diff
@@ -714,6 +715,18 @@ v::masked('6-12', v::creditCard(), 'X')->assert('4111111111111211');
...
+#### Patterned
+
+Decorates a validator to format input values using a pattern in error messages while still validating the original unformatted data. This is useful for displaying formatted values when the original input lacks formatting characters:
+
+```php
+v::patterned('0{3}.0{3}.0{3}-0{2}', v::cpf())->assert('12345678900');
+// → "123.456.789-00" must be a valid CPF number
+
+v::patterned('(0{2}) 0{5}-0{4}', v::phone())->assert('11987654321');
+// → "(11) 98765-4321" must be a valid telephone number
```
alganet: I was wondering if we could also document some common number patterns:

Thousands separation: `1000000` -> `1.000.000`
Comma separation: `10.00` => `10,00`

I think those are by far the most popular use cases for adding patterns to some display value.

tests/feature/Validators/PatternedTest.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+/*
+ * SPDX-License-Identifier: MIT
+ * SPDX-FileCopyrightText: (c) Respect Project Contributors
+ * SPDX-FileContributor: Henrique Moody <henriquemoody@gmail.com>
+ */
+
+declare(strict_types=1);
```
alganet, line 9: Can we add at least one test that showcases the behavior when using this together with `named()` and `templated()`?

This is also true for the `Masked` #1652 (already closed, I have not noticed back then) and the `Formatted` #1657 PRs.

docs/migrating-from-v2-to-v3.md
```diff
@@ -718,6 +719,18 @@ v::masked('6-12', v::creditCard(), 'X')->assert('4111111111111211');
...
+#### Patterned
+
+Decorates a validator to format input values using a pattern in error messages while still validating the original unformatted data. This is useful for displaying formatted values when the original input lacks formatting characters:
+
+```php
+v::patterned('0{3}.0{3}.0{3}-0{2}', v::cpf())->assert('12345678900');
+// → "123.456.789-00" must be a valid CPF number
+
+v::patterned('$00.00', v::phone())->assert(1297);
+// → "$12.97" must be a valid telephone number
```
alganet, line 731: This example is misleading. The use of `$` implies a monterary value, but the validator used is a phone number.

Suggestion: Keep the phone number, and instead add examples for the most popular use cases:

 - Thousands separator
 - Decimals separator
 - Dates

henriquemoody:
Noted the comments, but I will close this one in favor of #1657 with some aliases.

I will implemente your comments there.

PR #1656, opened 2026-01-30
Rename "Lazy" validator to "Dynamic"

henriquemoody:
The name "Dynamic" better conveys the validator's purpose: dynamically creating a validator at runtime based on the input value. "Lazy" was misleading as it suggested simple deferred evaluation rather than input-dependent validator construction.

Also renamed the constructor argument from `$validatorCreator` to `$factory` for improved expressiveness.


alganet:
Can we make the renaming of `Call`, `Callback` and `Lazy` in a single PR? Those are all similar (in the sense that they open doors to things Validation cannot do on its own) and I think it makes sense to consider them as a group and not individually.

henriquemoody:
Will do!

PR #1657, opened 2026-01-31
Create "Formatted" validator

henriquemoody:
The Formatted validator decorates another validator to transform how input values appear in error messages, while still validating the original unmodified input.

This is useful for improving the readability of error messages by displaying values in a user-friendly formatd.

The validator accepts any Respect\StringFormatter\Formatter implementation, allowing direct use of StringFormatter's fluent builder. As StringFormatter expands with more formatters in future releases, users will automatically benefit from the full range of formatting options.


henriquemoody:
I'm going to discard this one. We can achieve the same if FormatterBuilder from string Formatter has a __invoke()

henriquemoody:
Not true, `Call()` modifies before validation

src/Validators/Formatted.php
```diff
@@ -0,0 +1,42 @@
...
+
+/*
+ * SPDX-License-Identifier: MIT
+ * SPDX-FileCopyrightText: (c) Respect Project Contributors
+ * SPDX-FileContributor: Alexandre Gomes Gaigalas <alganet@gmail.com>
+ * SPDX-FileContributor: Danilo Benevides <danilobenevides01@gmail.com>
+ * SPDX-FileContributor: Graham Campbell <graham@mineuk.com>
+ * SPDX-FileContributor: Henrique Moody <henriquemoody@gmail.com>
+ * SPDX-FileContributor: Jayson Reis <santosdosreis@gmail.com>
+ * SPDX-FileContributor: Nick Lombard <github@jigsoft.co.za>
```
alganet: I see GitHub marking this file as newly added. Is this a moved file from somewhere else that GitHub didn't caught?

If it's a new file, perhaps we should review its FileContributor entries.

src/Transformers/FormattedPrefixTransformer.php
```diff
@@ -0,0 +1,90 @@
...
+use Respect\StringFormatter\FormatterBuilder;
+
+use function array_key_exists;
+use function array_slice;
+use function lcfirst;
+use function str_starts_with;
+use function strlen;
+use function substr;
+
+final readonly class FormattedPrefixTransformer implements Transformer
```
alganet: It is confusing to have both a generic `Prefix` transformer and a `FormattedPrefix` transformer.

It seems that this job could have been handled by the existing `Prefix` one just fine, and leveraged the unified pattern matching performance gains.

If the the issue is avoiding a god object that does everything (a valid concern!), we could explore how to break `Prefix` itself (that does double duty with `RULES_THAT_USE_SUFFIX_AS_ARGUMENT` and `RULES_THAT_PREFIX_OR_STAND_ALONE`). 

There are paths we could take that preserve the generation of a single fast matching expression and also allow for modularity (having a distinct class for each type of prefix). I recommend [this blog post](https://www.npopov.com/2014/02/18/Fast-request-routing-using-regular-expressions.html) about the FastRoute project that conveys the idea of combined regular expressions (one of the solutions I envision for this abstraction).

docs/migrating-from-v2-to-v3.md
```diff
@@ -659,6 +660,34 @@ v::dateTimeDiff('years', v::greaterThanOrEqual(18))->assert('2000-01-01'); // pa
...
+// → "****an email" must be an email address
+
+v::formatted(f::pattern('#### #### #### ####'), v::creditCard())->assert('1234123412341234');
+// → "1234 1234 1234 1234" must be a credit card number
+```
+
+Shorthand aliases are available that combine formatter creation with the validator:
+
+```php
+v::formattedAsMask('1-4', v::email())->assert('not an email');
```
alganet: This is virtually identical to the existing `v::masked()` but with another name.

From my perspective, `Masked` should be removed in this PR.

src/Transformers/FormattedPrefixTransformer.php
```diff
@@ -0,0 +1,88 @@
...
+{
+    /**
+     * Maps formatter names to the position of the Validator argument.
+     */
```
alganet: Why change the position of the argument?

I think it's much more reasonable to always use the `$validator` as the first argument. It estabilishes a consistent style for the prefix. With that, the user knows that all `formattedAs...` calls are always constructed the same way, no need to guess or consult documentation.

I understand some validators "read" better as a phrase, but I would prefer not to sacrifice consistency for that. It's the same needle/haystack ordering dillema, right? Sometimes it "feels" better to have needle first, sometimes haystack first, but the best feeling by far is having it be consistent.

This also simplifies the implementation with no need for complicated slicing.

docs/migrating-from-v2-to-v3.md
henriquemoody: I was wondering if I would remove it in a next one or in this one, but you're right, it makes sense to remove on this one

src/Transformers/FormattedPrefixTransformer.php
henriquemoody: You've got a point, but not only it reads better, but it's consistent with other validators like `named()` and `templated()`, that's why I made that decision, but I'm fully convinced of that either. I'm still making up my mind. Do you think we should also change `named()` and `templated()`?

src/Transformers/FormattedPrefixTransformer.php
henriquemoody: I will read the post later. I made that into a different transformer because it seemed specific enough to deserve it's own place. If we go for the validator in the first argument, I have more reasons to keep it in `PrefixTransformmer`

henriquemoody: Side tracking a little bit: The `Transformer` made more sense when we had several different ones (especially to keep backward compatibility), but now that we only have 1 I think we might not need it.

src/Transformers/FormattedPrefixTransformer.php
alganet: I think they're different. The `v::formattedAsXXXX` is like a door for `StringFormatter`. It translates to `f::XXX(args)`.

 - `v::named` - Leaf, it ends the dynamic lookup space there.
 - `v::templated` - Leaf, it ends the dynamic lookup space there.
 - `v::formatted` - Leaf, it ends the dynamic lookup space there (`f::` takes over explicitly)
 - `v::formatedAsXXXX` - Branch. It forwards a dynamic lookup and passes args down.

However, I would not mind having them use the first argument as the validator if that makes you sleep better!

As I mentioned earlier in private convos, I don't like having so many items in this category (v::named, v::templated, etc). Personally, I would pursue `v::as()` or `v::display` as a root branch for all things that alter result messages. 

```php
v::display(validator: v::something(), name: "Foo");
v::display(validator: v::something(), format: f::mask());
```

Especially when we want to combine them.

Today:

```php
v::templated(
    '{{subject must be fooish',
    v::formattedAsUppercase(
        v::named('Foo', v::something())
    )
);
```

What I think would work better:

```php
v::display(
    validator: v::something(),
    name: "Foo", 
    template: "{{subject}} must be fooish", 
    format: f::uppercase(), 
);
```

The name doesn't matter that much, it's the idea that there's a single thing you need to remember to change how results appear in the messages, and it has all the options available there. No need to guess "what if I named the formatted validator instead of the inner one?" or things like that which sprout from the nesting design.

henriquemoody: I'm more and more inclined to go for something like you're suggesting, actually. I like `as` better than `display`, but we'd have an issue because `as` is a reserved word.

After having this conversation, I'm inclined to remove the prefix `formatted`. StringFormatter is a package that we manage ourselves, but I think it's best not to have such a hard dependency on the name.

PR #1661, opened 2026-01-31
Improve Documentation

alganet:
- Added `composer docs-serve` to preview documentation locally.
 - Improved `ValidatorRelatedLinter` to not include invalid related entries.
 - Updated `CONTRIBUTING.md`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,17 +8,8 @@ SPDX-License-Identifier: MIT
 Contributions to Respect\Validation are always welcome. You make our lives
 easier by sending us your contributions through [pull requests][].
 
-Pull requests for bug fixes must be based on the oldest stable version's branch
-whereas pull requests for new features must be based on the `master` branch.
-
 Due to time constraints, we are not always able to respond as quickly as we
-would like. Please do not take delays personal and feel free to remind us here,
-on IRC, or on Gitter if you feel that we forgot to respond.
-
-Please see the [project documentation][] before proceeding. You should also know
-about [PHP-FIG][]'s standards and basic unit testing, but we're sure you can
-learn that just by looking at other validators. Pick the simple ones like `ArrayType`
-to begin.
+would like. Please do not take delays personally.
 
 Before writing anything, feature or bug fix:
 
@@ -28,7 +19,7 @@ Before writing anything, feature or bug fix:
     to work on that;
   - If there is, create a comment to notify everybody that you're going to
     work on that.
-- Make sure that what you need is not done yet
+- Make sure that what you need is not done yet.
 
 ## Adding a new validator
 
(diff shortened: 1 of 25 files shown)


CONTRIBUTING.md
```diff
@@ -40,10 +31,8 @@ A common validator on Respect\Validation is composed of three classes:
 The classes are pretty straightforward. In the sample below, we're going to
 create a validator that validates if a string is equal to "Hello World".
 
-- Classes should be `final` unless they are used in a different scope;
-- Properties should be `private` unless they are used in a different scope;
-- Classes should use strict typing;
-- Some docblocks are required.
+- Validator classes should be `final` whenever possible.
+- Properties should be `private` and `readonly` whenever possible.
```
henriquemoody, line 36:
```suggestion
- Validator classes should `readonly` whenever possible.
- Properties should be `private` and `readonly` whenever possible.
```

CONTRIBUTING.md
```diff
@@ -90,10 +79,8 @@ Finally, we need to test if everything is running smooth. We have `RuleTestCase`
 that allows us to make easier to test validators, but you fell free to use the
 `PHPUnit\Framework\TestCase` if you want or you need it's necessary.
 
-The `RuleTestCase` extends PHPUnit's `PHPUnit\Framework\TestCase` class, so you
-are able to use any methods of it. By extending `RuleTestCase` you should
-implement two methods that should return a [data provider][] with the validator as
-first item of the arrays:
+The `RuleTestCase` extends PHPUnit's `PHPUnit\Framework\TestCase` class with
```
henriquemoody, line 83: I'm actually fed up with the `RuleTestCase`, because it forces us to add the Validator to the data provider, we don't have code coverage for those. Apart from that, tracking the errors with the stack trace is challenging, and it's not very friendly with the IDEs, in my experience. I think we should recommend the `TestCase` that we have, instead of the `RuleTestCase` -- I want to get rid of it.

CONTRIBUTING.md
```diff
@@ -142,60 +129,49 @@ for it other than what is covered by `RuleTestCase`.
...
-```sh
-> composer phpunit
-```
+Typical workflow:
 
-No test should fail.
+ - (Optional) Write a new benchmark in `tests/benchmark`.
+ - Run `vendor/bin/phpbench run --tag=before` before making a change.
+ - Change the code as you like.
+ - Run `vendor/bin/phpbench run --ref=before` to compare performance with the
```
henriquemoody, line 170:
```suggestion
 - Run `vendor/bin/phpbench run --tag=after --ref=before` to compare performance with the
```

CONTRIBUTING.md
alganet: Makes sense! Can you open an issue for that?

CONTRIBUTING.md
alganet: We don't need to tag each run. 

There are valid usage patterns for `phpbench run --tag` and `phpbench log` for more complicate setups (comparing 3 implementations for competing tradeoffs). However, that's very advanced usage that the phpbench manual can help explain better than us.

CONTRIBUTING.md
alganet: I resolved it because I implemented your exact suggestion.

PR #1662, opened 2026-01-31
Improve Url validator to be more strict

alganet:
Previously, the URL validator accepted all kinds of URLs.

Combining it with other validators like `Domain` and `Ip` manually is clumbersome, since the rules are extensive and require small tweaks such as understanding bracketed IP addresses.

This change makes that composition built-in. The validator still uses the FILTER_VALIDATE_URL from PHP, but now it also goes deeper and validate portions of the URL that other validators support.

Changes to `Call` were made so that it can be serialized under certain circumstances (when invoking a static method call), making it more flexible.

A static internal method helper for trimming the IP was added to the Url validator class and marked as internal, as it cannot be both private, serializable and accessible to the `Call` instance all at the same time. The `@internal` annotation should advise users that it's not a public API according to phpdoc conventions.

### DIFF
--- a/src/Validators/Call.php
+++ b/src/Validators/Call.php
@@ -20,8 +20,6 @@ use Attribute;
 use Respect\Validation\Result;
 use Respect\Validation\Validator;
 
-use function call_user_func;
-
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
 final class Call implements Validator
 {
@@ -37,6 +35,6 @@ final class Call implements Validator
 
     public function evaluate(mixed $input): Result
     {
-        return $this->validator->evaluate(call_user_func($this->callable, $input));
+        return $this->validator->evaluate(($this->callable)($input));
     }
 }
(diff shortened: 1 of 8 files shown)


src/Validators/Call.php
```diff
@@ -20,23 +20,22 @@
...
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
 final class Call implements Validator
 {
     /** @var callable */
     private $callable;
 
+    /** @param callable $callable */
     public function __construct(
-        callable $callable,
+        callable|string|array $callable,
```
henriquemoody: This will result on unwanted behavior. Right now this works, you can try it on main:

``
v::call('Datetime::createFromTimestamp', v::instance(\DateTime::class))->assert(1769981039);
v::call([new DateTime(), 'format'], v::equals('2026-02-01'))->assert('Y-m-d');
``

With your changes, this validator will accept any string and any array, even if they're not a callable.

alganet: You're right, I don't need that type signature change.

henriquemoody (approved):

PR #1663, opened 2026-02-01
Create ShortCircuit validator and ShortCircuitable interface

henriquemoody:
This commit introduces a mechanism for validators to return early once the validation outcome is determined, rather than evaluating all child validators.

The ShortCircuit validator evaluates validators sequentially and stops at the first failure, similar to how PHP's && operator works. This is useful when later validators depend on earlier ones passing, or when you want only the first error message.

The ShortCircuitCapable interface allows composite validators (AllOf, AnyOf, OneOf, NoneOf, Each, All) to implement their own short-circuit logic:

- AllOf: stops at first failure (like &&)
- AnyOf: stops at first success (like ||)
- OneOf: stops when two validators pass (already invalid)
- NoneOf: stops at first success (already invalid)
- Each/All: stops at first failing item

Why "ShortCircuit" instead of "FailFast":

The name "FailFast" was initially considered but proved misleading. While AllOf stops on failure (fail fast), AnyOf stops on success (succeed fast), and OneOf stops on the second success. The common behavior is not about failing quickly, but about returning as soon as the outcome is determined—which is exactly what short-circuit evaluation means. This terminology is familiar to developers from boolean operators (&& and ||), making the behavior immediately understandable.

***
### Before
Screenshot 2026-02-02 at 00 29 41

### After
Screenshot 2026-02-02 at 00 29 09


src/ValidatorBuilder.php
```diff
@@ -73,29 +74,19 @@ public function validate(mixed $input, array|string|null $template = null): Resu
 
     public function isValid(mixed $input): bool
     {
-        return $this->evaluate($input)->hasPassed;
+        return (new ShortCircuit(...$this->validators))->evaluate($input)->hasPassed;
     }
 
     /** @param array<string|int, mixed>|callable(ValidationException): Throwable|string|Throwable|null $template */
-    public function assert(mixed $input, array|string|Throwable|callable|null $template = null): void
+    public function check(mixed $input, array|string|Throwable|callable|null $template = null): void
```
henriquemoody, line 90: @alganet, you kinda convinced me that we might want to bring this one back.

tests/feature/Validators/ShortCircuitTest.php
```diff
@@ -0,0 +1,112 @@
...
+        ->and($fullMessage)->toBe('- `false` must evaluate to `true`')
+        ->and($messages)->toBe(['trueVal' => '`false` must evaluate to `true`']),
+));
+
+test('With recursive', catchAll(
+    fn() => v::shortCircuit(v::each(v::intType()))->assert([1, 2, '3', 4]),
+    fn(string $message, string $fullMessage, array $messages) => expect()
+        ->and($message)->toBe('`.2` must be an integer')
+        ->and($fullMessage)->toBe('- `.2` must be an integer')
+        ->and($messages)->toBe([0 => '`.2` must be an integer']),
```
henriquemoody: Gotta do some work on the key here

docs/validators/ShortCircuit.md
```diff
@@ -0,0 +1,79 @@
...
+# ShortCircuit
+
+- `ShortCircuit()`
+- `ShortCircuit(Validator ...$validators)`
+
+Validates the input against a series of validators, stopping at the first failure.
+
+Like PHP's `&&` operator, it uses short-circuit evaluation: once the outcome is determined, remaining validators are
+skipped. Unlike [AllOf](AllOf.md), which evaluates all validators and collects all failures, `ShortCircuit` returns
+immediately.
```
alganet, line 17: We should compare somewhere what `v::shortCircuit()` does (localized stop-on-first-error behavior) and what `->check()` does (global stop-on-first-error behavior) and their differences, with helpful examples.

src/ValidatorBuilder.php
alganet: Personally, in your PRs context, I wish this method was called `shortCircuit` instead.

The only problem with that is that it collides with `v::shortCircuit` (adding a validator to the chain).

I would go for `evaluateShortCircuit`, in order to mirror what the interface does. It estabilishes a clear pattern that follows `$validator::evaluate / ValidatorBuilder::evaluate` introducing the same symmetry `$validator::evaluateShortCircuit / ValidatorBuilder::evaluateShortCircuit`.

Similarly, if we go for `check`, then I think we should keep the symmetry `$validator::check / ValidatorBuilder::check`.

This symmetry is not only cosmetic/naming-related. It helps with searching code, for example. It takes one or two hops reading classes until I connect that `new ShortCircuit` yields an inner flow that revolves around `evaluateShortCircuit` internally. If the names are similar, as it is with `evaluate()` today, it becomes easier to understand the abstraction.

tests/benchmark/CompositeValidatorsBench.php
```diff
@@ -0,0 +1,78 @@
...
+    public function provideValidatorBuilder(): Generator
+    {
+        yield 'allOf(10)' => ['allOf', $this->buildValidators(10)];
+        yield 'oneOf(10)' => ['oneOf', $this->buildValidators(10)];
+        yield 'anyOf(10)' => ['anyOf', $this->buildValidators(10)];
+        yield 'noneOf(10)' => ['noneOf', $this->buildValidators(10)];
+        yield 'allOf(100)' => ['allOf', $this->buildValidators(100)];
+        yield 'oneOf(100)' => ['oneOf', $this->buildValidators(100)];
+        yield 'anyOf(100)' => ['anyOf', $this->buildValidators(100)];
+        yield 'noneOf(100)' => ['noneOf', $this->buildValidators(100)];
```
alganet, line 77: On #1633, I decided not to tackle `Each`, but you did. You should also include a benchmark for it.

alganet (commented):
Overall seems like a solid prototype of what #1633 also aims to achieve, with similar gains and more ambitious (tackling `Each`, for example).

If we're aiming for completeness (no multi-step PRs), I do believe `KeySet` should also be available for global short circuit. Otherwise, I would prefer if you did it in smaller steps (logical AllOf/OneOf/NoneOf/AnyOf/Circuit first, then the others that require more heuristic to short circuit properly). It's up to you (single-shot or smaller PRs, I'm okay with either choice).

src/ValidatorBuilder.php
henriquemoody: I'm a bit confused by your comment, because this check() method does something different from the `evaluateShortCircuit()`, we can't have symmetry of `$validator::check / ValidatorBuilder::check` because they do different things. Now that you mentioned, it could be that `ValidatorBuilder` also implements `ShortCircuitable`, hence could also implement the method `evaluateShortCircuit()`.


My choice of using `check()` here is simply for familiarity, since that's what we had on v2, but we could have something like `assertShortCircuit()`

alganet: You're right. I'm confused.

> Now that you mentioned, it could be that ValidatorBuilder also implements ShortCircuitable, hence could also implement the method evaluateShortCircuit().

Great idea! That would be awesome.

henriquemoody: I made that change, and it was indeed needed since ValidatorBuilder handles multiple validators

tests/benchmark/CompositeValidatorsBench.php
henriquemoody: I added some for `Each`, `All`, and `Domain`, and I was surprised to see that it actually became slower in the majority of cases. I have to investigate a bit further. I think the different is negligible, but I would like to dig a bit deeper because we're creating less objects, I don't see an obvious reason.

alganet: You're not providing `Each` and `All` many items. You're passing only three items: 42, 43, 44.

The logical validators break on the list of validators. `Each` and `All` will execute all of them.

So, in order to properly evaluate them, you need to make series of random input items that are likely to break at different points instead of series of validators that run against the same input.

```php
ValidatorBuilder::__callStatic(...$params)->isValid($randomItems);
```

For the validator chain, I used 10 and 100 (a small chain with 10 validators, a large chain with 100 validators).

I would recommend using 10 and 100 as well for the number of items.

Small losses with fewer number of validators/items are expected due to the increase in abstraction. This is what you are observing because you're only using a size of 3: `[42, 43, 44]`. You can further track down the expensive abstractions doing `bench:profile` to see hot paths (methods that are most called) and further optimize. 2~3% loss in small chains is irrelevant though.

I also recommend adding a benchmark for ShortCircuit itself, since it's the core of it all.

henriquemoody:
@alganet, I think this might be blocking our release, but I'm short on time in the upcoming days. I wanted to track the performance issue with each and all, but I don't have much time at moment, and I'm prioritizing reviewing your merge request. I think we could try working on performance of all and each later on, and merge this as is, but let me know if you think otherwise.

PR #1666, opened 2026-02-01
Use Respect\Config instead of php-di

alganet:
Replaces the php-di implementation with Respect\Config, keeps PSR-11 compatibility.

Respect\Config is much lighter (20Kb package, php-di is 80kb), and also much faster (80% gain on cold ValidatorBuild::init) than uncompiled php-di.

While compiled php-di containers would be in theory even faster, it is awkward to commit their artifacts.

Users that want ultimate performance can still leverage php-di with compilation by doing it (together with their app DI defs) and doing `setContainer` on our `ContainerRegistry`.

---

⚠️ This change is experimental ⚠️ Respect\Config 3.0 is not launched yet.

**Benchmarks**:

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,8 +21,8 @@
     },
     "require": {
         "php": ">=8.5",
-        "php-di/php-di": "^7.1",
         "psr/container": "^2.0",
+        "respect/config": "dev-master",
         "respect/string-formatter": "^1.0",
         "respect/stringifier": "^3.0",
         "symfony/polyfill-mbstring": "^1.33"
--- a/docs/messages/placeholder-conversion.md
+++ b/docs/messages/placeholder-conversion.md
@@ -38,7 +38,7 @@ See [PlaceholderFormatter][] documentation for more information on creating cust
 
 ## Container configuration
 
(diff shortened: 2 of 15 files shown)


alganet:
I know this change would make some plans (autowiring sokil) harder. However, autowire *is not PSR-11*, and it would imply a soft dependency on php-di (or any other autowiring-enabled framework).

Since version 6, php-di has been adding bloated features that dramatically reduce its runtime performance, and instead [they've been focusing on compilation instead](https://php-di.org/doc/performances.html). 

While compilation is great (I used it in my [swoole-microfw](https://github.com/alganet/swoole-microfw) experiment), as a library it is awkward for us to deliver compiled containers. That kind of workflow works best when the user compiles the container for his/her own application instead (their stuff + our stuff all compiled together).

Respect\Config is much leaner and has outstanding runtime performance, with still more gains to squeeze (it's basically a PHP 5.3 codebase ported to PHP 8.5 recently). My plan is to keep this as a draft until I launch Config 3.0.

alganet:
I will close and archive this experiment for further reference.

Issue #1668, opened 2026-02-01
Deep Name Collision


alganet:
Affecting: 3.0 (unreleased)

Consider the following validator:

```php
v::allOf(
    v::named('alpha', v::allOf(
        v::contains('quick'),
        v::contains('fox'),
    )),
    v::named('zeta', v::allOf(
        v::contains('lorem'),
        v::contains('ipsum'),
    ))->assert("foo bar";
```

As an user, I would expect to see the messages neatly grouped by their names. Since the asserted string does not contain any of the words, we should see them all.

Currently, this is the result:

```
  'allOf' => [
      0 => '"foo bar baz" must contain "quick"',
      1 => '"foo bar baz" must contain "fox"',
      2 => [
          0 => '"foo bar baz" must contain "lorem"',
          1 => '"foo bar baz" must contain "ipsum"',
      ],
  ],
```

This is not only missing the named groups, but it's also malformed, with grouping inconsistent with the structure of the validators themselves.

What I expected is something like:

```
  '__root__' => '"foo bar baz" must pass all the rules',
  0 => [
      '__root__' => 'alpha must pass all the rules',
      0 => '"foo bar baz" must contain "quick"',
      1 => '"foo bar baz" must contain "fox"',
  ],
  1 => [
      '__root__' => 'zeta must pass all the rules',
      0 => '"foo bar baz" must contain "lorem"',
      1 => '"foo bar baz" must contain "ipsum"',
  ],
```

This is also consistent with the same unnamed version (which would produce identical `__root__` entries, thus preventing it from using as the key.

PR #1669, opened 2026-02-01
Handle more collision scenarios on NestedArrayFormatter

alganet:
See #1668 for more info.

There is a new test "Deep name collision" on AllOfTest that exemplifies the kind of collision this change solves.

Some extra `__root__` keys needed to be added to a select few other scenarios for consistency.

### DIFF
--- a/src/Message/Formatter/NestedArrayFormatter.php
+++ b/src/Message/Formatter/NestedArrayFormatter.php
@@ -15,7 +15,7 @@ use Respect\Validation\Message\ArrayFormatter;
 use Respect\Validation\Message\Renderer;
 use Respect\Validation\Result;
 
-use function array_reduce;
+use function array_values;
 use function count;
 use function current;
 use function is_array;
@@ -31,25 +31,30 @@ final readonly class NestedArrayFormatter implements ArrayFormatter
     public function format(Result $result, Renderer $renderer, array $templates): array
     {
         if ($result->children === []) {
-            return [$this->getKey($result) => $renderer->render($result, $templates)];
+            return [$result->path->value ?? $result->id->value => $renderer->render($result, $templates)];
         }
 
-        [$children, $hasString] = $this->prepareChildren($result->children);
+        $hasStringKey = false;
+        foreach ($result->children as $child) {
+            if (!is_numeric($child->path->value ?? $child->id->value)) {
+                $hasStringKey = true;
+                break;
+            }
+        }
 
-        $messages = array_reduce(
-            $children,
-            fn(array $messages, array $item) => $this->appendMessage(
+        $messages = [];
+        $childCount = count($result->children);
+        foreach ($result->children as $child) {
+            $messages = $this->formatChild(
                 $messages,
-                $item['key'],
-                $item['child'],
+                $child,
                 $renderer,
                 $templates,
-                $hasString,
+                $hasStringKey,
                 $result,
-                isParentVisible: count($children) > 1,
-            ),
-            [],
-        );
+                $childCount > 1,
+            );
+        }
 
         if (count($messages) > 1) {
             return ['__root__' => $renderer->render($result, $templates)] + $messages;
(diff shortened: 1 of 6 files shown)


src/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -96,33 +97,84 @@ private function appendMessage(
         Result $parent,
         bool $isParentVisible,
     ): array {
-        if ($hasString && is_numeric($key)) {
-            $key = $child->id->value;
-        }
+        $key = $this->normalizeKey($key, $hasString, $child);
+
+        $appended = $this->renderChild($child, $renderer, $templates, $parent, $isParentVisible);
```
henriquemoody: Could you rename this variable to `$current` or `$rendered` or `$message` or something like that? The problem I had reading this code, especially in the handle* methods is that I was wondering if the message was already appended by the time it arrived there.

src/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -96,33 +97,84 @@ private function appendMessage(
...
+    {
+        if (is_array($messages[$key])) {
+            if (isset($messages[$key]['__root__'])) {
+                $messages[$key] = [$messages[$key]];
+            }
+
+            $messages[$key][] = $appended;
+        } else {
+            $messages[$key] = [$messages[$key], $appended];
+        }
```
henriquemoody, line 131:
```suggestion
        if (!is_array($messages[$key])) {
            $messages[$key] = [$messages[$key], $appended];
        }
```
If you make this change, all feature tests pass, which means we're not _really_ using it. So, either we need to create a feature test or you might be able to remove it. At the same time, the unit test you added won't pass. That makes me confused, because I wonder how we could reproduce it with real validators.

src/Message/Formatter/NestedArrayFormatter.php
henriquemoody: I asked Claude Code to simplify this formatter, it came up with something pretty neat: https://github.com/alganet/Validation/pull/1

alganet: I named it `$appended` because the method is called `appendMessage`. That is what it appended 😅

alganet: @henriquemoody I left a comment in the PR.

I like some of Claude's tricks, but I don't like his change overall. He de-extracted lots of methods and made a long `formatChild`.

Are you OK with that? If you are, I can commit his version.

src/Message/Formatter/NestedArrayFormatter.php
alganet: Our tests for collisions are very poor. I'm sorry for making just the unit one.

I just added a complete real world feature test that showcases how this might happen, in addition to another variation of the previous test without the names.

tests/feature/Validators/AllOfTest.php
```diff
@@ -123,3 +123,123 @@
             'uppercase' => 'Template for "uppercase"',
         ]),
 ));
+
+test('Deep name collision', catchAll(
```
henriquemoody: Can you move this one to `tests/feature/Issue1668Test.php`?

src/Message/Formatter/NestedArrayFormatter.php
henriquemoody: I wrote on the PR I sent to your project, I think in this case, I prefer a longer method

henriquemoody: The name `$childMessage` is clearer than `$appended` because when the variable is create, it's not appended. When I read the code first time, I thought the message was already in the $messages, and struggles to understand what was happening to it at that point

src/Message/Formatter/NestedArrayFormatter.php
```diff
@@ -31,25 +31,30 @@
...
     {
         if ($result->children === []) {
-            return [$this->getKey($result) => $renderer->render($result, $templates)];
+            return [$result->path->value ?? $result->id->value => $renderer->render($result, $templates)];
         }
 
-        [$children, $hasString] = $this->prepareChildren($result->children);
+        $hasStringKey = false;
+        foreach ($result->children as $child) {
+            if (!is_numeric($child->path->value ?? $child->id->value)) {
```
henriquemoody, line 39: The the `$child->path->value ?? $child->id->value` is used twice, this one could be good to have in a separated method now that I think of it.

Nitpick, don't need to change

henriquemoody (approved):

PR #1670, opened 2026-02-02
Standardize and improve validation message templates

alganet:
- Remove redundant "valid" prefix: Date, DateTime, DateTimeDiff, Domain, Email, Iban, Imei, Ip, Isbn, Json, LanguageCode, LeapDate, LeapYear, Luhn, MacAddress, NfeAccessKey, Nif, Nip, Pesel, Phone, Pis, PolishIdCard, PostalCode, Roman, Slug, Tld, Url, Uuid, Version.

 - Remove redundant "value" suffix ArrayVal, BoolVal, Countable, FloatVal, IntVal, IterableVal, NumericVal, ScalarVal, StringVal.

 - Standardize "consist only of" phrasing Alnum, Alpha, Cntrl, Consonant, Digit, Graph, Lowercase, Printable, Punct, Space, Spaced, Uppercase, Vowel, Xdigit.

 - Improve file accessibility messages Directory, Executable, File, Image, Readable, SymbolicLink, Writable.

 - Improve grammar and article usage CreditCard, Extension, Mimetype, Regex, Size.

### DIFF
--- a/src/Validators/Alnum.php
+++ b/src/Validators/Alnum.php
@@ -22,13 +22,13 @@ use function ctype_alnum;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must contain only letters (a-z) and digits (0-9)',
-    '{{subject}} must not contain letters (a-z) or digits (0-9)',
+    '{{subject}} must consist only of letters (a-z) and digits (0-9)',
+    '{{subject}} must not consist only of letters (a-z) or digits (0-9)',
     self::TEMPLATE_STANDARD,
 )]
 #[Template(
-    '{{subject}} must contain only letters (a-z), digits (0-9), and {{additionalChars}}',
-    '{{subject}} must not contain letters (a-z), digits (0-9), or {{additionalChars}}',
+    '{{subject}} must consist only of letters (a-z), digits (0-9), or {{additionalChars}}',
+    '{{subject}} must not consist only of letters (a-z), digits (0-9), or {{additionalChars}}',
     self::TEMPLATE_EXTRA,
 )]
 final class Alnum extends FilteredString
--- a/src/Validators/Alpha.php
+++ b/src/Validators/Alpha.php
@@ -22,13 +22,13 @@ use function ctype_alpha;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must contain only letters (a-z)',
-    '{{subject}} must not contain letters (a-z)',
+    '{{subject}} must consist only of letters (a-z)',
+    '{{subject}} must not consist only of letters (a-z)',
     self::TEMPLATE_STANDARD,
 )]
 #[Template(
-    '{{subject}} must contain only letters (a-z) and {{additionalChars}}',
-    '{{subject}} must not contain letters (a-z) or {{additionalChars}}',
+    '{{subject}} must consist only of letters (a-z) or {{additionalChars}}',
+    '{{subject}} must not consist only of letters (a-z) or {{additionalChars}}',
     self::TEMPLATE_EXTRA,
 )]
 final class Alpha extends FilteredString
--- a/src/Validators/Charset.php
+++ b/src/Validators/Charset.php
@@ -30,8 +30,8 @@ use function mb_list_encodings;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must only contain characters from the {{charset|raw}} charset',
-    '{{subject}} must not contain any characters from the {{charset|raw}} charset',
+    '{{subject}} must consist only of characters from the {{charset|list:or}} character-set',
+    '{{subject}} must not consist only of characters from the {{charset|list:or}} character-set',
 )]
 final readonly class Charset implements Validator
 {
--- a/src/Validators/Cnh.php
+++ b/src/Validators/Cnh.php
@@ -27,8 +27,8 @@ use function preg_replace;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid CNH number',
-    '{{subject}} must not be a valid CNH number',
+    '{{subject}} must be a CNH',
+    '{{subject}} must not be a CNH',
 )]
 final class Cnh extends Simple
 {
--- a/src/Validators/Cnpj.php
+++ b/src/Validators/Cnpj.php
@@ -29,8 +29,8 @@ use function str_split;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid CNPJ number',
-    '{{subject}} must not be a valid CNPJ number',
+    '{{subject}} must be a CNPJ',
+    '{{subject}} must not be a CNPJ',
 )]
 final class Cnpj extends Simple
 {
--- a/src/Validators/Cpf.php
+++ b/src/Validators/Cpf.php
@@ -30,8 +30,8 @@ use function preg_replace;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid CPF number',
-    '{{subject}} must not be a valid CPF number',
+    '{{subject}} must be a CPF',
+    '{{subject}} must not be a CPF',
 )]
 final class Cpf extends Simple
 {
--- a/src/Validators/Date.php
+++ b/src/Validators/Date.php
@@ -30,8 +30,8 @@ use function strtotime;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid date in the format {{sample}}',
-    '{{subject}} must not be a valid date in the format {{sample}}',
+    '{{subject}} must be a date in the {{sample}} format',
+    '{{subject}} must not be a date in the {{sample}} format',
 )]
 final readonly class Date implements Validator
 {
--- a/src/Validators/DateTime.php
+++ b/src/Validators/DateTime.php
@@ -29,13 +29,13 @@ use function strtotime;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid date/time',
-    '{{subject}} must not be a valid date/time',
+    '{{subject}} must be a date/time',
+    '{{subject}} must not be a date/time',
     self::TEMPLATE_STANDARD,
 )]
 #[Template(
-    '{{subject}} must be a valid date/time in the format {{sample}}',
-    '{{subject}} must not be a valid date/time in the format {{sample}}',
+    '{{subject}} must be a date/time in the {{sample}} format',
+    '{{subject}} must not be a date/time in the {{sample}} format',
     self::TEMPLATE_FORMAT,
 )]
 final class DateTime implements Validator
--- a/src/Validators/DateTimeDiff.php
+++ b/src/Validators/DateTimeDiff.php
@@ -34,13 +34,13 @@ use function in_array;
     self::TEMPLATE_CUSTOMIZED,
 )]
 #[Template(
-    'For comparison with {{now|raw}}, {{subject}} must be a valid datetime',
-    'For comparison with {{now|raw}}, {{subject}} must not be a valid datetime',
+    'For comparison with {{now|raw}}, {{subject}} must be a datetime',
+    'For comparison with {{now|raw}}, {{subject}} must not be a datetime',
     self::TEMPLATE_NOT_A_DATE,
 )]
 #[Template(
-    'For comparison with {{now|raw}}, {{subject}} must be a valid datetime in the format {{sample|raw}}',
-    'For comparison with {{now|raw}}, {{subject}} must not be a valid datetime in the format {{sample|raw}}',
+    'For comparison with {{now|raw}}, {{subject}} must be a datetime in the format {{sample|raw}}',
+    'For comparison with {{now|raw}}, {{subject}} must not be a datetime in the format {{sample|raw}}',
     self::TEMPLATE_WRONG_FORMAT,
 )]
 final readonly class DateTimeDiff implements Validator
--- a/src/Validators/Decimal.php
+++ b/src/Validators/Decimal.php
@@ -25,8 +25,8 @@ use function var_export;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must have {{decimals}} decimals',
-    '{{subject}} must not have {{decimals}} decimals',
+    '{{subject}} must have {{decimals}} decimal places',
+    '{{subject}} must not have {{decimals}} decimal places',
 )]
 final readonly class Decimal implements Validator
 {
--- a/src/Validators/Digit.php
+++ b/src/Validators/Digit.php
@@ -22,13 +22,13 @@ use function ctype_digit;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must contain only digits (0-9)',
-    '{{subject}} must not contain digits (0-9)',
+    '{{subject}} must consist only of digits (0-9)',
+    '{{subject}} must not consist only of digits (0-9)',
     self::TEMPLATE_STANDARD,
 )]
 #[Template(
-    '{{subject}} must contain only digits (0-9) and {{additionalChars}}',
-    '{{subject}} must not contain digits (0-9) and {{additionalChars}}',
+    '{{subject}} must consist only of digits (0-9) or {{additionalChars}}',
+    '{{subject}} must not consist only of digits (0-9) or {{additionalChars}}',
     self::TEMPLATE_EXTRA,
 )]
 final class Digit extends FilteredString
--- a/src/Validators/Extension.php
+++ b/src/Validators/Extension.php
@@ -25,8 +25,8 @@ use const PATHINFO_EXTENSION;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must have {{extension}} extension',
-    '{{subject}} must not have {{extension}} extension',
+    '{{subject}} must have the {{extension}} extension',
+    '{{subject}} must not have the {{extension}} extension',
 )]
 final readonly class Extension implements Validator
 {
--- a/src/Validators/FloatVal.php
+++ b/src/Validators/FloatVal.php
@@ -26,8 +26,8 @@ use const FILTER_VALIDATE_FLOAT;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a float value',
-    '{{subject}} must not be a float value',
+    '{{subject}} must be a floating-point number',
+    '{{subject}} must not be a floating-point number',
 )]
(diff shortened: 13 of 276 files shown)


alganet:
GitHub is down:

https://www.githubstatus.com/incidents/xwn6hjps36ty

Please wait for it before any action on this PR.

src/Validators/Alnum.php
```diff
@@ -22,13 +22,13 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must contain only letters (a-z) and digits (0-9)',
-    '{{subject}} must not contain letters (a-z) or digits (0-9)',
+    '{{subject}} must consist only of letters (a-z) and digits (0-9)',
```
henriquemoody, line 25:
```suggestion
    '{{subject}} must consist only of letters and digits',
```
I don't think we need those, and if we really want to be correct, we have to say A-Z are also allowed.

src/Validators/Alpha.php
```diff
@@ -22,13 +22,13 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must contain only letters (a-z)',
-    '{{subject}} must not contain letters (a-z)',
+    '{{subject}} must consist only of letters (a-z)',
```
henriquemoody, line 25:
```suggestion
    '{{subject}} must consist only of letters',
```
Same here

src/Validators/Charset.php
```diff
@@ -30,8 +30,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must only contain characters from the {{charset|raw}} charset',
-    '{{subject}} must not contain any characters from the {{charset|raw}} charset',
+    '{{subject}} must consist only of characters from the {{charset|raw}} character-set',
```
henriquemoody:
```suggestion
    '{{subject}} must consist only of characters from the {{charset|list:or}} character-set',
```

Since this is an array of strings, I think this will make it much clearer. You might want to play with plural by having a different template if the number of charsets is bigger than 1

src/Validators/Cnh.php
```diff
@@ -27,8 +27,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid CNH number',
-    '{{subject}} must not be a valid CNH number',
+    '{{subject}} must be a CNH number',
```
henriquemoody:
```suggestion
    '{{subject}} must be a CNH',
```
Following what you did for BSN. I think this makes more sense.

src/Validators/Cnpj.php
```diff
@@ -29,8 +29,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid CNPJ number',
-    '{{subject}} must not be a valid CNPJ number',
+    '{{subject}} must be a CNPJ number',
```
henriquemoody:
```suggestion
    '{{subject}} must be a CNPJ',
```
Following what you did for BSN. I think this makes more sense.

src/Validators/Cpf.php
```diff
@@ -30,8 +30,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid CPF number',
-    '{{subject}} must not be a valid CPF number',
+    '{{subject}} must be a CPF number',
```
henriquemoody:
```suggestion
    '{{subject}} must be a CPF',
```
Following what you did for BSN. I think this makes more sense.

src/Validators/Date.php
```diff
@@ -30,8 +30,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid date in the format {{sample}}',
-    '{{subject}} must not be a valid date in the format {{sample}}',
+    '{{subject}} must be a date in the format {{sample}}',
```
henriquemoody:
```suggestion
    '{{subject}} must be a date in the {{sample}} format',
```

But both work, I think my suggestion is better as a sentence, but yours is better for visibility

src/Validators/DateTime.php
```diff
@@ -29,13 +29,13 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid date/time',
-    '{{subject}} must not be a valid date/time',
+    '{{subject}} must be a date/time',
```
henriquemoody, line 32:
```suggestion
    '{{subject}} must be a date and time',
```

src/Validators/Date.php
henriquemoody:
```suggestion
    '{{subject}} must follow the format: {{sample}}',
```

We could also consider that, the {{sample}} will imply it's a date anyways.

src/Validators/DateTime.php
```diff
@@ -29,13 +29,13 @@
...
+    '{{subject}} must not be a date/time',
     self::TEMPLATE_STANDARD,
 )]
 #[Template(
-    '{{subject}} must be a valid date/time in the format {{sample}}',
-    '{{subject}} must not be a valid date/time in the format {{sample}}',
+    '{{subject}} must be a date/time in the format {{sample}}',
```
henriquemoody:
```suggestion
    '{{subject}} must be a date and time in the {{sample}} format',
```
or
```suggestion
    '{{subject}} must follow the format: {{sample}}',
```

src/Validators/DateTimeDiff.php
```diff
@@ -34,13 +34,13 @@
     self::TEMPLATE_CUSTOMIZED,
 )]
 #[Template(
-    'For comparison with {{now|raw}}, {{subject}} must be a valid datetime',
-    'For comparison with {{now|raw}}, {{subject}} must not be a valid datetime',
+    'For comparison with {{now|raw}}, {{subject}} must be a datetime',
```
henriquemoody, line 37:
```suggestion
    '{{subject}} must be a valid date and time to compare it with {{now}}',
```

or 
```suggestion
    'to compare it with {{now}}',
```
And we make this one as adjacent of `DateTime`

src/Validators/Decimal.php
```diff
@@ -25,8 +25,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must have {{decimals}} decimals',
-    '{{subject}} must not have {{decimals}} decimals',
+    '{{subject}} must have {{decimals}} decimal places',
```
henriquemoody, line 28:
```suggestion
    '{{subject}} must have exactly {{decimals}} decimal places',
```
But the current one works well too

src/Validators/Digit.php
```diff
@@ -22,13 +22,13 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must contain only digits (0-9)',
-    '{{subject}} must not contain digits (0-9)',
+    '{{subject}} must consist only of digits (0-9)',
```
henriquemoody, line 25:
```suggestion
    '{{subject}} must consist only of digits',
```

src/Validators/Digit.php
```diff
@@ -22,13 +22,13 @@
...
+    '{{subject}} must not consist only of digits (0-9)',
     self::TEMPLATE_STANDARD,
 )]
 #[Template(
-    '{{subject}} must contain only digits (0-9) and {{additionalChars}}',
-    '{{subject}} must not contain digits (0-9) and {{additionalChars}}',
+    '{{subject}} must consist only of digits (0-9) or {{additionalChars}}',
```
henriquemoody, line 30:
```suggestion
    '{{subject}} must consist only of digits or {{additionalChars}}',
```

src/Validators/Extension.php
```diff
@@ -25,8 +25,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must have {{extension}} extension',
-    '{{subject}} must not have {{extension}} extension',
+    '{{subject}} must have the {{extension}} extension',
```
henriquemoody, line 28:
```suggestion
    '{{subject}} must have the {{extension|quote}} extension',
```
Not sure, though

src/Validators/FloatVal.php
```diff
@@ -26,8 +26,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a float value',
-    '{{subject}} must not be a float value',
+    '{{subject}} must be a float',
```
henriquemoody:
```suggestion
    '{{subject}} must be a decimal number',
```
I think the *Val validators are way more user-friendly, hence I think a message like that makes more sense.

Alternatively 

```suggestion
    '{{input}} must be a floating-point number',
```
More scientific, but it works.

src/Validators/IntVal.php
```diff
@@ -29,8 +29,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be an integer value',
-    '{{subject}} must not be an integer value',
+    '{{subject}} must be an integer',
```
henriquemoody, line 32:
```suggestion
    '{{subject}} must be a number without decimals',
```
Same argument about the *Val validators being more end-user focused

src/Validators/LanguageCode.php
```diff
@@ -29,8 +29,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid language code',
-    '{{subject}} must not be a valid language code',
+    '{{subject}} must be a language code',
```
henriquemoody, line 32:
```suggestion
    '{{subject}} must be an ISO 639-3 language code',
```
I'm not so sure about this one, but if we go down that road  we should change all of the *Code validators

src/Validators/Length.php
```diff
@@ -36,8 +36,8 @@
     self::TEMPLATE_STANDARD,
 )]
 #[Template(
-    '{{subject}} must be a countable value or a string',
-    '{{subject}} must not be a countable value or a string',
+    '{{subject}} must be countable or a string',
```
henriquemoody, line 39:
```suggestion
    '{{subject}} must be a string or a countable value',
```

In this context, just "countable" sound a bit weird to me

src/Validators/Luhn.php
```diff
@@ -25,8 +25,8 @@
 /** @see https://en.wikipedia.org/wiki/Luhn_algorithm */
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid Luhn number',
-    '{{subject}} must not be a valid Luhn number',
+    '{{subject}} must be a Luhn number',
```
henriquemoody, line 28: Got some suggestions for Gemini:

* `{{subject}} must be a valid Luhn-compliant number`
* `{{subject}} must pass the Luhn checksum`

Not sure "Luhn number" is a _thing_

src/Validators/Phone.php
```diff
@@ -34,13 +34,13 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid telephone number',
-    '{{subject}} must not be a valid telephone number',
+    '{{subject}} must be a telephone number',
```
henriquemoody:
```suggestion
    '{{subject}} must be a phone number',
```
Telephone sounds very old fashioned

src/Validators/Phone.php
```diff
@@ -34,13 +34,13 @@
...
+    '{{subject}} must not be a telephone number',
     self::TEMPLATE_INTERNATIONAL,
 )]
 #[Template(
-    '{{subject}} must be a valid telephone number for country {{countryName|trans}}',
-    '{{subject}} must not be a valid telephone number for country {{countryName|trans}}',
+    '{{subject}} must be a telephone number for country {{countryName|trans}}',
```
henriquemoody:
```suggestion
    '{{subject}} must be a valid phone number for {{countryName}}',
```

src/Validators/PostalCode.php
```diff
@@ -33,8 +33,8 @@
 /** @see http://download.geonames.org/export/dump/countryInfo.txt */
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid postal code on {{countryCode}}',
-    '{{subject}} must not be a valid postal code on {{countryCode}}',
+    '{{subject}} must be a postal code on {{countryCode}}',
```
henriquemoody:
```suggestion
    '{{subject}} must be a postal code for {{countryCode}}',
```

or 
```suggestion
    '{{subject}} must be a postal code in {{countryCode}}',
```

But "on" it grammatically incorrect

src/Validators/Satisfies.php
```diff
@@ -24,8 +24,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be valid',
-    '{{subject}} must be invalid',
+    '{{subject}} must be satisfied by the given callback',
```
henriquemoody: This feels like leaky abstraction, I'm not sure we want to expose that much details to the user.

src/Validators/ScalarVal.php
```diff
@@ -21,8 +21,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a scalar value',
-    '{{subject}} must not be a scalar value',
+    '{{subject}} must be a scalar',
```
henriquemoody, line 24: In this case, I think how it was is better, just "scalar" is a bit cryptic, and since scalar is a pseudo-type, saying "scalar value" sounds better.

src/Validators/Time.php
```diff
@@ -30,8 +30,8 @@
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 #[Template(
-    '{{subject}} must be a valid time in the format {{sample}}',
-    '{{subject}} must not be a valid time in the format {{sample}}',
+    '{{subject}} must be a time in the format {{sample}}',
```
henriquemoody:
```suggestion
    '{{subject}} must be a time in the {{sample}} format',
```

Or

```suggestion
    '{{subject}} must follow the format: {{sample}}',
```

src/Validators/Uuid.php
```diff
@@ -34,13 +34,13 @@
...
-    '{{subject}} must be a valid UUID',
-    '{{subject}} must not be a valid UUID',
+    '{{subject}} must be a UUID',
+    '{{subject}} must not be a UUID',
     self::TEMPLATE_STANDARD,
 )]
 #[Template(
-    '{{subject}} must be a valid UUID version {{version|raw}}',
-    '{{subject}} must not be a valid UUID version {{version|raw}}',
+    '{{subject}} must be a UUID version {{version|raw}}',
```
henriquemoody:
```suggestion
    '{{subject}} must be a UUID v{{version|raw}}',
```

What do you think about this one?

src/Validators/Alnum.php
alganet: `ç` and `á` are letters but `Alnum` and other validators that use `ctype_*` functions do not allow them.

It's not much about the case, but the range support (only simple letters in the ASCII table).

src/Validators/DateTime.php
alganet: We support dates without time in this validator.

src/Validators/DateTimeDiff.php
alganet: If we want to go that route, the correct phrase would be "In order to compare {{subject}}...". 

> And we make this one as adjacent of DateTime

That sounds like a deeper refactor. Not against it, but out of scope for this PR.

src/Validators/IntVal.php
alganet: `0xAAFF32` is a number without decimals. It's not an integer though.

I understand how these validators might be used, and I see your point. In that case, I would prefer `must be a whole number`.

henriquemoody: I thought about whole number as well, but in _theory_ whole numbers are only positive

src/Validators/LanguageCode.php
alganet: If we're going the usability route, `must be a 2-letter language code` or `must be a 3-letter language code` are better than dropping the ISO bomb.

src/Validators/Luhn.php
alganet: The friendliest would be `must pass digit verification`, without mentioning Luhn.

src/Validators/ScalarVal.php
alganet: This would validate an input that supports both scalar and complex values. It's a deeply technical validator. For an API that validates against a schema (GraphQL, json-schema, etc), for example.

"scalar" is not a pseudo-type, it's a concept that denotes vector space (the thing grows only in one dimension). PHP (and many other languages) adopted the terminology to identify values that can be referenced by a single pointer (as opposed to a list or array which have many pointers).

So, the user here is a programmer or a mathematician. In this context, it's not cryptic at all.

Maybe there are creative ways of using `v::scalar` for other scenarios (a front-end form for example). I could not think of any.

src/Validators/Charset.php
alganet: Nice catch with the `list:or`. Singular is fine, since we don't actually match multiple datasets, we match the first one as per `mb_detect_encoding` semantics (it's a logic or).

Plural would be actually innacurate, since it would imply the characters must match _both_ charsets (which do happen across ASCII, ISO-8859-1 and UTF-8 for example). However, mb_detect_encoding picks just the first that matches.

src/Validators/Date.php
alganet: I think having saying it needs to be a date is important. `2025-02-30` follows a format, but it's not a date.

src/Validators/Decimal.php
alganet: Probably a deeper more meaningful refactor would be to accept a comparator. In that case, we're likely to refactor the messages again.

src/Validators/Extension.php
alganet: It already quotes as it is:

 - `"filename.gif" must not have the "gif" extension`

src/Validators/IntVal.php
alganet: Perhaps "an integer" is the best one then.

src/Validators/DateTimeDiff.php
henriquemoody: I like the `In order to compare {{subject}}`

src/Validators/IntVal.php
henriquemoody: Yes, it's better than.

src/Validators/Luhn.php
henriquemoody: Let's leave it as is, then

src/Validators/Extension.php
henriquemoody: Bad name, the `|quote` would add backticks

src/Validators/Alnum.php
henriquemoody: fail enough

src/Validators/LanguageCode.php
henriquemoody: Yeah, you're right

src/Validators/Extension.php
alganet: Perhaps `|code` would be nice!

henriquemoody: Yeah, it's complicated though, because it depends on the quoter from the stringifier, so it makes more sense in the context of stringifier/formatter, but it's confusing here.

PR #1671, opened 2026-02-03
Update SPDX checker, and run the automatic fixer in the codebase

henriquemoody:


src-dev/Commands/LintSpdxCommand.php
```diff
@@ -3,41 +3,48 @@
...
+ * SPDX-FileContributor: Alexandre Gomes Gaigalas <alganet@gmail.com>
+ * SPDX-FileContributor: Henrique Moody <henriquemoody@gmail.com>
  */
 
 declare(strict_types=1);
 
 namespace Respect\Dev\Commands;
 
-use Respect\Validation\Exceptions\ValidationException;
-use Respect\Validation\ValidatorBuilder;
```
alganet, line 13: While I appreciate the new checker, I think removing a use case of ourselves using our own tools is bad.

There are several parts of the new implementation that could use validation internally.

If we can't use it, how can we ask our uses to?

src-dev/Spdx/Contributor.php
```diff
@@ -0,0 +1,53 @@
...
+        'github@jigsoft.co.za' => ['Nick Lombard', 'github@jigsoft.co.za'],
+        'nick@jigsoft.co.za' => ['Nick Lombard', 'github@jigsoft.co.za'],
+        'emmersonsiqueira@gmail.com' => ['Emmerson Siqueira', 'emmersonsiqueira@gmail.com'],
+        'jayson.reis@sabbre.com.br' => ['Jayson Reis', 'santosdosreis@gmail.com'],
+        'kolyshkin@.sqlmaze.local' => ['Kir Kolyshkin', 'kolyshkin@gmail.com'],
+        'pathumhdes@gmail.com' => ['Pathum Harshana De Silva', 'pathumhdes@gmail.com'],
+        'andre@andre.(none)' => ['Carlos André Ferrari', 'caferrari@gmail.com'],
+        'mazanax@yandex.ru' => ['Aleksandr Gorshkov', 'mazanax@yandex.ru'],
+        'paulkarikari1@gmail.com' => ['Paul Karikari', 'paulkarikari1@gmail.com'],
+    ];
```
alganet, line 33: I appreciate that you made a revised script to re-assign FileContributors.

However, I don't believe this list should be commited and part of that script.

I think it's a one time thing that we need to do once (or perhaps twice, if we notice an error). Then, we can perform much lighter checks that do not even depend on git.

src-dev/Commands/LintSpdxCommand.php
henriquemoody: I was a bit sad as well, you got a point. I think there is a fundamental change in which the command is more of a fixer, to it builds more than validates. But there are indeed several places that we should use

src-dev/Spdx/Contributor.php
henriquemoody: The idea is to automatically update the header, hence I thought it would make sense to keep the list here, but it bothers me.

I would say though that we ourselves have used this list a couple of times already, so it doesn't seem to weird to keep it

henriquemoody:
I got this in a couple iterations with an AI because I saw myself having to add the header to files I forgot. I think this is more useful for us maintainers than non-frequent contributors

henriquemoody:
I will merge it as is, refactor later, I muyself don't want to spend too much time on it, it's only worth it if it's cheap, but the part of using Validation is a strong argument, we should probably check all the code in src-dev/

PR #1675, opened 2026-02-04
Update translation docs

alganet:
The documents on translation were updated to feature symfony with
an array provider. Duplicated container notes were extracted to
a single configuration.md file.

An API for accessing the messages, so users don't have to copy
and paste them from the source or docs, was provided and
TemplateResolver was refactored to use it.

### DIFF
--- a/src/Message/TemplateRegistry.php
+++ b/src/Message/TemplateRegistry.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * SPDX-License-Identifier: MIT
+ * SPDX-FileCopyrightText: (c) Respect Project Contributors
+ * SPDX-FileContributor: Alexandre Gomes Gaigalas <alganet@gmail.com>
+ */
+
+declare(strict_types=1);
+
+namespace Respect\Validation\Message;
+
+use InvalidArgumentException;
+use ReflectionClass;
+use Respect\Validation\Validator;
+
+use function sprintf;
+
+final class TemplateRegistry
+{
+    /** @var array<class-string<Validator>, array<Template>> */
+    private array $templates = [];
+
+    /**
+     * @param class-string<Validator> $validatorClass
+     *
+     * @return array<Template>
+     */
+    public function getTemplates(string $validatorClass): array
+    {
+        if (!isset($this->templates[$validatorClass])) {
+            $reflection = new ReflectionClass($validatorClass);
+            foreach ($reflection->getAttributes(Template::class) as $attribute) {
+                $this->templates[$validatorClass][] = $attribute->newInstance();
+            }
+        }
+
+        return $this->templates[$validatorClass] ?? [];
+    }
+
+    /** @param class-string<Validator> $validatorClass */
+    public function get(string $validatorClass, string $id = Validator::TEMPLATE_STANDARD): Template
+    {
+        foreach ($this->getTemplates($validatorClass) as $template) {
+            if ($template->id === $id) {
+                return $template;
+            }
+        }
+
+        throw new InvalidArgumentException(sprintf(
+            'Template with id "%s" not found in validator "%s".',
+            $id,
+            $validatorClass,
+        ));
+    }
+}
--- a/tests/feature/TranslatorTest.php
+++ b/tests/feature/TranslatorTest.php
@@ -10,30 +10,38 @@
 declare(strict_types=1);
 
 use Respect\Validation\ContainerRegistry;
+use Respect\Validation\Message\TemplateRegistry;
+use Respect\Validation\Validators as vs;
 use Symfony\Component\Translation\Loader\ArrayLoader;
 use Symfony\Component\Translation\Translator;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
-$translator = new Translator('en');
+$templates = new TemplateRegistry();
+$translator = new Translator('pt_BR');
 $translator->addLoader('array', new ArrayLoader());
 $translator->addResource(
     'array',
     [
-        '{{subject}} must pass all the rules' => 'Todas as regras requeridas devem passar para {{subject}}',
-        'The length of' => 'O comprimento de',
-        '{{subject}} must be a string' => '{{subject}} deve ser uma string',
-        '{{subject}} must be between {{minValue}} and {{maxValue}}' => '{{subject}} deve possuir de {{minValue}} a {{maxValue}} caracteres',
-        '{{subject}} must be a phone number for country {{countryName|trans}}' => '{{subject}} deve ser um número de telefone válido para o país {{countryName|trans}}',
+        // Directly translating validator messages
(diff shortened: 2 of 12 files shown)


src/Message/Template.php
```diff
@@ -23,4 +29,33 @@ public function __construct(
         public string $id = Validator::TEMPLATE_STANDARD,
     ) {
     }
+
+    /** @param class-string $validatorClass */
+    public static function from(string $validatorClass, string $id = Validator::TEMPLATE_STANDARD): self
```
henriquemoody: I see the appeal to add this here, but I think this deserves it's own class. I would also prefer that we do that in just one place. The [TemplateResolver](https://github.com/Respect/Validation/blob/main/src/Message/Formatter/TemplateResolver.php#L72-L81) does something similar to that, but it also keep an internal state so we don't have to do introspection every time.

For the use-case you're describing here, I think having a `TemplateExtractor` or something similar would work best, and we could use that extractor on `TemplateResolver` somehow

tests/feature/TranslatorTest.php
```diff
@@ -10,30 +10,37 @@
...
 
 use Respect\Validation\ContainerRegistry;
+use Respect\Validation\Message\Template;
+use Respect\Validation\Validators as vs;
 use Symfony\Component\Translation\Loader\ArrayLoader;
 use Symfony\Component\Translation\Translator;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
-$translator = new Translator('en');
+$translator = new Translator('pt_BR');
```
henriquemoody, line 20: nice catch

src/Message/Template.php
alganet: The name `TemplateExtractor` doesn't express the intent of keeping a global store of templates, it just says it extracts them. We know that each validator has a global list of templates, but a developer encountering the name `TemplateExtractor` might not have that knowledge.

Instead, I went with `TemplateRegistry` and also registered it in the `ContainerRegistry` and updated the docs. Now `TemplateResolver` uses it, and the examples use `TemplateRegistry` directly to access the global store of templates.

There are further internal refactors to `TemplateResolver` that can be done now, but I think that is too much scope for this PR. I will probably follow up on them later, since they don't change the public API.

src/Message/TemplateRegistry.php
```diff
@@ -0,0 +1,62 @@
...
+use Respect\Validation\Validator;
+
+use function sprintf;
+
+final class TemplateRegistry
+{
+    /** @var array<class-string<Validator>, array<Template>> */
+    private array $templates = [];
+
+    public static function getInstance(): self
```
henriquemoody: I can't see the reason why this method needs to exist. Can you elaborate on your reasoning?

src/Message/Template.php
henriquemoody: `TemplateRegistry` is a better name, and I think `TemplateFactory` could also work, both names are better but I see a reason to support and question both choices.

src/Message/TemplateRegistry.php
alganet: It provides an API for getting the instance decoupled from the `ContainterRegistry`. It uses the container inside, but the user doesn't need to know about it. It's good for those scenarios (when knowledge of the container is not needed).

For example, if the user wants to reuse existing templates like this: `v::templated(TemplateRegistry::getInstance()->get(IntVal::class), v::callback(...` or in the `assert` replacements.

alganet: In fact, I still prefer having a `Template::from` for those kinds of scenarios. Can I move it back to the `Template` class just as a façade for the `TemplateRegistry` implementation?

henriquemoody: If that's the use case, I prefer `Template::from` as well. But then, I don't see a reason to have the `TemplateRegistry`.


I also think that if we want to do things like this `v::templated(TemplateRegistry::getInstance()->get(IntVal::class), v::callback(...`, we should provide a cleaner API for it, but I'm not sure I want to go there, if I'm honest.

I would keep the `TemplateRegistry`, as the in-memory caching layer is useful, but I would remove the `getInstance()` method. I think we should keep the cases in which we use the `ContainterRegistry` as minimum. When I created it, my goal was to enable it to be used just because we have a `__callStatic()` in the `ValidatorBuilder`, we should favor dependency injection in other parts of the library, IMO.

alganet: The use case is:

> Q: As a user, how do I get a template for a given validator programaticallly?

It's a reasonable thing to ask. The answer is going to be something like this:

```
$container = ContainerRegistry::getContainer();
$templateRegistry = $container->get(TemplateRegistry::class);
$template = $templateRegistry->get(IntVal::class);
```

I think that answer sucks.

This answer would be better:

```
$template = Template::from(IntVal::class);
```

There's nothing preventing us from using all the cool tricks inside (a global registry, clean dependencies injected, etc). It doesn't prevent the user from providing his own `TemplateRegistry` (injecting dependencies), it just makes it easier to access the default one.

We never **actually need** the syntatic sugar, ever, right? The user can always deal with the raw building blocks directly. However, it's nice to have them.

henriquemoody: > Q: As a user, how do I get a template for a given validator programmatically?

I think different use cases might call different solutions.

> As an user, I want to have access to all the translations from Validation so I can translate them.

It's much easier if we provide them with a collection of messages we already have. It's static, there's no need for the user to get them dynamically for that scenario. But if that's the case, the `TemplateRegistry` works great, even better than the `Template::from()`, IMO.

> As an user, I want to set the template of my custom validator as an existing validator.

That calls for something dynamic, but to me, something like `v::templated(Template::from(IntVal::class)->default, v::callback(...` sucks. If we want to provide the users with something like that, we could create a validator that does that, and then uses the `TemplateRegistry/Factory` as a dependency.

alganet: I removed the `getInstance()`.

IMHO, our docs now are worse than the initial version I presented. We could totally have both (a beginner introduction with easier APIs and a more advanced section with `TemplateRegistry`).

In the end, I think we're shifting the complexity to the users. Look at how many distractions there are in the task of setting up translations. Façades like `Template::from` easy that burden while not invalidating under-the-hood implementations like `TemplateRegistry`.

henriquemoody: We could have `Template::from()` using `TemplateRegistry` via `ContainerRegistry`, I'd be fine with that as well, but I don't think need the `TemplateRegistry::getInstance()`

alganet: That is precisely what I was suggesting [here](https://github.com/Respect/Validation/pull/1675#discussion_r2773766777) 😆 

> alganet said: In fact, I still prefer having a Template::from for those kinds of scenarios. Can I move it back to the Template class just as a façade for the TemplateRegistry implementation?

henriquemoody: Dude, I'm so sorry... You said exactly that 😕

alganet: No problem! I'll leave that improvement for another day.

PR #1692, opened 2026-02-10
 Alias latest branch to dev-main 

alganet:
This makes having a specific branch for the latest version
unecessary, allowing us to work on `main` directly.

The only necessary branches now are for maintained versions
of earlier releases (right now, 2.3 and 2.4).

For released versions, the tags are used instead (3.0.0,
3.0.1 and so on).

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ SPDX-License-Identifier: MIT
 -->
 # Respect\Validation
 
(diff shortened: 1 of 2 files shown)


henriquemoody:
There are a few things in this PR that we should keep, could you reopen with only the URL fixes against 3.0?

PR #1702, opened 2026-02-20
Introduce Trimmed validator

alganet:
This commit introduces the `Trimmed` validator that ensures a string cannot start or end with a list of specific values.

The default values used are a selected list of Unicode invisible characters.

To support this change, the StartsWith and EndsWith validators were modified so they can also support multiple values to check for.

While StartsWith and EndsWith are more generic, and also perform start-of-array and end-of-array kinds of checks, Trimmed is more focused on string inputs, which tailors to a more specific use case.

### DIFF
--- a/src/Validators/Trimmed.php
+++ b/src/Validators/Trimmed.php
@@ -1,0 +1,87 @@
+<?php
+
+/*
+ * SPDX-License-Identifier: MIT
+ * SPDX-FileCopyrightText: (c) Respect Project Contributors
+ * SPDX-FileContributor: Alexandre Gomes Gaigalas <alganet@gmail.com>
+ */
+
+declare(strict_types=1);
+
+namespace Respect\Validation\Validators;
+
+use Attribute;
+use Respect\Validation\Message\Template;
+use Respect\Validation\Validator;
+use Respect\Validation\Validators\Core\Envelope;
+
+#[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
+#[Template(
+    '{{subject}} must not contain leading or trailing whitespace',
+    '{{subject}} must contain leading or trailing whitespace',
+    Validator::TEMPLATE_STANDARD,
+)]
+#[Template(
+    '{{subject}} must not contain leading or trailing {{trimValues|list:or}}',
+    '{{subject}} must contain leading or trailing {{trimValues|list:or}}',
+    self::TEMPLATE_CUSTOM,
+)]
+final class Trimmed extends Envelope
+{
+    public const string TEMPLATE_CUSTOM = '__custom__';
+
+    /** Unicode whitespace and zero-width characters. */
+    private const array DEFAULT_TRIM_VALUES = [
+        "\u{0009}", // CHARACTER TABULATION
+        "\u{000A}", // LINE FEED
+        "\u{000B}", // LINE TABULATION
+        "\u{000C}", // FORM FEED
+        "\u{000D}", // CARRIAGE RETURN
+        "\u{0020}", // SPACE
+        "\u{0085}", // NEXT LINE
+        "\u{00A0}", // NO-BREAK SPACE
+        "\u{1680}", // OGHAM SPACE MARK
+        "\u{180E}", // MONGOLIAN VOWEL SEPARATOR
+        "\u{2000}", // EN QUAD
+        "\u{2001}", // EM QUAD
+        "\u{2002}", // EN SPACE
+        "\u{2003}", // EM SPACE
+        "\u{2004}", // THREE-PER-EM SPACE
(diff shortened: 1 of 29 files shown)


src/Validators/Trimmed.php
```diff
@@ -0,0 +1,103 @@
...
+            new Not(
+                new AnyOf(
+                    new StartsWith(...$trimValues),
+                    new EndsWith(...$trimValues),
+                ),
+            ),
+        );
+    }
+
+    public function evaluate(mixed $input): Result
```
henriquemoody: Couldn't you just extend `Envelop` instead?

alganet:
@henriquemoody 

I thought about the `$side` parameter, but that is essentially what `notStartsWith` and `notEndsWith` are. That was my rationale for adding multiple alternatives instead of a new validator with `$side`, it felt more natural.

For the `StartsWith` and `EndsWith` multiple parameters, there's a big difference versus oneOf/anyOf: the message. It says `Must start with "Mr." or "Ms."` (for example) instead of a more verbose "One of the rules: - Must start with "Mr" - Must start with "Ms".

I will work on those scenarios to make them more evident. The lorem ipsum is not making justice to the change!

Regarding `mb_trim`, it is certainly a possibility. We're doing more than trimming single characters. With the current setup, we can also check for whole strings: `v::trimmed('<!--', '-->', '#', '//')` and this is not possible with `mb_trim`. I agree though that it might be a niche case. I wouldn't mind simplifying it.

What set of templates would you use for a `v::trimmed($side, ...$chars)`? Would you use an Enum for the side or plain string?

henriquemoody:
I see. When I read `v::trimmed('<!--', '-->', '#', '//')` it seemed like another validator, something like a `EdgesWith`, `WrapsWith`. Perhaps we're just a bit stuck to "trim" because of familiarity, but in fact, we only care about the "trimming" when that validator handles spaces. 

I think we could still have `Trimmed`, but be that about those trim values. I'm not sure.

> What set of templates would you use for a v::trimmed($side, ...$chars)? Would you use an Enum for the side or plain string?

After your last comment, I'm reconsidering the $side, because people could just use `StartsWith` and `EndsWith`. I think the message should be more geared towards being wrapped by a value, something like:

```
{{subject}} must start be wrapped by {{startValue}}
```

I have to think a bit more about it, though

alganet:
@henriquemoody 

The names `EdgesWith` and `WrapsWith` are the *opposite* case, right? Trimmed ensures something *does not* have things at the edge.

This one is hard to "unreverse". One could argue that in this case, something like `v::not(v::surroundedWith('...` would apply. It makes sense when you read it: "oh, not surrounded with is the same as trimmed", but what does `surroundedWith` without the `not` does?

Here, let me use examples:

```php
// A surroundedWith designed to be used with "not" for the case of trimming:
v::surroundedWith('{{', '}}')->assert('{{foo}}'); // this would pass
v::surroundedWith('{{', '}}')->assert('{{foo'); // this should ALSO pass if we want not() to behave as trimmed
```

In reality, the standalone non-negated `surroundedWith` should use an internal `AND` (has stuff at the beginning and end), while the check trimmed operation is an `OR`.

The "un-negated" version of the hypothetical validator that, if negated would result in a trim, would be something like `v::startsOrEndsWith` (or something like that).

I have also considered several other names: `v::delimitedWith` (implies something is "delimited" by a common marker at the beginning OR-non-exclusive ending.

In the end, `trimmed` still sounds to me like the most familiar one.

I've updated the PR with real use cases (such as ensuring a name doesn't have a title):

```php
v::trimmed('Dr.', 'Mr.', ', PhD')->assert('John Doe, PhD');
// → "John Doe, PhD" must not contain leading or trailing "Dr.", "Mr.", or ", PhD"
```

I also changed the message to "must not contain leading or trailing", which sounds more natural.

Let me know if you have better ideas or examples!

henriquemoody:
Oh, I think it got merged in the wrong branch, though, it should have been `main`. The `3.0` is only for patch versions

alganet:
@henriquemoody aren't we getting rid of `main`?

We could rename 3.0 to 3.1, then open a new 3.0 up to the previous sha.

henriquemoody:
> aren't we getting rid of `main`?

I don't know... My thought process is to keep the `main` and I accept new features into it. Depending on the changes we can either decide to create a MAJOR or MINOR branch. We could also, just create a MAJOR and MINOR, and get rid of `main`

> We could rename 3.0 to 3.1, then open a new 3.0 up to the previous sha.

That works! It's the same as creating a 3.1 and reverting the commits from 3.0. Whatever is easier.

alganet:
@henriquemoody I renamed the branches.

We can keep using `main`, making just a very short temporary exception for this 3.1 release.

On monday, I'll release 3.0.2 and 3.1.0, then we'll be back to having `main` as the next minor instead of `3.1`.

PR #1703, opened 2026-02-22
Fix PublicSuffix validator and UpdateDomainSuffixesCommand

alganet:
- Parce PSL ICANN section into structured sections (rules,
   wildcards, exceptions) according to the format.
 - Updates PublicSuffix semantics for complete application of
   the rules.
 - Includes private domain suffixes now.
 - Refreshes the existing data.
 - Fixes the update-regionals.yml workflow, set it to run
   twice a week.

References: https://github.com/publicsuffix/list/wiki/Format#format


alganet:
Copilot said it couldn't review, but produced a review nonetheless. I inspected the agent session and fixed the comments it has generated. I'll try requesting the review again.

alganet:
I am fully aware that this validator got "too big" now, and a separate repository containing the data would be better. However, that is also a tricky move (changing dependencies, making the DataLoader cross vendor or designing a standalone interface for the data).

I decided to just deliver it, since the bulk of it is easily removable (data folder) if we want to change course later.

henriquemoody (approved):

PR #1709, opened 2026-03-04
Update main branch to latest minor release

alganet:
Done via PR to comply with branch protection rules.


alganet:
This should probably be an automatic workflow in the future.

alganet:
@henriquemoody I'm trying to keep the branches protected (no force pushes, no linear history, no changes without a PR), so we might need them for now.

Please check #1710 for the automated solution that would make these kinds of temporary PRs unecessary.

Issue #1710, opened 2026-03-04
RFC: Automatic Branching Model


alganet:
# RFC: Automatic Branching Model

## Summary

This RFC proposes a branching model and set of GitHub Actions workflows that
automate version management, forward-porting, backporting, and releases for
the Respect/Validation library. The goal is to keep multiple minor versions
maintainable at once with minimal manual effort.

## Motivation

The project already maintains long-lived branches for each minor version
(`3.0`, `3.1`, …) alongside `main`. Today, propagating fixes across versions,
creating releases, and keeping `main` in sync are all manual tasks. As the
number of maintained versions grows, this becomes error-prone and
time-consuming.

An automated branching model would:

- Let contributors target a single branch without worrying about propagation.
- Let maintainers backport and forward-port with a label click.
- Keep `main` automatically in sync with the latest released version.
- Make patch releases a one-click operation.
- Enforce a clear, documented lifecycle for every version branch.

## Branching Layout

```
main  ← always mirrors the latest version branch (read-only sync)
 │
 ├── 3.2  ← latest version branch (active development)
 ├── 3.1  ← previous version (maintained, receives backports)
 ├── 3.0  ← older version (maintained, receives backports)
 └── 2.4  ← legacy version (security fixes only)
```

### Branch roles

| Branch          | Purpose                              | Accepts PRs | Automated |
|-----------------|--------------------------------------|:-----------:|:---------:|
| `main`          | Mirror of latest version branch      | No          | Yes       |
| `X.Y` (latest)  | Active development for next patch    | Yes         | —         |
| `X.Y` (older)   | Maintained version, backports only   | Via bot     | Yes       |
| `X.Y` (EOL)     | No longer maintained                 | No          | —         |

### Rules

1. **`main` is read-only.** It is force-updated automatically whenever
   the latest version branch receives a push.
2. **All PRs target a version branch**, never `main`.
3. **Contributors target the oldest applicable branch.** A bug fix that affects
   `3.0`, `3.1`, and `3.2` should be opened against `3.0`. A new feature
   should target the latest version branch.
4. **Forward-porting is automatic.** After a commit lands on an older branch,
   the change is cherry-picked upward to every newer maintained branch.
5. **Backporting is label-driven.** When a PR merged into a newer branch also
   needs to land on an older one, a maintainer adds a `backport:X.Y` label
   and the bot creates a cherry-pick PR.
6. **Rebase only.** All branches maintain a linear history. PRs are integrated
[...]

Issue #1681, opened 2026-02-08
Ship a PHPStan extension for type narrowing via assert()/check()


henriquemoody:
After calling `assert()` or `check()` on a validation chain, PHPStan should be able to narrow the type of the asserted value. Currently, PHPStan cannot do this because the method chain goes through `@mixin` + `__callStatic`/`__call`, which prevents PHPStan from tracking type information across the call chain.

Shipping a bundled `MethodTypeSpecifyingExtension` (auto-discovered via `phpstan/extension-installer`) would let PHPStan understand that passing validation implies a type guarantee — the same way `phpstan-phpunit` handles `assertInstanceOf()`, `assertIsString()`, etc.

## Type narrowing by prefix

Validation has a consistent method naming pattern with prefixes (`all*`, `nullOr*`, `not*`, `key*`, `property*`) applied to base validators. Each prefix modifies the semantics and could produce different PHPStan type narrowing:

### Base type validators

The core `*Type()` and `instance()` methods narrow to a single type:

```php
ValidatorBuilder::stringType()->assert($input);  // $input is string
ValidatorBuilder::arrayType()->assert($input);   // $input is array
ValidatorBuilder::objectType()->assert($input);  // $input is object
ValidatorBuilder::instance(Foo::class)->assert($input); // $input is Foo
```

Applies to: `arrayType`, `boolType`, `callableType`, `floatType`, `intType`, `iterableType`, `nullType`, `objectType`, `resourceType`, `stringType`, `instance`.

### `all*` prefix — array element narrowing

Narrows the input to `array<T>`:

```php
ValidatorBuilder::allStringType()->assert($input);            // $input is array<string>
ValidatorBuilder::allInstance(Foo::class)->assert($input);     // $input is array<Foo>
ValidatorBuilder::allBoolType()->check($input);                // $input is array<bool>
```

### `nullOr*` prefix — nullable narrowing

Narrows the input to `T|null`:

```php
ValidatorBuilder::nullOrStringType()->assert($input);          // $input is string|null
ValidatorBuilder::nullOrInstance(Foo::class)->assert($input);   // $input is Foo|null
ValidatorBuilder::nullOrObjectType()->assert($input);           // $input is object|null
```

### `not*` prefix — negative narrowing

Excludes a type from the input (useful when the input is already a union):

```php
/** @param int|string $input */
function example(int|string $input): string {
    ValidatorBuilder::notIntType()->assert($input);
    return $input; // $input is string
}
```

### Chained calls

[...]

PR #1729, opened 2026-03-23
Integrate Respect/Fluent and Respect/FluentGen

alganet:
Replace the in-house factory and codegen infrastructure with Fluent and FluentGen packages. The runtime now uses ComposingLookup with ComposableMap for prefix resolution via a FluentValidatorFactory adapter that preserves exception types. Validators use #[Composable] from [Fluent ](https://github.com/Respect/Fluent)instead of the custom #[Mixin] attribute. Mixin generation uses [FluentGen](https://github.com/Respect/FluentGen), producing PrefixConstants in place of PrefixMap. The old src-dev/CodeGen directory is removed. All public APIs and BC are preserved.

---

This PR is currently molded as a 3.x release, but *we could push the integration further* and remove more code by integrating deeper with Respect\Fluent. However, that breaks BC, and it would be a 4.x release instead.

That BC break would be kind of "soft", in the sense that validators keep working. It would only affect really advanced users that have their own ValidatorFactory implementations. This is related to #1710 somehow: if we had automated backports, a 4.x release would make more sense (easier to keep both 3.x and this 4.x iteration, since the bulk of validators don't change, just some wiring on fluent stuff).

---

Integration with Respect/FluentAnalysis is not yet in this PR, but its use should be possible with it with some boilerplate configuration.

Those three together unlock **reusable, declarative advanced fluent composition, type inference and IDE support** with extensibility. For example, Validation extensions with custom namespaces could ship `#[Composable]` annotations and end users would get IDE support for custom and extended rules too (requires some wiring on composer scripts).

---

This PR introduces two new projects. I also welcome reviews on them as issues on their specific repositories, linked to this PR number so we can track them as a group.


src/ContainerRegistry.php
```diff
@@ -68,9 +70,18 @@ public static function createContainer(array $definitions = []): Container
             'respect.validation.formatter.messages' => autowire(NestedArrayFormatter::class),
             'respect.validation.ignored_backtrace_paths' => [__DIR__ . '/ValidatorBuilder.php'],
             'respect.validation.rule_factory.namespaces' => ['Respect\\Validation\\Validators'],
-            ValidatorFactory::class => factory(static fn(Container $container) => new NamespacedValidatorFactory(
-                $container->get(Transformer::class),
-                $container->get('respect.validation.rule_factory.namespaces'),
+            ValidatorFactory::class => factory(static fn(Container $container) => new FluentValidatorFactory(
```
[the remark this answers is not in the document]
alganet: Transformers are an internal interface, undocumented and considered to be removed in previous PRs.

src/Mixins/PrefixConstants.php
```diff
@@ -0,0 +1,56 @@
...
+        'nullOr' => true,
+        'allOf' => true,
+        'all' => true,
+        'key' => true,
+        'max' => true,
+        'min' => true,
+        'not' => true,
+    ];
+    public const array COMPOSABLE_WITH_ARGUMENT = ['key' => true, 'property' => true];
+    public const array FORBIDDEN = [
```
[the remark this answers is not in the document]
alganet: It should have had an `@internal` marker, but we often don't use those. This is a valid concern, and marking internal stuff should become an issue at some point. I'll consider opening a ticket for it.

src/ContainerRegistry.php
```diff
@@ -68,9 +70,18 @@ public static function createContainer(array $definitions = []): Container
...
+                        new Ucfirst(),
+                        Validator::class,
+                        ...$container->get('respect.validation.rule_factory.namespaces'),
+                    ),
+                    new ComposableMap(
+                        PrefixConstants::COMPOSABLE,
+                        PrefixConstants::COMPOSABLE_WITH_ARGUMENT,
+                    ),
+                ),
             )),
```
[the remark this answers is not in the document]
alganet: Trimming the namespace actually hurts future plans for Fluent/FluentGen (ability to compose namespace prefixes) and it's a waste of CPU cycles.

This behavior should be deprecated, but you're right for now!

composer.json
```diff
@@ -48,6 +49,7 @@
         "sokil/php-isocodes-db-only": "^4.0",
         "squizlabs/php_codesniffer": "^4.0",
         "symfony/console": "^7.4 || ^8.0",
+        "respect/fluentgen": "^1.0",
         "symfony/translation": "^8.0",
         "symfony/var-exporter": "^6.4 || ^7.0 || ^8.0"
```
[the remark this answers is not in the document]
alganet: Sorted now.

src/FluentValidatorFactory.php
```diff
@@ -0,0 +1,53 @@
...
+    /** @param array<int, mixed> $arguments */
+    public function create(string $ruleName, array $arguments = []): Validator
+    {
+        try {
+            $instance = $this->factory->create($ruleName, $arguments);
+        } catch (CouldNotResolve) {
+            throw new ComponentException(sprintf('"%s" is not a valid rule name', $ruleName));
+        } catch (CouldNotCreate | Error $e) {
+            throw new InvalidClassException($e->getMessage());
+        }
```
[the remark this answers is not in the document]
alganet: Fixed. Now they bubble up.

alganet:
@TheRespectPanda benchmark

TheRespectPanda:
Triggered phpbench benchmarks for PR #1729 — workflow run: https://github.com/Respect/Validation/actions/workflows/ci-perf.yml

alganet:
Some updates and issues I see with the current design:

 - `Composable` could reuse `class-string` entries instead of plain `string`. 
 - `Composable` parameter targetting would be better off if we split the annotation into a TARGET_CLASS attribute and a TARGET_PARAMETER attribute. 
 - `FluentAssertion` is likely better to be `AssuranceAssertion` (following an upcoming FluentAnalysis version), and to be targetted on individual methods instead of TARGET_CLASS.

Basically, removing all references we have for parameters and methods that are stored in strings like `'methodName'` or `'paramName'`. Illustrative example:

```php
// Before:
#[Composable(prefixParameter: 'foo')]
class Instance {
    public function __construct(private string $foo) {}
}

// After:
#[Composable]
class Instance {
    public function __construct(
        #[ComposableParameter] private string $foo,
    ) {}
}
```

For future type narrowing (as described in #1681), I just can't quite get a good design for the type algebra parameters. It's likely they will be strings with phpdoc union types that are parsed by the extension.

alganet:
I addressed the concerns in my previous comment. Much cleaner `#[Composable(with: [Length::class, Max::class, Min::class])]` style attributes, more modular attributes (ValidationBuilder methods now addresses by `#[AssuranceAssertion]` directly.

I also went full static analysis on a separate PR #1730 that takes this approach even further and it in draft. ~The line count looks bigger, but it's mostly the `fluent.neon` phpstan configuration file that has all the type inference and narrow information compiled (2KLOC)~ (I was able to reduce it actually). That PR also breaks BC (also migrates ValidationBuilder, requires no factory), while this one does not.

alganet:
I was able to integrate more of my local changes into #1730.

I'm still not satisfied with some things, like:

```php
#[Assurance(type: 'array|Countable')]
final class Countable extends Simple
```

I can't find a good way of expressing those union types for phpstan narrowing. 

The obvious:

```php
#[Assurance(type: ['array', Countable::class])]
final class Countable extends Simple
```

... is not so good future-proofing for more complex expressions. We can't assume it will always be unions.

src/Mixins/Builder.php
```diff
@@ -182,11 +182,11 @@ public static function iterableVal(): Chain;
 
     public static function json(): Chain;
 
-    public static function key(string|int $key, Validator $validator): Chain;
+    public static function key(int|string $key, Validator $validator): Chain;
```
henriquemoody, line 187: Not important, but do you know why this happened?

src/ContainerRegistry.php
```diff
@@ -68,10 +72,26 @@ public static function createContainer(array $definitions = []): Container
             'respect.validation.formatter.messages' => autowire(NestedArrayFormatter::class),
             'respect.validation.ignored_backtrace_paths' => [__DIR__ . '/ValidatorBuilder.php'],
             'respect.validation.rule_factory.namespaces' => ['Respect\\Validation\\Validators'],
-            ValidatorFactory::class => factory(static fn(Container $container) => new NamespacedValidatorFactory(
-                $container->get(Transformer::class),
-                $container->get('respect.validation.rule_factory.namespaces'),
-            )),
+            ValidatorFactory::class => factory(static function (Container $container) {
+                $namespaces = array_map(
+                    static fn($ns) => trim($ns, '\\'),
```
henriquemoody: I think that could just be in the `NamespaceLookup`. If `NamespaceLookup` was called `PrefixLookup` it would make more sense to keep it here

src/ValidatorBuilder.php
```diff
@@ -31,6 +39,10 @@
 use function is_string;
 
 /** @mixin Builder */
+#[FluentNamespace(new ComposingLookup(
+    new NamespaceLookup(new Ucfirst(), Validator::class, 'Respect\\Validation\\Validators'),
+    new ComposableMap(PrefixConstants::COMPOSABLE, PrefixConstants::COMPOSABLE_WITH_ARGUMENT),
```
henriquemoody: A pity that we can't infer that from the annotations in the classes. I reckon it's because it's unviable to load all the classes in a namespace to identify which classes provide prefixes or not. I liked your idea of generating a class like `PrefixConstants`! Passing an object to `ComposableMap` from an object instead? And why not passing the `FORBIDDEN` constant?

henriquemoody (approved):
This is absolutely awesome!

I haven't read the code from Fluent and FluentGen yet, but it seems really great!

src/ContainerRegistry.php
alganet: Copilot also flagged this. My goal is to remove it (Fluent would not trim it, it's up to the user to provide a trimmed config).

src/ValidatorBuilder.php
alganet: Yep, scanning the vendor folder is bad (it did at some point, then I realized that in large projects it would take too much time).

---

Passing an object (PrefixConstants) would mean one more interface, and coupling between FluentGen and FluentAnalysis.

FluentGen: generates mixins (and the PrefixConstants map too!)
FluentAnalysis: phpstan static analysis

Currently, both depend on Fluent, but they don't depend on each other. You can in fact generate mixins for classes that do not have `#[FluentNamespace]`, just like previous incubated CodeGen did, and use the factory outside the attribute too.

So, it's my way of keeping it decoupled. FluentAnalysis provides a `ComposableMap` that accepts bare arrays. It doesn't care how you generate them. Fluent also provides a `ComposableAttributes` version that does it without the map (albeit slower).

---

Good catch on the `FORBIDDEN`!

src/Mixins/Builder.php
alganet: I probably over-optimized something on FluentGen and behavior changed. I could probably follow the original behavior to make the diffs smaller.

henriquemoody: I was just wondering why. This is better, though, because at least it's following the same contract as `Key` now

src/ContainerRegistry.php
henriquemoody: > Fluent would not trim it, it's up to the user to provide a trimmed config

I think it's best for Fluent to sanitize it, though, unless it doesn't need (I don't know the implementation) or unless it's a _prefix_ rather than a namespace. But again, I don't know the implementation, so feel free to ignore this

src/ValidatorBuilder.php
henriquemoody: If the interface belongs to `Fluent` that would be alright, wouldn't it? Since FluentGen depends on Fluent already

alganet: > Since FluentGen depends on Fluent already

In `require-dev` though. You could totally use FluentGen to generate mixins, or FluentAnalysis for phpstan, but never ship `respect/fluent` in the main `require` group! Fluent is 90% inert attributes, 10% stuff extracted from Validation.

I decided to do it (ship `respect/fluent`) in Validation because it fits (decouples the validation+reporting engine from the fluent interface), but other projects might decide they just want the attributes in `require-dev` time.

alganet:
I decided I'll work more on Fluent, FluentGen and FluentAnalysis and first integrate them into Data+Relational ($mapper->table auto-complete, syntax analysis), Rest (routine composition, static analysis that potentially understand basic HTTP semantics) and Config (container auto-complete).

Those projects are in incubation mode and I can iterate much faster on them, which presents more dogfooding opportunities. Once I'm happy with the abstractions there, I'll come back to Validation.

henriquemoody:
Hey, @alganet ! Do you have any plans of getting back to this? I'm considering upgrading Assertion and this would be incredibly useful!

alganet:
@henriquemoody I do have plans for it! 

Is this the PR you actually need? (this replaces stub generation, mostly).

The one with narrowing is #1730, related to #1681. That one needs more work (and minor releases of Fluent and FluentAnalysis).

We can chose either one or the other, #1730 obsoletes this one, but it's also larger and less incremental. I'm fine with either sequence (#1729 first, then a revised #1730; or #1730 directly).

Let me know, and I'll update a new amend for review in the one you actually need right now.

henriquemoody:
In fact, #1730 is really what I need, even though I suspect there would be a some work to make it work with assertion.

My goal is allowing this:

```php
Assertion::intType($var);

echo substr('String', $var);
```

not to fail because $var is not an integer, but also allow IDEs to identify those as much possible, which probably means reading the attributes and translating them into docblocks with `@phpstan-assert`.

No rush, though, it might take some time for me to get deep into that.

alganet:
The way `FluentAnalysis` does is without `@phpstan-assert` though. It registers an extension to do things that docblocks cannot cleanly do (which happen a lot on Validation). It works only for phpstan for now, I have not tested it in IDEs (and I suspect it doesn't work).

The division is:

 - Fluent: core reusable DSL runtimes ("make your own chain like Validation") and chain-enriching annotations.
 - FluentGen: generates stubs. This one works with IDEs (which are often stub based). It does what you said (read attributes and generate shit).
 - FluentAnalysis: static analysis extensions, for phpstan (now), maybe psalm later (haven't investigated it yet).

If we want stubs with `@phpstan-assert` and `@phpstan-assert-if-true` (which very likely can coexist with extensions but less rich), `FluentGen` would be the place to do it.

Respect\Assertion would probably need nothing for case 1 (the substr not failing phpstan) if the extension works as intended. The information is ingested by phpstan in runtime and it knows Assertion is an adapter for Validation (which carries semantics of the types). I haven't tested case 2 (the IDE itself knowing this deeply).

src-dev/Commands/LintMixinCommand.php
```diff
@@ -89,9 +93,10 @@ interfaces: [
...
 
-        $prefixMapGenerator = new PrefixMapGenerator(
+        $prefixMapGenerator = new PrefixConstantsGenerator(
             config: $config,
-            outputClassName: 'PrefixMap',
+            scanner: $scanner,
+            outputClassName: 'PrefixConstants',
         );
 
         $files = $generator->generate() + $prefixMapGenerator->generate();
```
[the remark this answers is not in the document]
alganet: What? `$prefixConstantsGenerator` does not exist.

tests/unit/FluentValidatorFactoryTest.php
```diff
@@ -0,0 +1,127 @@
...
+    #[Test]
+    public function itShouldBubbleUpErrorWhenInstantiationFails(): void
+    {
+        $factory = new FluentValidatorFactory(
+            new NamespaceLookup(new Ucfirst(), null, self::TEST_NAMESPACE),
+        );
+
+        $this->expectException(Error::class);
+
+        $factory->create('myAbstractClass');
```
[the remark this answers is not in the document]
alganet: Solved.

src-dev/Commands/LintMixinCommand.php
```diff
@@ -67,8 +68,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             outputNamespace: 'Respect\\Validation\\Mixins',
         );
 
+        $scanner = new NamespaceScanner();
+
         $generator = new MixinGenerator(
             config: $config,
+            scanner: $scanner,
             methodBuilder: new MethodBuilder(
                 excludedTypePrefixes: ['Sokil', 'Egulias'],
```
henriquemoody, line 77: Couldn't those be also be annotations in the the validators?

That would be useful also to generate the Markdown if we want to omit some arguments there as well.

src/ContainerRegistry.php
```diff
@@ -68,10 +70,21 @@ public static function createContainer(array $definitions = []): Container
             'respect.validation.formatter.messages' => autowire(NestedArrayFormatter::class),
             'respect.validation.ignored_backtrace_paths' => [__DIR__ . '/ValidatorBuilder.php'],
             'respect.validation.rule_factory.namespaces' => ['Respect\\Validation\\Validators'],
-            ValidatorFactory::class => factory(static fn(Container $container) => new NamespacedValidatorFactory(
-                $container->get(Transformer::class),
-                $container->get('respect.validation.rule_factory.namespaces'),
-            )),
+            ValidatorFactory::class => factory(static function (Container $container) {
```
henriquemoody, line 73: It would be great to also deprecate `NamespacedValidatorFactory`

henriquemoody (commented):
This is really awesome! 

I have just a few questions, but it looks good. Another question of mine is whether you've benchmarked those changes. Have you seen any significant differences?

henriquemoody:
Another thought that occurred to me is how we'll integrate Respect\Parameters with the `FluentValidatorFactory`. Have you thought about that already?

src/ContainerRegistry.php
alganet: That happens in #1730.

alganet:
@henriquemoody I'm working on the container cleanup, that will come soon.

src-dev/Commands/LintMixinCommand.php
alganet: They could! We could in fact annotate the parameter directly with some kind of `#[ComposableIgnore]` (individual parameters can accept attributes).

It is a deeper change though, and it would require a Fluent+FluentGen release.

It's tracked: https://github.com/Respect/Fluent/issues/27

I would like to tackle it after this, the upcoming changes to container/injection and #1730 come online (in that order), but I'm willing to work on it right now if you feel this is urgent, it's just a little more release logistics.

alganet:
@TheRespectPanda benchmark

TheRespectPanda:
Triggered phpbench benchmarks for PR #1729 — workflow run: https://github.com/Respect/Validation/actions/workflows/ci-perf.yml

src-dev/Commands/LintMixinCommand.php
henriquemoody: Nah, since we've already established that that wouldn't be a BC break, I'm really fine with it

alganet:
@henriquemoody benchmarks will be here 

https://github.com/Respect/Validation/actions/runs/28156799528/job/83387267406

`CompositeValidatorsBench` is the most relevant here. There seems to be a cost, but I don't know if it's a stale baseline or not ([I remember no regression](https://github.com/Respect/Validation/pull/1729#issuecomment-4114183604), so I need to double check, and I will before merging).

henriquemoody:
> I'm working on the container cleanup, that will come soon.

Sounds good to me! I think we should hold version 3.2 until we finish that stuff, so we don't generate BC breaks, do you agree?

alganet:
@henriquemoody what BC break are we talking about? So much stuff lately that I'm lost 😅 

Either way, probably yes (lots of new features, this is 3.2 or 4.0)

henriquemoody:
I was talking about the `FluentValidatorFactory`. We'd probably need to inject the container to it, or the resolver. It might be an optional argument, so not to break the BC, but if we can just sort this out before releasing it will be easier for us.

No API breaks in `main` so far.

alganet:
@henriquemoody for the Parameter thing, you're talking about?

We can implement a local [FluentFactory](https://github.com/Respect/Fluent/blob/main/src/FluentFactory.php) that receives the parameter resolver. Then, we pass it to the FluentValidatorFactory instead of the `ComposingLookup`.

It's a composite pattern, these lookup classes can be nested and arranged and a user of the `Fluent` project can design their own. This way, we can keep Fluent without needing to depend on a dead-weight respect/parameter (or later move the auto-wiring lookup to Fluent with an optional dependency if we discover the design is reusable).

henriquemoody (approved):

alganet:
I ran a [rebaseline](https://github.com/Respect/Validation/actions/runs/28163047672/job/83408081214) through manual workflow invokation, then, I ran a [benchmark](https://github.com/Respect/Validation/actions/runs/28163329464/job/83409034977) again.

There are regressions in the composite validator timings, likely due to the indirection of going through one more class. They're more than I would like, but manageable.

I will be investigating them, but likely merge as it is. It's probable that the parameter work will add even more delay if my hypothesis is correct (if it is, indirection in the builder is bad and I'll work towards perf optimizing it).

PR #1788, opened 2026-06-11
Add Given validator

henriquemoody:
`When` is useful as a standalone ternary (`when → then : else`), but its default-AlwaysInvalid else branch makes it painful inside AllOf or ShortCircuit chains: every unmatched condition becomes a hard failure, forcing callers to pass an explicit AlwaysValid else just to opt out.

Given solves this by dropping the else parameter entirely. It validates $then when $when passes and passes silently otherwise — equivalent to When($when, $then, AlwaysValid), but with a clearer name and no boilerplate at call sites.

Includes the validator class, unit and feature tests, a doc page, cross-links in When.md, a SmokeTestProvider entry, and regenerated mixin interfaces so v::given(...) is available across all builder and chain surfaces.


alganet (approved):
Looks good as it is!

henriquemoody:
Awesome, thanks!! I will merge soon and tag 3.2 later -- I might add a few other things, so maybe it's best to wait a bit to release  3.2 -- but, I won't wait much. Better having a new version with just one extra validator than having it hanging, IMO

alganet:
@henriquemoody agreed, there's no need to wait. Shipping often is a good thing. Also, few people are on the 3.x as of right now, so the blast radius is small.

PR #1790, opened 2026-06-12
Resolve validator dependencies in the factory instead of validators

henriquemoody:
A few validators (CountryCode, CurrencyCode, LanguageCode, Phone, SubdivisionCode, and Uuid) fetched their dependencies straight from the ContainerRegistry. That coupled every validator to a global registry and forced anyone providing a custom container to define those services, even when they only wanted to customize something unrelated.

Validators now receive their dependencies as optional constructor arguments. The new ContainerArgumentsResolver inspects a constructor and augments the given arguments with whatever the container can provide for the parameters they do not fill, skipping PHP internal classes (such as DateTimeImmutable) that are value-like and must never come from the container. NamespacedValidatorFactory and the Attributes validator share this resolver, so rules created through the fluent API and through PHP attributes both honor services defined in the container. The resolver hides behind the ArgumentsResolver interface, allowing clients to replace the resolution strategy entirely.

When no service is available, validators fall back to instantiating their default dependency directly, and they detect missing optional packages with class_exists() instead of container lookups. This makes the MissingComposerDependencyException accurate regardless of how bare the client's container is, but it also means the missing-package paths can no longer be simulated by swapping containers, so the tests doing that were removed.

Creating a rule that resolves services from the container costs about 1µs more than before, the price of querying the container and mapping named arguments at creation time. Rules without resolvable parameters pay around 0.1µs, and rules whose arguments already fill the constructor skip the parameter inspection entirely.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -72,7 +72,8 @@
             "Respect\\Dev\\": "src-dev/",
             "Respect\\Validation\\": "tests/unit/",
             "Respect\\Validation\\Test\\": "tests/src/"
-        }
+        },
+        "files": ["tests/src/Functions/namedFunctionWithTransformer.php"]
     },
     "scripts": {
         "bench-profile": "vendor/bin/phpbench xdebug:profile",
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2,6 +2,7 @@
 SPDX-License-Identifier: MIT
 SPDX-FileCopyrightText: (c) Respect Project Contributors
 SPDX-FileContributor: Alexandre Gomes Gaigalas <alganet@gmail.com>
+SPDX-FileContributor: Henrique Moody <henriquemoody@gmail.com>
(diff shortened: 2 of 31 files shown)


alganet:
I believe the work done on [Respect\Parameter](https://github.com/Respect/Parameter) is related to this idea. Have you seen it?

henriquemoody:
Respect\Parameter doesn't augment parameter, it will resolve all dependencies from a callable, but I agree that Respect\Parameter is where this logic should live, though. I'm going to send a PR with some changes there as soon as I can

alganet:
With the merging of #1801, is there anything left from this one we would like to port as well? Maybe some optional feature that I missed or something.

Let me know!

henriquemoody:
I'm closing this, since this is no longer relevant. I don't think there's something still relevant here

PR #1793, opened 2026-06-24
Add EachKey validator for validating array keys

henriquemoody:
Each validates array values but there is no built-in way to validate their keys. The only workaround is transforming the input first with v::after('array_keys', v::each(...)), which is indirect and loses the composability that makes the library expressive. Without EachKey, validating something as common as array<string, int> requires an awkward transformation instead of a readable composition.

EachKey fills that gap by validating each key in an iterable against an inner validator, mirroring how Each validates values. This makes array-key constraints a first-class concern: v::eachKey(v::stringType()) is immediately readable, and combining it with Each via AllOf gives a clean expression for typed arrays like
v::allOf(v::eachKey(v::stringType()), v::each(v::intType())).

### DIFF
--- a/tests/feature/Validators/EachKeyTest.php
+++ b/tests/feature/Validators/EachKeyTest.php
@@ -1,0 +1,284 @@
+<?php
+
+/*
+ * SPDX-License-Identifier: MIT
+ * SPDX-FileCopyrightText: (c) Respect Project Contributors
+ * SPDX-FileContributor: Henrique Moody <henriquemoody@gmail.com>
+ */
+
+declare(strict_types=1);
+
+test('Non-iterable', catchAll(
+    fn() => v::eachKey(v::stringType())->assert(null),
+    fn(string $message, string $fullMessage, array $messages) => expect()
+        ->and($message)->toBe('`null` must be iterable')
+        ->and($fullMessage)->toBe('- `null` must be iterable')
+        ->and($messages)->toBe(['eachKey' => '`null` must be iterable']),
+));
+
+test('Default', catchAll(
+    fn() => v::eachKey(v::stringType())->assert([0 => 'a', 1 => 'b', 2 => 'c']),
+    fn(string $message, string $fullMessage, array $messages) => expect()
+        ->and($message)->toBe('Key `.0` must be a string')
+        ->and($fullMessage)->toBe(<<<'FULL_MESSAGE'
+        - Each key of `["a", "b", "c"]` must be valid
+          - Key `.0` must be a string
+          - Key `.1` must be a string
+          - Key `.2` must be a string
+        FULL_MESSAGE)
+        ->and($messages)->toBe([
+            '__root__' => 'Each key of `["a", "b", "c"]` must be valid',
+            0 => 'Key `.0` must be a string',
+            1 => 'Key `.1` must be a string',
+            2 => 'Key `.2` must be a string',
+        ]),
+));
+
+test('Inverted', catchAll(
+    fn() => v::not(v::eachKey(v::stringType()))->assert(['a' => 1, 'b' => 2, 'c' => 3]),
+    fn(string $message, string $fullMessage, array $messages) => expect()
+        ->and($message)->toBe('Key `.a` must not be a string')
+        ->and($fullMessage)->toBe(<<<'FULL_MESSAGE'
+        - Each key of `["a": 1, "b": 2, "c": 3]` must be invalid
+          - Key `.a` must not be a string
+          - Key `.b` must not be a string
+          - Key `.c` must not be a string
+        FULL_MESSAGE)
+        ->and($messages)->toBe([
+            '__root__' => 'Each key of `["a": 1, "b": 2, "c": 3]` must be invalid',
+            'a' => 'Key `.a` must not be a string',
+            'b' => 'Key `.b` must not be a string',
+            'c' => 'Key `.c` must not be a string',
+        ]),
+));
+
+test('With name, default', catchAll(
+    fn() => v::named('Outer', v::eachKey(v::named('Inner', v::stringType())))->assert([0 => 'a', 1 => 'b', 2 => 'c']),
+    fn(string $message, string $fullMessage, array $messages) => expect()
+        ->and($message)->toBe('Key Inner must be a string')
+        ->and($fullMessage)->toBe(<<<'FULL_MESSAGE'
+        - Each key of Outer must be valid
+          - Key `.0` must be a string
+          - Key `.1` must be a string
+          - Key `.2` must be a string
+        FULL_MESSAGE)
+        ->and($messages)->toBe([
+            '__root__' => 'Each key of Outer must be valid',
+            0 => 'Key `.0` must be a string',
+            1 => 'Key `.1` must be a string',
+            2 => 'Key `.2` must be a string',
+        ]),
+));
+
+test('With name, inverted', catchAll(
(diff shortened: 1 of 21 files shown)


tests/feature/Validators/EachKeyTest.php
```diff
@@ -0,0 +1,279 @@
...
+
+test('Default', catchAll(
+    fn() => v::eachKey(v::stringType())->assert([0 => 'a', 1 => 'b', 2 => 'c']),
+    fn(string $message, string $fullMessage, array $messages) => expect()
+        ->and($message)->toBe('The key of `.0` must be a string')
+        ->and($fullMessage)->toBe(<<<'FULL_MESSAGE'
+        - Each key in `["a", "b", "c"]` must be valid
+          - The key of `.0` must be a string
+          - The key of `.1` must be a string
+          - The key of `.2` must be a string
```
alganet: I would try to make the message like this:

```
        - Each key of `["a", "b", "c"]` must be valid
          - The key `0` must be a string
          - The key `1` must be a string
          - The key `2` must be a string
```

Heading using "of" instead of "in", and each item with the explicit key value (instead of the dot representation).

tests/feature/Validators/EachKeyTest.php
```diff
@@ -0,0 +1,279 @@
...
+    expect($validator->isValid([]))->toBeTrue();
+});
+
+test('short-circuit: non-iterable input', catchAll(
+    fn() => v::shortCircuit(v::eachKey(v::stringType()))->assert(null),
+    fn(string $message, string $fullMessage, array $messages) => expect()
+        ->and($message)->toBe('`null` must be iterable')
+        ->and($fullMessage)->toBe('- `null` must be iterable')
+        ->and($messages)->toBe(['eachKey' => '`null` must be iterable']),
+));
```
alganet, line 221: Here (in this test, not necessarily those lines) I think we should test with an SplObjectStorage that contains non-scalar keys.

It helps the robustness of the validator. We devs are used to arrays being simple strings or ints, but in fact during an iteration context, PHP allows keys to be whatever and SplObjectStorage is the perfect test subject for that. In it, a key can be a whole object we can validate further.

https://www.php.net/manual/en/splobjectstorage.offsetset.php (the manual has an example with an object as key).

alganet (commented):
Looks good, just a few nits:

 - The messages are effort sensitive. If it makes it too hard to use the actual keys instead of the `.dot` representation we can make a compromise.
 - The wording (switching "in" to "of" in the main message, dropping "of" in the items) is my opinion. I suggest passing the messages through an LLM for grammar checking.
 - Non-scalar keys (in advanced iterators) are niche and I understand if that complicates too much the implementation. If we decide to not do it, perhaps we should think of making the `IterableType` check more stricter (raw `ArrayType` that would block those advanced iterators from being used or equivalent).

tests/feature/Validators/EachKeyTest.php
henriquemoody: `SplObjectStorage` iterates with integer keys. When you `foreach ($storage as $key => $value)`, the keys are just sequential integers, and the values are the attached objects.

The "object keys" that are defined in `SplObjectStorage` are not exposed through foreach iteration, they become values.

alganet: You're right!

henriquemoody:
The problem is that some validators (like `each()`, `key()` , `property()`) set a path path on child results, and by not creating a path on `EachKey` it will inherit the parent path and become indistinguishable:

### Using actual keys (no `->withPath()`)

**`getMessage()`:**

| Scenario | Output |
|---|---|
| Standalone | `The key "a" must be an integer` ✅ |
| Inside `key(data)` | `The key .data must be an integer` ❌ both keys identical |
| Deep `key(user, key(settings))` | `The key .user.settings must be an integer` ❌ same |

**`getFullMessage()`:**

```
- Each key of `.data` must be valid
  - The key `.data` must be an integer
  - The key `.data` must be an integer
```

So using actual keys isn't viable when composed inside `key()`. The compromise I'd suggest is keeping the `.dot` representation but keeping "The key of":

### "The key of"  (current) — `getMessage()`

| Scenario | Output |
|---|---|
| Inside `key(data)` | `The key of .data.a must be an integer` |
| Deep `key(user, key(settings))` | `The key of .user.settings.theme must be an integer` |

### "The key" — `getMessage()`

| Scenario | Output |
|---|---|
| Inside `key(data)` | `The key .data.a must be an integer` |
| Deep `key(user, key(settings))` | `The key .user.settings.theme must be an integer` |

"The key `.data.a`" reads as if `.data.a` is a key name — but `a` is the key and `.data` is just the parent path.

"The key **of** `.data.a`" separates "key" from the path — `.data.a` is the location, not the key's name. The root message already scopes it ("Each key of `.data` must be valid"), so the child just needs to say which location and what failed.

I think "The key of" is the best compromise here.

alganet:
@henriquemoody what about dropping the entire prefix?

".data.a must be an integer"
".user.settings.theme must be an integer"

Just that, no "The key" nor "The key of".

Edit: in retrospect, not using the word "key" might make the message opaque and misleading (user thinks it's a value or something).

I'm ok with the compromise.

tests/feature/Validators/EachKeyTest.php
henriquemoody: I wrote a more elaborate comment about the "The key of", but about the "Each key of", I checked with different AIs, and they all say "in" is more idiomatic (which I agree), and it seems like it can be too repetitive if we keep the "The key of"

henriquemoody:
How about something like that?

```
Full message:
- Every key in `["a", "b", "c"]` must be valid
  - Key `.0` must be a string
  - Key `.1` must be a string
  - Key `.2` must be a string

Main message:
Key `.0` must be a string
 ```

tests/feature/Validators/EachKeyTest.php
alganet: @henriquemoody Opus 4.8 disagrees. It says "of" is better. Here's the session (which includes my framing in the prompt) and the answer:

https://claude.ai/share/4bc1494a-baf5-4ea7-a795-e091d2ae58ef


> "Each key of [...]" is better here.
> "In" reads as locating keys inside a collection, which invites the reader to look for them among the serialized contents — and your serialization shows values ("a", "b", "c"), not keys, so "in" actively misleads. "Of" frames the keys as a property derived from the structure, which is exactly what eachKey does: it iterates the keys regardless of how the container is rendered. It also matches your sub-messages, which already say "The key of .0".
So "of" is both more accurate to the semantics and consistent with the child lines.

henriquemoody: I'm happy to change. Could you also check if `Key` rather `The they of` would work well too? (I'm out of Opus tokens)

alganet: Sure!

https://claude.ai/share/4bc1494a-baf5-4ea7-a795-e091d2ae58ef (same chat, share updated with new question)

> "Key .0 must be a string" is better.
> "The key of .0" is circular: .0 is the key (or path to it), so "the key of the key" reads oddly — and it gets worse with nesting, where "the key of .foo.bar.baz" suggests baz has some further key, when baz is itself the thing being asserted.
> "Key .0" treats the identifier as a label for the thing under assertion, which scales cleanly to paths: "Key .foo.bar.baz must be a string" reads as "the key located at this path," which is exactly right.
> It's also terser and matches the implicit grammar of path-based messages, where the identifier names the node rather than pointing to something it contains.

henriquemoody: I've made the changes

henriquemoody: Thanks!

alganet (approved):

PR #1799, opened 2026-06-25
Extract PropertyResolver strategy

henriquemoody:
Previously, the resolution of validators from property types and explicit new resolution strategies required editing the class, and a class-typed property annotated with #[Attributes] was validated twice — once by the explicit-attribute branch and again by the declared-type branch — which tripped the circular-reference guard on the second pass.

This commit moves that responsibility to a dedicated PropertyResolver interface with composable implementations, so each strategy can be developed and composed independently. The composite collapses duplicate Attributes entries, ensuring each property is validated exactly once even when multiple resolution paths produce the same instance. The Attributes class is reduced to its single responsibility, and resolver chains become pluggable through the constructor.

### DIFF
--- a/src/Validators/Attributes.php
+++ b/src/Validators/Attributes.php
@@ -28,8 +28,7 @@ use Respect\Validation\Validators\Attributes\DeclaredTypePropertyResolver;
 use Respect\Validation\Validators\Attributes\ExplicitAttributePropertyResolver;
 use Respect\Validation\Validators\Attributes\PropertyResolver;
 use Respect\Validation\Validators\Core\Reducer;
-
-use function spl_object_id;
+use WeakMap;
 
 #[Composable(without: [All::class, Key::class, Property::class, Not::class, UndefOr::class])]
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
@@ -42,8 +41,8 @@ final class Attributes implements Validator
 {
     public const string TEMPLATE_CIRCULAR_REFERENCE = '__circular_reference__';
 
-    /** @var array<int, true> */
-    private array $visited = [];
+    /** @var WeakMap<object, true> */
+    private WeakMap $visited;
 
     private readonly PropertyResolver $propertyResolver;
 
@@ -54,6 +53,7 @@ final class Attributes implements Validator
             new ExplicitAttributePropertyResolver(new BypassResolver()),
             new DeclaredTypePropertyResolver(),
         );
+        $this->visited = new WeakMap();
     }
 
     public function evaluate(mixed $input): Result
@@ -64,12 +64,11 @@ final class Attributes implements Validator
             return $objectType->withId($id);
         }
 
-        $objectId = spl_object_id($input);
-        if (isset($this->visited[$objectId])) {
+        if ($this->visited->offsetExists($input)) {
             return Result::failed($input, $this, [], self::TEMPLATE_CIRCULAR_REFERENCE)->withId($id);
         }
 
-        $this->visited[$objectId] = true;
+        $this->visited[$input] = true;
 
         $reflection = new ReflectionObject($input);
         $validators = [...$this->getClassValidators($reflection), ...$this->getPropertyValidators($reflection)];
@@ -127,5 +126,16 @@ final class Attributes implements Validator
         }
 
         return $properties;
+    }
+
+    /** @return list<string> */
+    public function __sleep(): array
+    {
+        return ['propertyResolver'];
+    }
+
+    public function __wakeup(): void
+    {
+        $this->visited = new WeakMap();
     }
 }
--- a/src/Validators/Attributes.php
+++ b/src/Validators/Attributes.php
@@ -15,17 +15,18 @@ namespace Respect\Validation\Validators;
 use Attribute;
 use ReflectionAttribute;
 use ReflectionClass;
-use ReflectionIntersectionType;
-use ReflectionNamedType;
 use ReflectionObject;
 use ReflectionProperty;
-use ReflectionUnionType;
 use Respect\Fluent\Attributes\Composable;
-use Respect\Parameter\Resolver;
 use Respect\Validation\Id;
 use Respect\Validation\Message\Template;
 use Respect\Validation\Result;
 use Respect\Validation\Validator;
+use Respect\Validation\Validators\Attributes\BypassResolver;
+use Respect\Validation\Validators\Attributes\CompositePropertyResolver;
+use Respect\Validation\Validators\Attributes\DeclaredTypePropertyResolver;
+use Respect\Validation\Validators\Attributes\ExplicitAttributePropertyResolver;
+use Respect\Validation\Validators\Attributes\PropertyResolver;
 use Respect\Validation\Validators\Core\Reducer;
 
 use function spl_object_id;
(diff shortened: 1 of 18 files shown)


src/Validators/Attributes/DeclaredTypePropertyResolver.php
```diff
@@ -0,0 +1,58 @@
...
+
+/*
+ * SPDX-License-Identifier: MIT
+ * SPDX-FileCopyrightText: (c) Respect Project Contributors
+ * SPDX-FileContributor: Henrique Moody <henriquemoody@gmail.com>
+ */
+
+declare(strict_types=1);
+
+namespace Respect\Validation\Validators\Attributes;
```
alganet, line 11: I'm not sure about the namespacing.

Until now, we kept the `Validators` namespace containing exclusively Validator instances.

The `Fluent` family even builds on this convention, scanning folders for classes that it assumes are all the same type within a namespace (it has an interface check, but still).

The way I see, the resolvers are *not* validators and should instead be somewhere else.

src/Validators/Attributes/PropertyResolver.php
```diff
@@ -0,0 +1,24 @@
...
+use Respect\Validation\Validator;
+use Respect\Validation\Validators\Attributes;
+
+/**
+ * Resolves validators from properties.
+ */
+interface PropertyResolver
+{
+    /** @return array<Validator> */
+    public function resolve(ReflectionProperty $property, Attributes $attributes): array;
```
alganet, line 23: I'm not sure about `Attributes` being here. 

Feels like juggling around the instance so you can return it back. 

From the perspective of the interface itself, a cleaner contract would be "just return me validators for *this property*", and `Attributes` is not one of them.

Maybe I missed something though.

alganet (changes requested):
I see the namespace as a blocker (a better choice outside of the Validators/ pool would make more sense).

The PropertyResolver interface is more a design question/refinement. If there's a deeper reason for it that can't be cleaned up, then it's fine.

src/Validators/Attributes/PropertyResolver.php
alganet: Clarifying further, maybe:

```
    public function resolve(ReflectionProperty $property, string $type): array;
```

Would be the best interface, passing `Validator::class` as a `class-string` to $type.

In the future, I see this as possibly a component reuse for Respect\Data (entity attribute resolution). This is, however, for a future generalization and out of scope. But hopefully the direction clarifies why I'm picking on the interface!

henriquemoody: This is precisely how we prevent the circular cycle since we have Attributes being called recursively. That's why it's in the `Attributes` namespace, it's really specific to this validator.

src/Validators/Attributes/DeclaredTypePropertyResolver.php
henriquemoody: I'm fine to move it somewhere else, but they're not generic. They're very `Attributes` specific.

I would have been happier if that wasn't the case, but dealing with the cycle here was the path I found. Maybe we could pass the `WeakMap` and a `$depth` here, but I struggle to visualize this working now (without testing it first)

src/Validators/Attributes/PropertyResolver.php
alganet: That is an important implementation detail! Makes sense. Would it be possible to make that without declaring it in the `resolve` signature though?

"Get me these validators, but don't recurse on *this one specifically*" feels like a weird abstraction that gives away information about the underlying implementation, and instead better handled somewhere else.

Maybe you already tried. If that's the case, I can take a deeper look with fresh eyes to see if I can come up with something.

src/Validators/Attributes/DeclaredTypePropertyResolver.php
alganet: Am I missing something? Does the namespace declaration has any role in the recursion detection?

I'm discussing that `namespace Respect\Validation\Validators\Attributes;` is a resolve-mechanism inside the bucket that previously only held Validators. Maybe `namespace Respect\Validation\Attributes;` would fit better.

src/Validators/Attributes/PropertyResolver.php
henriquemoody: You might be getting into something... We could pass a key => value array with the string-class => Validator and replace them dynamically. It might make it even more robust actually

src/Validators/Attributes.php
```diff
@@ -41,11 +41,19 @@ final class Attributes implements Validator
...
-    private array $visited = [];
+    /** @var WeakMap<object, true> */
+    private WeakMap $visited;
 
-    public function __construct(private readonly Resolver|null $resolver = null)
-    {
+    private readonly PropertyResolver $propertyResolver;
+
+    public function __construct(
+        PropertyResolver|null $propertyResolver = null,
```
henriquemoody, line 50: @alganet because of this new injection, I would like to only release version 3.2 after this has been merged. However, there are a few things that are not optimal with this.

For starters, I think we should have a resolver that works on the class level rather than the property level because that will allow us to also use doc blocks in the future. It would then make sense that once of those class resolver take care of the properties by itself, so we don't need to call any PropertyResolver here.

src/Validators/Attributes/BypassResolver.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Respect\Validation\Validators\Attributes;
+
+use ReflectionFunctionAbstract;
+use Respect\Parameter\Resolver;
+
+final class BypassResolver implements Resolver
```
henriquemoody, line 8: I created this one here quickly, just to signal the use-case, but this should be in `Respect\Parameter`

src/Validators/Attributes/ExplicitAttributePropertyResolver.php
```diff
@@ -0,0 +1,55 @@
...
+use ReflectionClass;
+use ReflectionProperty;
+use Respect\Parameter\Resolver;
+use Respect\Validation\Validator;
+use Respect\Validation\Validators\Attributes;
+
+final class ExplicitAttributePropertyResolver implements PropertyResolver
+{
+    public function __construct(
+        private readonly Resolver $resolver,
```
henriquemoody, line 23: @alganet I'm wondering if we should pass the factory here instead of the resolver.

Part of me thinks it would be a big overhead, but if we want to do something like [I mentioned in Parameter](https://github.com/Respect/Parameter/issues/35#issuecomment-4810009732) we will need somewhere to mediate between the Resolver and the creation of each validator.

Another part of me want to keep only the Resolver here, to make things more simpler at the cost of making things inflexible.

src/Validators/Attributes.php
alganet: Are you worried about breaking BC on the parameter signature?

For me, `Attributes` is parameter-less. We could document that the parameter is internal and subject to change, and only say it's public when it's done.

We need to do more of that (what is internal vs what is public) so we can have more freedom. Not all mechanisms in validation are public API surface.

---

The resolver you mention is a subsystem of Validation that is starting to couple cache ideas, docblock scanning and so on. It has now become more than a validator.

I think this subsystem deserves a careful design, and it's likely to take a lot of time, so I'm concerned about blocking release of other features while we do it. Perhaps there's a way we can deliver recursive attributes (lower tier improvement) before going full-feature.

src/Validators/Attributes/ExplicitAttributePropertyResolver.php
alganet: I still don't understand the real use case for that. I mean, I get the `Animal -> Goat, Cat` analogy but I don't get why we need that in Validation.

PR #1801, opened 2026-06-25
Integrate Respect/Config and Respect/Parameter

alganet:
Replace the PHP-DI container with Respect/Config: ContainerRegistry now builds a Respect\Config\Container, registering the rules' external dependencies (PHP ISO Codes, libphonenumber, ramsey/uuid) and a ParameterResolver.

Instantiate rules through a new AutowiringLookup: a Respect\Fluent FluentFactory that resolves the rule name and autowires its constructor dependencies from the container via Respect\Parameter, wrapped by FluentValidatorFactory. The dead NamespacedValidatorFactory is removed.

Drop static ContainerRegistry access from CountryCode, CurrencyCode, LanguageCode, SubdivisionCode, Phone and Uuid: each accepts its external dependency and otherwise constructs it directly, guarded by class_exists for the missing-package error. Inject a ParameterResolver into Attributes so attribute-declared rules are autowired too.

Fix Email: drop the func_num_args() guard, which autowiring defeats (the resolver always passes the defaulted argument), so v::email() uses egulias again instead of silently falling back to filter_var.

Exclude the autowired dependency types from the generated mixins, update the docs, and add AutowiringLookupTest.


alganet:
@TheRespectPanda benchmark

TheRespectPanda:
Triggered phpbench benchmarks for PR #1801 — workflow run: https://github.com/Respect/Validation/actions/workflows/ci-perf.yml

docs/validators/Phone.md
```diff
@@ -10,6 +10,7 @@ SPDX-FileContributor: Henrique Moody <henriquemoody@gmail.com>
 
 - `Phone()`
 - `Phone(string $countryCode)`
+- `Phone(string $countryCode, PhoneNumberUtil $phoneNumberUtil)`
```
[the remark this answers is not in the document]
alganet: That's a docs-lint bug, unfortunatelly.

src/Validators/Attributes.php
```diff
@@ -107,7 +112,12 @@ private function getPropertyInnerValidators(ReflectionProperty $property): array
...
         foreach ($property->getAttributes(Validator::class, ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
-            $propertyValidator = $attribute->getName() === self::class ? $this : $attribute->newInstance();
+            if ($attribute->getName() === self::class) {
+                $propertyValidator = $this;
+            } else {
+                $propertyValidator = $this->instantiateAttribute($attribute);
+            }
+
             $hasExplicitAttributes = $propertyValidator === $this;
             $propertyValidators[] = $propertyValidator;
```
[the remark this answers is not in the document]
alganet: Pre-existing bug. Fixed either way.

src/AutowiringLookup.php
```diff
@@ -0,0 +1,72 @@
...
+            if ($constructor === null) {
+                return $reflection->newInstanceArgs($arguments);
+            }
+
+            return $reflection->newInstanceArgs($this->parameterResolver->resolve($constructor, $arguments));
+        } catch (ReflectionException $exception) {
+            throw new CouldNotCreate(
+                sprintf('Could not instantiate "%s": %s', $name, $exception->getMessage()),
+                previous: $exception,
+            );
```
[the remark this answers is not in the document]
alganet: Fixed.

alganet:
Benchmarks show regression similar to previous FluentGen PR, which is strange.

https://github.com/Respect/Validation/actions/runs/28167881836

~I'm still investigating those regressions.~ They're real regressions. Indirection has slowed things down a handful of microseconds. I don't have a fix for it right now (both resolving arguments and building the chain got a little slower).

A possible fix is to mark validators that take injectable parameters explicitly, and moving the `AutowiringLookup` to `Fluent` so we skip a composite method call. I would prefer to deal with those in a separate PR!

src/Validators/Phone.php
```diff
@@ -50,31 +49,38 @@ final class Phone implements Validator
...
+
+    public function __construct(
+        string|null $countryCode = null,
+        Countries|null $countries = null,
+        PhoneNumberUtil|null $phoneNumberUtil = null,
+    ) {
+        if ($phoneNumberUtil === null && !class_exists(PhoneNumberUtil::class)) {
             throw new MissingComposerDependencyException(
                 'Phone rule requires libphonenumber for PHP',
                 'giggsey/libphonenumber-for-php',
```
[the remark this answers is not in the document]
alganet: Decided not to change. libphonenumber-for-php should still be compatible, and fighting the codecov patch coverage for this change would be a nightmare.

docs/validators/Attributes.md
```diff
@@ -8,6 +8,7 @@ SPDX-FileContributor: Henrique Moody <henriquemoody@gmail.com>
 # Attributes
 
 - `Attributes()`
+- `Attributes(ParameterResolver $resolver)`
```
henriquemoody: ~~Will be reverting thise one~~ Oh, nevermind

src/Validators/Attributes.php
```diff
@@ -43,6 +44,10 @@ final class Attributes implements Validator
     /** @var array<int, true> */
     private array $visited = [];
 
+    public function __construct(private readonly ParameterResolver|null $resolver = null)
```
henriquemoody: It's okay to merge this, but I will be changing the contract later.

src/AutowiringLookup.php
```diff
@@ -0,0 +1,72 @@
...
+    public function __construct(
+        private NamespaceLookup $lookup,
+        private FluentResolver $resolver,
+        private ParameterResolver $parameterResolver,
+    ) {
+    }
+
+    public function withNamespace(string $namespace): static
+    {
+        return new self($this->lookup->withNamespace($namespace), $this->resolver, $this->parameterResolver);
```
henriquemoody:
```suggestion
        return clone ($this, ['lookup' => $this->lookup->withNamespace($namespace)]);
```
Nitpick

src/ContainerRegistry.php
```diff
@@ -59,50 +63,56 @@ final class ContainerRegistry
...
+            ResultFilter::class => new Instantiator(OnlyFailedChildrenResultFilter::class),
+            'respect.validation.formatter.message' => new Autowire(FirstResultStringFormatter::class),
+            'respect.validation.formatter.full_message' => new Autowire(NestedListStringFormatter::class),
+            'respect.validation.formatter.messages' => new Autowire(NestedArrayFormatter::class),
             'respect.validation.ignored_backtrace_paths' => [__DIR__ . '/ValidatorBuilder.php'],
             'respect.validation.rule_factory.namespaces' => ['Respect\\Validation\\Validators'],
-            ValidatorFactory::class => factory(static function (Container $container) {
+            ParameterResolver::class => static fn(Container $container) => new Resolver($container),
+            ValidatorFactory::class => static function (Container $container) {
+                $lookup = new NamespaceLookup(
```
henriquemoody, line 84: Shouldn't we autowire those dependencies as well?

src/NamespacedValidatorFactory.php
```diff
@@ -1,98 +0,0 @@
...
-use Respect\Validation\Transformers\Transformer;
-use Respect\Validation\Transformers\ValidatorSpec;
-
-use function array_merge;
-use function Respect\Stringifier\stringify;
-use function sprintf;
-use function trim;
-use function ucfirst;
-
-final readonly class NamespacedValidatorFactory implements ValidatorFactory
```
henriquemoody, line 27: I would prefer to only  deprecate this one, so not to release another MAJOR version after such a short period since the last one

henriquemoody (commented):
There's some documentation to update, look for `ContainerRegistry` in the docs/. It looks good, though!

henriquemoody:
Do like the idea of being deliberate to which validators we want to resolve arguments. I saw most of the performance hits are no validators like AllOf, AnyOf that have constructor arguments. I'm also fine doing it on another PR.

src/ContainerRegistry.php
alganet: Sorry, which ones?

src/NamespacedValidatorFactory.php
alganet: This one is final, and never mentioned in the docs. Is it really part of the public API?

I was going to wait as well, but I noticed no one used it.

src/Validators/Attributes.php
alganet: You got the idea, right? It resolves its own resolver. So, if you introduce another parameter, that's fine, at any position.

I think not doing that makes those architectural sidecars which are harder to justify.

The fact that users can feed their own resolvers is mostly a bonus.

henriquemoody: I got the ideal, and I'm totally on board with it. I was just signaling I will be changing this later

alganet:
> Do like the idea of being deliberate to which validators we want to resolve arguments.

Yes, that's what I mentioned [here](https://github.com/Respect/Validation/pull/1801#issuecomment-4798929956)! I think there could be a regain in performance.

The question is how to mark that explicitly in a way that the ecosystem supports (eg custom validators).

Attributes are a probable answer, but that's complicated (where should it stay?).

 - Attribute lookup stays in Validation: another introspection layer (I'm trying to move those out)
 - Attribute goes into Fluent (now Fluent needs to know Parameter)
 - Attribute goes into Config (likely Parameter stops being the main interface, and the container resolves)

I currently dislike all of those options 😞 I don't actually know if that's where most of the perf loss went. The ideal path is to measure first what are the hot paths. I need a day or two for that and for a nice refined solution.

src/ContainerRegistry.php
henriquemoody: First I thought about all dependencies from `FluentValidatorFactory::__construct`, but on second thought, only the `NamespaceLookup`. It's more of a question, not a request

src/NamespacedValidatorFactory.php
henriquemoody: It's the only factory we have. The way I see, any class we expose is part of our public API, whether we mention it in the documentation or not (unless we tag them with `@internal`). This change will break `ValidationBundle`, for example, but it will break to any person who has been overwriting the default container.

henriquemoody:
> Attributes are a probable answer, but that's complicated (where should it stay?).

Seems like something for the resolver, to be honest. Probably means we'd have two different resolvers as well, one that automatically handlers that, and one that expects a PHP attribute. 


--
Edit: The reality is that dealing with arguments like we are doing here is something that can only be solved by the Resolver, it's not a standard behavior, so I think it should be there.

src/ContainerRegistry.php
alganet: NamespaceLookup is an internal. We don't want to allow users to override it (not yet, I haven't imagine a use case for it that is not covered by `respect.validation.rule_factory.namespaces`).

If we want to change it later, making it a key of the container would mean BC worries.

src/NamespacedValidatorFactory.php
alganet: That's unfortunate. The integration point is supposed to be the interface `ValidatorFactory`, not the concrete `NamespacedValidatorFactory`.

It means the bundle is relying on an implementation instead of the interface that declares the contract.

I will restore it and mark as deprecated. However, I still stand on the hill that this is not public and doesn't break "theoretical" BC.

alganet:
@henriquemoody addressed the items.

https://github.com/Respect/Parameter/pull/33 is still under review and I would like to have it and release another `respect\parameter` though, but it would be nice if you could resolve/elaborate on the other PR comments meanwhile.

src/NamespacedValidatorFactory.php
henriquemoody: Perhaps we should talk about that in depth another time, because it seems like we have different ideas what constitutes a BC break. I appreciate you adding the deprecation, though

src/NamespacedValidatorFactory.php
```diff
@@ -24,6 +24,11 @@
 use function trim;
 use function ucfirst;
 
+/**
+ * @deprecated This has been superseeded by Respect\Fluent
+ *
+ * @see \Respect\Validation\FluentValidatorFactory
+ */
 final readonly class NamespacedValidatorFactory implements ValidatorFactory
```
henriquemoody, line 32:
```suggestion
#[Deprecated('Use Respect\Validation\FluentValidatorFactory instead', '3.2')]
final readonly class NamespacedValidatorFactory implements ValidatorFactory
```

alganet: image

Do you know how to make it phpstan-compatible?

henriquemoody: Oh, my bad. I guess this attribute only works on functions?

alganet: I actually don't know! I never used it before, just found out about it.

PR #1802, opened 2026-06-26
Add IDE type narrowing to generated validator mixins

alganet:
Part of a PR group: [Fluent](https://github.com/Respect/Fluent/pull/28) - [FluentGen](https://github.com/Respect/FluentGen/pull/2) - [Validation](https://github.com/Respect/Validation/pull/1802)

---

Make validator chains narrow types for IDEs and PHPStan without the FluentAnalysis extension, driven entirely by the generated src/Mixins PHPDoc.

- Annotate src/Validators with #[Assurance] / #[AssuranceSubject] declaring each rule's assured type.
- Regenerate src/Mixins: Chain becomes generic (@template-covariant TSure); static entry methods narrow to Chain<concrete>; assert()/check() carry an unconditional @phpstan-assert TSure. Container rules (key/property/length/ max/min) and the concrete prefix forms (nullOrIntType, keyIntType, allIntType) narrow; argument-wrapping and compose forms stay Chain<mixed> so a raw (non-fluent) Validator argument is still accepted.
- isValid() intentionally does not narrow: its only conditional form is a two-way guard, unsound for inexact rules on the false branch.
- Add the tests/inference static-narrowing suite (no extension config) and run it in CI; scope the phpstan/phpcs accommodations for generated mixins.

---

This PR group adds advanced static (no PHPStan extension needed) type support and narrowing for fluent chains, pioneering it for Validation (StringFormatter should also be compatible with this approach)

image

In the example above, the devsense extension for VSCode is correctly infering an `int[]` (iterable int) from a `each(int())` fluent validation chain.

More examples in `tests/inference/assertions/static-narrowing.php`.

---

Notes:

 - Most of the big diff is generated mixins, the PR is actually small.
 - This work is related, but distinct from #1730. FluentAnalysis offers dynamic type narrowing with much richer narrowings (but IDEs can't run the extension). This one offers basic, static narrowing for IDEs such as devsense (tested) and IntelliJ (untested).

### DIFF
--- a/src-dev/Commands/LintMixinCommand.php
+++ b/src-dev/Commands/LintMixinCommand.php
@@ -18,6 +18,7 @@ use Respect\FluentGen\Fluent\InterfaceConfig;
 use Respect\FluentGen\Fluent\MethodBuilder;
 use Respect\FluentGen\Fluent\MixinGenerator;
 use Respect\FluentGen\Fluent\PrefixConstantsGenerator;
+use Respect\FluentGen\Fluent\TerminalMethod;
 use Respect\FluentGen\NamespaceScanner;
 use Respect\Validation\Mixins\Chain;
 use Respect\Validation\Validator;
@@ -82,6 +83,7 @@ final class LintMixinCommand extends Command
                     suffix: 'Builder',
                     returnType: Chain::class,
                     static: true,
+                    emitNarrowing: true,
                 ),
                 new InterfaceConfig(
                     suffix: 'Chain',
@@ -89,6 +91,29 @@ final class LintMixinCommand extends Command
                     rootExtends: [Validator::class],
                     rootComment: '@mixin ValidatorBuilder',
                     rootUses: [ValidatorBuilder::class],
+                    emitNarrowing: true,
+                    templateParam: 'TSure',
+                    terminalMethods: [
+                        new TerminalMethod(
+                            name: 'assert',
+                            returnType: 'void',
+                            parameters: ['input' => 'mixed'],
+                            comments: ['@phpstan-assert TSure $input', '@psalm-assert TSure $input'],
+                            optionalParameters: ['template' => 'mixed'],
+                        ),
+                        new TerminalMethod(
+                            name: 'check',
+                            returnType: 'void',
+                            parameters: ['input' => 'mixed'],
+                            comments: ['@phpstan-assert TSure $input', '@psalm-assert TSure $input'],
+                            optionalParameters: ['template' => 'mixed'],
+                        ),
+                        new TerminalMethod(
+                            name: 'isValid',
+                            returnType: 'bool',
+                            parameters: ['input' => 'mixed'],
+                        ),
+                    ],
                 ),
             ],
         );
--- a/src/Validators/BoolVal.php
+++ b/src/Validators/BoolVal.php
@@ -18,6 +18,7 @@ declare(strict_types=1);
 namespace Respect\Validation\Validators;
 
 use Attribute;
+use Respect\Fluent\Attributes\Assurance;
 use Respect\Validation\Message\Template;
 use Respect\Validation\Validators\Core\Simple;
 
@@ -32,6 +33,7 @@ use const FILTER_VALIDATE_BOOLEAN;
     '{{subject}} must be a boolean',
     '{{subject}} must not be a boolean',
 )]
+#[Assurance(type: 'scalar|null')]
 final class BoolVal extends Simple
 {
     public function isValid(mixed $input): bool
--- a/src/Validators/DateTimeDiff.php
+++ b/src/Validators/DateTimeDiff.php
@@ -14,6 +14,7 @@ namespace Respect\Validation\Validators;
 use Attribute;
 use DateTimeImmutable;
 use DateTimeInterface;
+use Respect\Fluent\Attributes\Assurance;
 use Respect\Validation\Exceptions\InvalidValidatorException;
 use Respect\Validation\Helpers\CanValidateDateTime;
 use Respect\Validation\Message\Template;
@@ -44,6 +45,7 @@ use function in_array;
     'For comparison with {{now|raw}}, {{subject}} must not be a datetime in the format {{sample|raw}}',
     self::TEMPLATE_WRONG_FORMAT,
 )]
+#[Assurance(type: ['string', DateTimeInterface::class])]
 final readonly class DateTimeDiff implements Validator
 {
     use CanValidateDateTime;
--- a/src/Validators/DateTimeDiff.php
+++ b/src/Validators/DateTimeDiff.php
@@ -14,6 +14,7 @@ namespace Respect\Validation\Validators;
 use Attribute;
 use DateTimeImmutable;
 use DateTimeInterface;
+use Respect\Fluent\Attributes\Assurance;
 use Respect\Validation\Exceptions\InvalidValidatorException;
 use Respect\Validation\Helpers\CanValidateDateTime;
 use Respect\Validation\Message\Template;
@@ -44,6 +45,7 @@ use function in_array;
     'For comparison with {{now|raw}}, {{subject}} must not be a datetime in the format {{sample|raw}}',
     self::TEMPLATE_WRONG_FORMAT,
 )]
+#[Assurance(type: ['string', 'int', DateTimeInterface::class])]
 final readonly class DateTimeDiff implements Validator
 {
     use CanValidateDateTime;
--- a/src/Validators/FalseVal.php
+++ b/src/Validators/FalseVal.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 namespace Respect\Validation\Validators;
 
 use Attribute;
+use Respect\Fluent\Attributes\Assurance;
 use Respect\Validation\Message\Template;
 use Respect\Validation\Validators\Core\Simple;
 
@@ -26,6 +27,7 @@ use const FILTER_VALIDATE_BOOLEAN;
     '{{subject}} must evaluate to `false`',
     '{{subject}} must not evaluate to `false`',
 )]
+#[Assurance(type: 'scalar|null')]
 final class FalseVal extends Simple
 {
     public function isValid(mixed $input): bool
--- a/src/Validators/Identical.php
+++ b/src/Validators/Identical.php
@@ -15,6 +15,9 @@ declare(strict_types=1);
 namespace Respect\Validation\Validators;
 
 use Attribute;
+use Respect\Fluent\Attributes\Assurance;
+use Respect\Fluent\Attributes\AssuranceFrom;
+use Respect\Fluent\Attributes\AssuranceParameter;
 use Respect\Fluent\Attributes\Composable;
 use Respect\Validation\Message\Template;
 use Respect\Validation\Result;
@@ -26,9 +29,11 @@ use Respect\Validation\Validator;
     '{{subject}} must be identical to {{compareTo}}',
     '{{subject}} must not be identical to {{compareTo}}',
 )]
+#[Assurance(from: AssuranceFrom::Value)]
 final readonly class Identical implements Validator
 {
     public function __construct(
+        #[AssuranceParameter]
         private mixed $compareTo,
     ) {
     }
--- a/src/Validators/Json.php
+++ b/src/Validators/Json.php
@@ -20,6 +20,7 @@ declare(strict_types=1);
 namespace Respect\Validation\Validators;
 
 use Attribute;
+use Respect\Fluent\Attributes\Assurance;
 use Respect\Validation\Message\Template;
 use Respect\Validation\Validators\Core\Simple;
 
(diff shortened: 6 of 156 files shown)


alganet:
@henriquemoody these are in draft because of the branch-alias (we need to amend the real composer versions when we do the release dance).

However, although not packagist-ready yet, they are open for review of the code. I requested review only here to avoid spam but you're free to comment in any of the 3.

src/Validators/TrueVal.php
```diff
@@ -26,6 +27,7 @@
     '{{subject}} must evaluate to `true`',
     '{{subject}} must not evaluate to `true`',
 )]
+#[Assurance(type: 'scalar')]
```
henriquemoody, line 30:
```suggestion
#[Assurance(type: 'true')]
```

Can we do that?

src/Validators/Undef.php
```diff
@@ -28,6 +29,7 @@
     '{{subject}} must be undefined',
     '{{subject}} must be defined',
 )]
+#[Assurance(type: "null|''", exact: true)]
```
henriquemoody:
```suggestion
#[Assurance(type: ['null', "''", exact: true)]
```
Isn't that possible?

src/Validators/BoolVal.php
```diff
@@ -32,6 +33,7 @@
     '{{subject}} must be a boolean',
     '{{subject}} must not be a boolean',
 )]
+#[Assurance(type: 'scalar|null')]
```
henriquemoody, line 36:
```suggestion
#[Assurance(type: ['scalar', 'null'])]
```

I won't make comments like this one, but this is in other places as well. I'm just wondering what's the diffeerence?

src/Validators/DateTimeDiff.php
```diff
@@ -44,6 +45,7 @@
     'For comparison with {{now|raw}}, {{subject}} must not be a datetime in the format {{sample|raw}}',
     self::TEMPLATE_WRONG_FORMAT,
 )]
+#[Assurance(type: ['string', DateTimeInterface::class])]
```
henriquemoody, line 48:
```suggestion
#[Assurance(type: ['string', 'integer', DateTimeInterface::class])]
```
I think we also allow passing timestamps

src/Validators/FalseVal.php
```diff
@@ -26,6 +27,7 @@
     '{{subject}} must evaluate to `false`',
     '{{subject}} must not evaluate to `false`',
 )]
+#[Assurance(type: 'scalar|null')]
```
henriquemoody, line 30:
```suggestion
#[Assurance(type: 'scalar|null|false')]
```

src/Validators/Identical.php
```diff
@@ -26,6 +28,7 @@
     '{{subject}} must be identical to {{compareTo}}',
     '{{subject}} must not be identical to {{compareTo}}',
 )]
+#[Assurance(from: AssuranceFrom::Value, exact: true)]
```
henriquemoody: How doesn't it know which value? Is it always the first one?

src/Validators/Json.php
```diff
@@ -31,6 +32,7 @@
     '{{subject}} must be a JSON string',
     '{{subject}} must not be a JSON string',
 )]
+#[Assurance(type: 'string')]
```
henriquemoody, line 35: I think the implementation of validator might be "incorrect"

```
php > var_dump(json_validate(1));
bool(true)
php > var_dump(json_validate(1.0));
bool(true)
php > var_dump(json_validate(true));
bool(true)
php > var_dump(json_validate('{}'));
bool(true)
```

I feel like a real JSON is a string, though, so probably there's nothing we should do about here -- but I'm not so sure.

src/Validators/KeyExists.php
```diff
@@ -29,6 +30,7 @@
     '{{subject}} must be present',
     '{{subject}} must not be present',
 )]
+#[Assurance(type: ['array', ArrayAccess::class])]
```
henriquemoody, line 33: Don't we have any "array key exists" assurance? It would be quite important

src/Validators/PropertyExists.php
```diff
@@ -28,6 +29,7 @@
     '{{subject}} must be present',
     '{{subject}} must not be present',
 )]
+#[Assurance(type: 'object')]
```
henriquemoody, line 32: Don't we have any assurance for "property exists"?

src/Validators/ScalarVal.php
```diff
@@ -24,6 +25,7 @@
     '{{subject}} must be a scalar',
     '{{subject}} must not be a scalar',
 )]
+#[Assurance(type: 'int|float|bool|string', exact: true)]
```
henriquemoody:
```suggestion
#[Assurance(type: 'scalar', exact: true)]
```

Or am I missing something?

src/Validators/ShortCircuit.php
```diff
@@ -13,11 +13,14 @@
...
 use Attribute;
+use Respect\Fluent\Attributes\Assurance;
+use Respect\Fluent\Attributes\AssuranceCompose;
 use Respect\Validation\Helpers\CanEvaluateShortCircuit;
 use Respect\Validation\Result;
 use Respect\Validation\Validator;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
+#[Assurance(compose: AssuranceCompose::Intersect, exact: true)]
 final readonly class ShortCircuit implements Validator
```
henriquemoody, line 21: Because of how this validator works, I'm not sure we should assure something here. What do you think?

src/Validators/Templated.php
```diff
@@ -12,12 +12,15 @@
...
 use Attribute;
+use Respect\Fluent\Attributes\Assurance;
+use Respect\Fluent\Attributes\AssuranceCompose;
 use Respect\Fluent\Attributes\Composable;
 use Respect\Validation\Result;
 use Respect\Validation\Validator;
 
 #[Composable(without: [All::class, Key::class, Property::class, Not::class, NullOr::class, UndefOr::class])]
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
+#[Assurance(compose: AssuranceCompose::Intersect, composeRange: [1, 1], exact: true)]
```
henriquemoody: I just want to say that it's incredibly awesome that you managed to make this work!

henriquemoody (commented):
This is trully trully awesome!!

src/Validators/ShortCircuit.php
alganet: The FluentAnalysis extension can do this one. I ported the `#[Assurance` blocks from that PR, and some of them have no outcome in this PR. 

This is one of them.

Perhaps it will be better if I remove the ones that do not yield any significant narrowing for this PR.

henriquemoody: I understood that it can, but should it? Because we short circuit, isn't it arguable if it will make all those assertions? I'm not sure, not saying we shouldn't, but I'm not sure

src/Validators/Identical.php
alganet: The `Fluent` API uses `#[AssuranceParameter]` to index the value (if missing, it takes the first one).

Thanks for noticing this, it's better explicit. And while doing, I noticed a bug (a "class-string" infer that slipped through a temporary spike, see Instance.php which is missing from: that specifies the nature of the parameter indexing).

src/Validators/KeyExists.php
alganet: That's a great question.

For IDE support, no (I don't think it's possible just with mixin annotations). 

For `phpstan analyze`, we can do that (but it's not fully annotated in `KeyExists`).

src/Validators/Templated.php
alganet: Don't get too excited, this has no deep effect in this PR.

I'm considering now dropping the Assurances that would only work on #1730 to avoid confusion, then adding them later.

src/Validators/TrueVal.php
alganet: We could, but `TrueVal` is not strict. It accepts "on" and "yes" and so on, and those are strings.

Users would likely need to use `identical(true)` for this.

src/Validators/Undef.php
alganet: It's a choice. Keeping the type encapsulated makes it easier to develop the phpstan extension, and also for refactoring (search & replace the same string-notation as the docblock).

For classes, having `FooBar::class` makes it easier to refactor, while `string|MyClass` is more opaque, so there's this tradeoff.

This belongs in the Fluent&FluentGen PRs though (and the FluentAnalysis repo), where you can see how this is used internally.

src/Validators/FalseVal.php
alganet: A false is a scalar though. Might be better off for clarity. Same issue as `TrueVal`

src/Validators/DateTimeDiff.php
alganet: It's not on the tests though, so we might want to tackle refining that later?

src/Validators/Json.php
alganet: We do this

```
        if (!is_string($input) || $input === '') {
            return false;
        }
```

before json_validate. Assurance is correct to our behavior.

src/Validators/ScalarVal.php
alganet: For this PR, it's the same.

For FluentAnalysis, there are some things we don't do yet (like unpacking `scalar` into the list for multi-node hop narrowing).

So, in the future, we might have to pre-unpack all types in declaration instead of using the more elegant one, just to give FluentAnalysis more surface. Or we might want to implement the unpacking in FluentAnalysis.

src/Validators/Json.php
henriquemoody: Yeah, I read the code before commenting. I was more questioning the validator itself than the assurance.

src/Validators/ShortCircuit.php
alganet: For this PR, it's a null-op. It only matters for FluentAnalysis. So I'll remove it, and we discuss when I integrate it!

src/Validators/Json.php
alganet: If we start refactoring validators in this PR I think we would be in trouble! 😅 

Anyways, I like that `v::json()` accepts only strings. It's what makes more sense in the real world, despite `json_validate` juggling behaviors.

A `1.0` passed to `json_validate` and going through is a PHP footgun that this validator solves. We all expect JSONs to be strings in most scenarios (controllers, logs, etc).

alganet:
I have removed the `#[Assurance` attributes that are not relevant for this PR, keeping only the ones that actively contribute to the IDE-only static narrowing. The ones removed belong to the full `FluentAnalysis` phpstan-extension integration that doesn't really affect IDE behavior.

Hopefully, this would make clear what IDEs can do and what they cannot (the static-narrowing.php file is a good reference card though, and remains the same).

src/Validators/KeyExists.php
alganet: This would need FluentAnalysis (phpstan extension), not currently possible just by generating mixins.

src/Validators/PropertyExists.php
alganet: Also only possible with a phpstan extension, not viable using mixins.

src/Validators/ScalarVal.php
alganet: We can't instruct IDEs into unpacking `scalar` into the list, and some of they don't do that. PHPStan does, and the FluentAnalysis extension can be made to do that by using the types package from phpstan.

For that reason, I would prefer using the unpacked explicit list. `scalar` is an alias for that list, and if by exposing the list we target more IDEs, that would be my preference.

src/Validators/DateTimeDiff.php
alganet: Done.

src-dev/Commands/LintMixinCommand.php
```diff
@@ -82,13 +83,37 @@ interfaces: [
...
+                    emitNarrowing: true,
                 ),
                 new InterfaceConfig(
                     suffix: 'Chain',
                     returnType: Chain::class,
                     rootExtends: [Validator::class],
                     rootComment: '@mixin ValidatorBuilder',
                     rootUses: [ValidatorBuilder::class],
+                    emitNarrowing: true,
+                    templateParam: 'TSure',
```
henriquemoody, line 95: Clever

tests/inference/assertions/static-narrowing.php
```diff
@@ -0,0 +1,908 @@
...
+}
+
+// ============================================================================
+// Container subject: the input is narrowed to the container type (key/property/
+// length/max/min), and the prefixed form narrows identically.
+// ============================================================================
+
+function keyNarrowsToContainer(mixed $x): void
+{
+    v::key('a', v::intType())->assert($x);
```
henriquemoody, line 773: It would have been amazing if PHPStan or Psalm would allows us to assert the type of a specific key. They don't have it, to my knowledge

henriquemoody (approved):
This is absolutely amazing! I'm so happy you got this one working, it really puts Validation on the next level!

PR #980, opened 2018-06-01
Apply contribution guidelines to "Cpf" rule

williamespindola:

PR #1821, opened 2026-07-29
Guard Cpf and NfeAccessKey against non-scalar input

gaoflow:
`Cpf` and `NfeAccessKey` call string functions on raw `mixed` input, so `v::cpf()->isValid([])` (or an object, `null`, etc.) throws a `TypeError` where every sibling in the document/checksum family (`Cnpj`, `Bsn`, `Cnh`, `Pis`, …) returns `false` — they all start with an `is_scalar` guard that these two never got (#980 tried to standardise `Cpf` but was closed unmerged).

The fix mirrors `Cnpj`: guard non-scalars, then cast scalars to string, so `v::cpf()->isValid(11598647644)` behaves like the string form instead of crashing. I checked every `Simple` subclass that hits raw `$input` with a string function and only these two were affected (`Luhn` looks similar but delegates to the guarded `Digit` first). Tests cover non-scalar input returning `false` and the int form of a valid CPF still passing.

### DIFF
--- a/src/Validators/Cpf.php
+++ b/src/Validators/Cpf.php
@@ -24,6 +24,7 @@ use Respect\Validation\Message\Template;
 use Respect\Validation\Validators\Core\Simple;
 
 use function intval;
+use function is_scalar;
 use function mb_strlen;
 use function preg_match;
 use function preg_replace;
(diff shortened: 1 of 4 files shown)


alganet (approved):

alganet:
Thanks for your contribution! :panda_face:

PR #1832, opened 2026-08-13
Fix double validation and state leaks in Attributes recursion

alganet:
Recursive validation for nested objects guarded against validating a nested object twice by checking whether the property already carried #[Attributes] directly on the property. That guard did not recognise the rule inside a wrapper, and the state backing the recursion was never released, producing five defects:

- #[NullOr(new Attributes())] on a class-typed property validated the nested object twice and reported every nested failure twice. This was the documented way of validating a nullable nested object before recursion existed, so upgrading silently duplicated messages.
- The same object held by two sibling properties failed as a circular reference, because visited objects accumulated for the whole traversal instead of the current path.
- An Attributes instance could only be used once: nothing ever cleared the visited objects, so every evaluation after the first failed.
- A union type whose value satisfied more than one of its class members recursed once per member, and the second one reported the object it had just visited as a circular reference.
- A custom validator attribute with cyclic internal state made wrapped-rule detection recurse indefinitely.

Detect the rule anywhere inside a property's attributes rather than only at the top level, so a wrapped Attributes suppresses implicit recursion as an explicit one does. Make the rule immutable and give each recursion level its own instance carrying the path to the object being evaluated, so visited objects represent the path from the root rather than every object ever seen. Collapse union recursion into a single Given over the disjunction of its class members. Track the validators visited while inspecting a wrapped rule, preventing a cyclic validator graph from overflowing the stack.

The detection only runs for properties whose type can hold an object, so attributes with large arguments, such as #[In], no longer pay for it.

---

This is a smaller, more focused fix for the recursion problem that does not introduce an external resolver like #1799. It is intended to focus on the behavior, leaving open the possibility of introducing such resolver in the dependency chain as a substitute non-breaking change.

### DIFF
--- a/src/Validators/Attributes.php
+++ b/src/Validators/Attributes.php
@@ -19,6 +19,8 @@ use ReflectionIntersectionType;
 use ReflectionNamedType;
 use ReflectionObject;
 use ReflectionProperty;
+use ReflectionReference;
+use ReflectionType;
 use ReflectionUnionType;
 use Respect\Fluent\Attributes\Composable;
 use Respect\Parameter\Resolver;
@@ -28,6 +30,7 @@ use Respect\Validation\Result;
 use Respect\Validation\Validator;
 use Respect\Validation\Validators\Core\Reducer;
 
+use function is_array;
 use function spl_object_id;
 
 #[Composable(without: [All::class, Key::class, Property::class, Not::class, UndefOr::class])]
@@ -37,15 +40,16 @@ use function spl_object_id;
     '{{subject}} must contain a circular reference',
     Attributes::TEMPLATE_CIRCULAR_REFERENCE,
 )]
-final class Attributes implements Validator
+final readonly class Attributes implements Validator
 {
     public const string TEMPLATE_CIRCULAR_REFERENCE = '__circular_reference__';
 
     /** @var array<int, true> */
-    private array $visited = [];
+    private array $path;
 
-    public function __construct(private readonly Resolver|null $resolver = null)
+    public function __construct(private Resolver|null $resolver = null)
     {
+        $this->path = self::rootPath();
     }
 
     public function evaluate(mixed $input): Result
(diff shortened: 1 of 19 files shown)


src/Validators/Attributes.php
```diff
@@ -110,46 +119,105 @@ private function getPropertyValidators(ReflectionObject $reflection): array
...
+        $instances = [];
+        foreach ($type->getTypes() as $innerType) {
+            if (!$innerType instanceof ReflectionNamedType || $innerType->isBuiltin()) {
+                continue;
+            }
+
+            /** @var class-string $class */
+            $class = $innerType->getName();
+            $instances[] = new Instance($class);
+        }
```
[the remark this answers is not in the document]
alganet: Done.

src/Validators/Attributes.php
```diff
@@ -110,46 +119,105 @@ private function getPropertyValidators(ReflectionObject $reflection): array
...
+    /** @param array<int, true> $visited */
+    private static function containsSelfIn(mixed $value, array &$visited): bool
+    {
+        if ($value instanceof self) {
+            return true;
+        }
+
+        if (is_array($value)) {
+            foreach ($value as $item) {
+                if (self::containsSelfIn($item, $visited)) {
```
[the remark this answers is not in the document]
alganet: Fixed.

src/Validators/Attributes.php
```diff
@@ -37,15 +40,16 @@
...
     public const string TEMPLATE_CIRCULAR_REFERENCE = '__circular_reference__';
 
     /** @var array<int, true> */
-    private array $visited = [];
+    private array $path;
 
-    public function __construct(private readonly Resolver|null $resolver = null)
+    public function __construct(private Resolver|null $resolver = null)
     {
+        $this->path = self::rootPath();
```
henriquemoody, line 52: Nica catch. Could everything that's related to tracking this be a different object? The class became quite big after this change, and it's already not the smallest 😅

alganet: @henriquemoody likely yes, but then injecting those objects and designing that surface becomes a concern. 

The class would shrink a little (it's actually only 267 lines total with my change), but the PR would grow significantly.

It is something I want to do eventually. Also, I want to make the `array<int, true> $path` readonly as well, but that also requires deeper surgery.

henriquemoody: I think we don't event need to inject it as a dependency, just instantiating the object (that could also hold the state) is enough. I see it more as a way to separate concerns but not change the contract of the constructor.

henriquemoody (approved):
I'm fine to keep it as is. I will pre-approve it, feel free to merge it if you would prefer keeping all the code in the `Attributes`

PR #1843, opened 2026-09-01
Report unknown size instead of passing false or 0 bytes to the inner validator

ZamboniL:
`Size` casted the return of `filesize()` to an integer, and `(int) false === 0`, so a
filename that does not exist or is not readable validated as 0 bytes. Every upper-bound
check passed on a file that was never there:

    v::size('KB', v::lessThan(2))->assert('/path/that/does/not/exist');

`SplFileInfo::getSize()` throws when `stat` fails, so the same missing file escaped as an
uncaught `RuntimeException` instead of a validation failure; its documented return type is
`int|false`, so overrides may return `false` rather than throw ([Symfony's `Stream`](https://github.com/symfony/symfony/blob/5e9597f80aa6570bfea5fb3852362d8b41b48240/src/Symfony/Component/HttpFoundation/File/Stream.php#L21)
does exactly that). PSR-7 streams and uploaded files may legitimately report `null` for an
unknown size, and that was reported as `TEMPLATE_WRONG_TYPE`, which is misleading: the type
is fine, only the size is not knowable.

Check the input type up front, so unsupported types keep reporting `TEMPLATE_WRONG_TYPE`,
and report a supported input whose size cannot be determined under the new
`TEMPLATE_UNKNOWN_SIZE`. Guard `filesize()`, catch the `RuntimeException` from `SplFileInfo`
and narrow every `getSize()` result with `is_int()`. `getSize()` was also being called twice
per evaluation; it is now called once.

Inputs that previously validated against a bogus 0 now fail, which is a behaviour change. The existing
template constant and message are untouched.

---

There is no open issue for this. I found it in a Symfony API validating user-supplied
files, which arrive as `HttpFoundation\File\File` an `SplFileInfo` subclass.

### DIFF
--- a/src/Validators/Size.php
+++ b/src/Validators/Size.php
@@ -20,9 +20,11 @@ use Respect\Validation\Exceptions\InvalidValidatorException;
 use Respect\Validation\Message\Template;
 use Respect\Validation\Result;
 use Respect\Validation\Validator;
+use RuntimeException;
 use SplFileInfo;
 
 use function filesize;
+use function is_file;
 use function is_string;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
@@ -36,9 +38,16 @@ use function is_string;
     '{{subject}} must not be a filename, an instance of SplFileInfo or a PSR-7 interface',
     self::TEMPLATE_WRONG_TYPE,
 )]
+#[Template(
+    'It is not possible to determine the size of {{subject}}',
+    'It is not possible to determine the size of {{subject}}',
+    self::TEMPLATE_UNKNOWN_SIZE,
+)]
 final readonly class Size implements Validator
 {
     public const string TEMPLATE_WRONG_TYPE = '__wrong_type__';
+
+    public const string TEMPLATE_UNKNOWN_SIZE = '__unknown_size__';
 
     private const array DATA_STORAGE_UNITS = [
         'B' => ['name' => 'bytes', 'bytes' => 1],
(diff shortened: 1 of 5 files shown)


alganet:
Looks good to me! Thanks for your contribution :panda_face: 

I'll wait a couple of days to give a chance for @henriquemoody to review it.

I would probably remove the `// coments`. It's our style to not have them, but currently nothing enforces their absence so I won't require you to remove them (would be nice though; that would also be best as an *amend* instead of a separate commit).

This will probably be in either `3.2` or `4.0` (the main branch will increase one of those counters soon). Let me know if you need a backport for a 3.1.x or 2.x release, if it's something important I can work it out.

ZamboniL:
Thank you for the quick response, I deleted the comments with a ammend.

I can wait til the next `minor`/`major`, no problem!

src/Validators/Size.php
```diff
@@ -64,37 +74,60 @@ public function __construct(
...
-            return $input->getSize();
+            return is_int($size) ? $size : null;
         }
 
-        if ($input instanceof StreamInterface) {
-            return $input->getSize();
+        if ($input instanceof UploadedFileInterface || $input instanceof StreamInterface) {
+            $size = $input->getSize();
+
+            return is_int($size) ? $size : null;
```
henriquemoody:
```suggestion
            return $input->getSize();
```

Nitpick: getSize already returns either null or the size.

src/Validators/Size.php
```diff
@@ -64,37 +74,60 @@ public function __construct(
...
-        }
+            try {
+                $size = $input->getSize();
+            } catch (RuntimeException) {
+                return null;
+            }
 
-        if ($input instanceof UploadedFileInterface) {
-            return $input->getSize();
+            return is_int($size) ? $size : null;
```
henriquemoody:
```suggestion
            return $size === false ? null : $size;
```

It's a nitpick, but it makes us have 1 less import, which makes it slight better

henriquemoody (commented):
Thanks for those changes, @ZamboniL!

I made a few comments, all nick pick. Not required, but would make the code a bit neater.

@alganet, I'm fine if you want to merge it as is.

ZamboniL:
Nits applied, thanks for the review.

alganet (approved):

